PR #3, opened 2020-08-11
reduced hand rotation in DigCircular

damianacatanoso:
Ticket: https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-445
This is to avoid collision between grinder and terrain during scooping, in particular DigCircular when radial==False. I implemented both Ussama's suggestion of reducing the hand rotation and Terry's suggestion of relocating the grinder to a 180 deg angle from the scoop. Both work, but Ussama's approach is simpler and avoids the "floating grinder" visualization. So I pushed only this one. I have Terry's approach in another local branch, which I didn't push.

Try calling DigCircular and check that there is collision. Try calling other services that might be compromised by this modification. Let me know your thoughts, thanks!

### DIFF
--- a/ow_lander/scripts/activity_full_digging_traj.py
+++ b/ow_lander/scripts/activity_full_digging_traj.py
@@ -175,7 +175,7 @@ def dig_circular(move_arm, move_limbs, args):
     go_to_Z_coordinate(move_limbs, x_start, y_start, z_start)
 
     # Rotate hand perpendicular to radial direction
-    change_joint_value(move_arm, constants.J_HAND_YAW, -math.pi/2.2)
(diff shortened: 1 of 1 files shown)


Samahu:
@damianacatanoso I can only see the change in to the arm location, what about the other change to move the grinder? is that in a different repo or a branch?

damianacatanoso:
@Samahu, here I only modified the target angle that the hand should achieve after its rotation to collect the sample in DigCircular. No modifications are required in the grinder motion.

Samahu:
Looks about right:
image

damianacatanoso:
> @damianacatanoso I can only see the change in to the arm location, what about the other change to move the grinder? is that in a different repo or a branch?

I have Terry's approach in another local branch that I didn't push!

PR #4, opened 2020-08-18
OCEANWATER-346: Use a different topic name for dynamic terrain model and visual plugins

Samahu:
## Summary of changes:
* Collision between the terrain and the end effectors is now enabled
* Topics corresponding to /ow_dynamic_terrain/modify_terrain_* are left intact. Plugin-specific topics {visual and collision} are added to the *DynamicTerrainVisual* and *DynamicTerrainModel* plugins
* Python scripts `modify_terrain_scoop_pub.py` and `modify_terrain_grinder_pub.py` corresponding to the scoop and grinder has been updated to use the Plugin-specific topics with a slightly different parameters for visual vs collision.
 - The `modify_terrain_*/collision` message is merely a scaled up version of the `modify_terrain_*/visual` message  
* README file updated with the new changes

## Verify
Launch the simulation against the `europa_terminator_workspace`:
```bash
```

Use the rqt service caller to perform any arm planning operation that may interact with the terrain. e.g.
- /planning/arm/dig_circular
- /planning/arm/dig_linear
- /planning/arm/grind
Followed by a service call to `/planning/arm/publish_trajectory` with use_latest=True

**Observe the arm modifies the terrain and the arm is not colliding with the terrain.**

## Experiments 
To see the actual size of the terrain set the `modify_terrain_*/visual` message scale value in the  `modify_terrain_scoop_pub.py` and `modify_terrain_grinder_pub.py` to the same scale value of the `modify_terrain_*/collision` message.

### DIFF
--- a/ow_dynamic_terrain/include/memory_ext.h
+++ b/ow_dynamic_terrain/include/memory_ext.h
@@ -1,0 +1,21 @@
+#ifndef MEMORY_EXT_H
+#define MEMORY_EXT_H
+
+// TODO: consider moving the file to a common place
+
+#include <memory>
+
+#if __cplusplus < 201402L  // if less than C++14 is detected add the extension
+
+namespace std
+{
+template <typename T, typename... Args>
+std::unique_ptr<T> make_unique(Args&&... args)
+{
+  return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
+}
+}  // namespace std
+
+#endif
+
+#endif  // MEMORY_EXT_H
--- a/ow_dynamic_terrain/include/memory_ext.h
+++ b/ow_dynamic_terrain/include/memory_ext.h
@@ -1,5 +1,14 @@
+// The Notices and Disclaimers for Ocean Worlds Autonomy Testbed for Exploration
+// Research and Simulation can be found in README.md in the root directory of
+// this repository.
+
 #ifndef MEMORY_EXT_H
 #define MEMORY_EXT_H
+
+/**
+ * This module adds the missing 'make_unique' when compiling against C++11 standard. For more info on this refer to:
+ * Scott Meyers' Effective Modern C++ Item 21: Prefer std::make_unique and std::make_shared to direct use of new.
+ */
 
 // TODO: consider moving the file to a common place
 
--- a/ow_dynamic_terrain/scripts/modify_terrain_grinder_pub.py
+++ b/ow_dynamic_terrain/scripts/modify_terrain_grinder_pub.py
@@ -29,6 +29,15 @@ class ModifyTerrainGrinder:
         "/gazebo/link_states", LinkStates, self.handle_link_states)
 
   def compose_modify_terrain_circle_message(self, position, scale=1.0):
+    """ Composes a modify_terrain_circle that matches the grinder end effector
+    it is positioned downwards ready for digging. When the grinder is position
+    downward for digging the shape it projects on the terrain as a circle.
+
+    Parameters:
+      position: Corresponds to the center of the modify_terrain_circle operation
+      scale: A value that uniformaly scales the generated modify_terrain_circle
+        message
+    """
     return modify_terrain_circle(position=Point(position.x, position.y, position.z),
                                  outer_radius=0.008*scale, inner_radius=0.001*scale,
                                  weight=-0.2*scale, merge_method="min")
--- a/ow_dynamic_terrain/scripts/modify_terrain_grinder_pub.py
+++ b/ow_dynamic_terrain/scripts/modify_terrain_grinder_pub.py
@@ -21,15 +21,17 @@ class ModifyTerrainGrinder:
   def __init__(self, *args):
     rospy.init_node("modify_terrain_grinder_pub", anonymous=True)
     self.last_translation = np.zeros(3)
-    self.pub = rospy.Publisher(
-        'ow_dynamic_terrain/modify_terrain_circle', modify_terrain_circle, queue_size=1)
+    self.pub_visual = rospy.Publisher(
+        'ow_dynamic_terrain/modify_terrain_circle/visual', modify_terrain_circle, queue_size=1)
+    self.pub_collision = rospy.Publisher(
+        'ow_dynamic_terrain/modify_terrain_circle/collision', modify_terrain_circle, queue_size=1)
     self.states_sub = rospy.Subscriber(
         "/gazebo/link_states", LinkStates, self.handle_link_states)
 
-  def compose_modify_terrain_circle_message(self, position):
+  def compose_modify_terrain_circle_message(self, position, scale=1.0):
     return modify_terrain_circle(position=Point(position.x, position.y, position.z),
-                                 outer_radius=0.008, inner_radius=0.001,
-                                 weight=-0.2, merge_method="min")
+                                 outer_radius=0.008*scale, inner_radius=0.001*scale,
+                                 weight=-0.2*scale, merge_method="min")
 
   def check_and_submit(self, new_position, new_rotation):
     """ Checks if position has changed significantly before submmitting a message """
@@ -46,12 +48,12 @@ class ModifyTerrainGrinder:
     if all(np.isclose(current_translation, self.last_translation, atol=1.e-4)):
       return  # no significant change, abort
 
-
-
     self.last_translation = current_translation
 
     msg = self.compose_modify_terrain_circle_message(new_position)
-    self.pub.publish(msg)
+    self.pub_visual.publish(msg)
+    msg = self.compose_modify_terrain_circle_message(new_position, scale=2.0)
+    self.pub_collision.publish(msg)
     
     rospy.logdebug_throttle(1, "modify_terrain_grinder message:\n" + str(msg))
 
--- a/ow_dynamic_terrain/scripts/modify_terrain_scoop_pub.py
+++ b/ow_dynamic_terrain/scripts/modify_terrain_scoop_pub.py
@@ -15,23 +15,24 @@ from gazebo_msgs.msg import LinkStates
 from tf.transformations import euler_from_quaternion
 from ow_dynamic_terrain.msg import modify_terrain_ellipse
 
-
 class ModifyTerrainScoop:
 
   def __init__(self, *args):
     rospy.init_node("modify_terrain_scoop_pub", anonymous=True)
     self.last_translation = np.zeros(3)
-    self.pub = rospy.Publisher(
-        'ow_dynamic_terrain/modify_terrain_ellipse', modify_terrain_ellipse, queue_size=1)
+    self.visual_pub = rospy.Publisher(
+        'ow_dynamic_terrain/modify_terrain_ellipse/visual', modify_terrain_ellipse, queue_size=1)
+    self.collision_pub = rospy.Publisher(
+        'ow_dynamic_terrain/modify_terrain_ellipse/collision', modify_terrain_ellipse, queue_size=1)
     self.states_sub = rospy.Subscriber(
         "/gazebo/link_states", LinkStates, self.handle_link_states)
 
-  def compose_modify_terrain_ellipse_message(self, position, orientation):
+  def compose_modify_terrain_ellipse_message(self, position, orientation, scale=1.0):
     return modify_terrain_ellipse(position=Point(position.x, position.y, position.z),
                                   orientation=orientation,
-                                  outer_radius_a=0.002, outer_radius_b=0.005,
-                                  inner_radius_a=0.001, inner_radius_b=0.001,
-                                  weight=-0.025, merge_method="min")
+                                  outer_radius_a=0.002*scale, outer_radius_b=0.005*scale,
+                                  inner_radius_a=0.001*scale, inner_radius_b=0.001*scale,
+                                  weight=-0.025*scale, merge_method="min")
 
   def check_and_submit(self, new_position, new_rotation):
     """ Checks if position has changed significantly before submmitting a message """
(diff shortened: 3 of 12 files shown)


damianacatanoso:
Get "unstable" behavior similar to guarded_move: https://youtu.be/STjibPuHp4Y

damianacatanoso:
Here is another instance of the same problem when publishing DigLinear: https://youtu.be/nJH5f5cYCy4

kmdalal:
I see the same thing in Rviz.

﻿On 8/18/20, 3:31 PM, "damianacatanoso" <notifications@github.com> wrote:

    Get "unstable" behavior similar to guarded_move: https://youtu.be/STjibPuHp4Y

damianacatanoso:
> I see the same thing in Rviz. ﻿On 8/18/20, 3:31 PM, "damianacatanoso" <notifications@github.com> wrote: Get "unstable" behavior similar to guarded_move: https://youtu.be/STjibPuHp4Y

Mike, do you run Gazebo at all?

kmdalal:
Just the server.  The client (GUI) never ran in any VM for me.  I may try to adjust the VM resources again, but I am planning to get a native Linux machine.

﻿On 8/18/20, 3:40 PM, "damianacatanoso" <notifications@github.com> wrote:


    I see the same thing in Rviz. On 8/18/20, 3:31 PM, "damianacatanoso" notifications@github.com wrote: Get "unstable" behavior similar to guarded_move: https://youtu.be/STjibPuHp4Y


    Mike, do you run Gazebo at all?

damianacatanoso:
> Just the server. The client (GUI) never ran in any VM for me. I may try to adjust the VM resources again, but I am planning to get a native Linux machine. ﻿On 8/18/20, 3:40 PM, "damianacatanoso" <notifications@github.com> wrote: I see the same thing in Rviz. On 8/18/20, 3:31 PM, "damianacatanoso" notifications@github.com wrote: Get "unstable" behavior similar to guarded_move: https://youtu.be/STjibPuHp4Y Mike, do you run Gazebo at all?

Oh I see, you are running the server. It makes sense we see the same behaviour!

damianacatanoso:
@Samahu I just pulled ow_simulator. Should I pull ow_europa or others as well?

kmdalal:
Yeah, it wasn't obvious to me what repo this branch is for.  I just guessed only ow_simulator.

damianacatanoso:
> Yeah, it wasn't obvious to me what repo this branch is on. I just guessed only ow_simulator.

If you place your cursor on "master" or "OCEANWATER-346_Use_a_diffe..[name of branch]" it explicitely says the repo belongs to ow_simulator. I asked just to be sure!

ow_dynamic_terrain/src/DynamicTerrainBase.h
```diff
@@ -0,0 +1,65 @@
...
+
+protected:
+  std::unique_ptr<ros::NodeHandle> m_node_handle;
+
+protected:
+  ros::CallbackQueue m_callback_queue;
+
+protected:
+  std::vector<ros::Subscriber> m_subscribers;
+};
```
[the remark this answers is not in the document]
Samahu: It is something I picked from Gazebo code (https://github.com/osrf/gazebo/blob/gazebo9/gazebo/rendering/Heightmap.hh). I don't always use this style but since this was a Gazebo plugin I went with it. No strong preference, I like it partially because it it aligns with C# and other languages and I don't need to worry much about grouping fields that have the same access modifier.

Samahu:
> Get "unstable" behavior similar to guarded_move: https://youtu.be/STjibPuHp4Y

@damianacatanoso I didn't see this while testing but recently I was only testing with the grinder. I'll try and let you know.

>Note: The changes are only made to ow_simulator

Samahu:
@damianacatanoso @mikedalal  I did verify that the same problems shows up on my machine. I was not sure at what point this problem started to take place. The problem wasn't occurring when I was testing against the scoop otherwise I would have noticed it. After trying few things I found out that the problem actually started when I enabled collision between the terrain and the grinder

I don't know exactly why as the grinder is not effectively touching the terrain in the case of a circle  dig.

mogumbo:
I have only had time to try dig_circular. I see the same arm jumping motion that Damiana recorded in her video. I've been working on the move_guarded problem in OCEANWATER-399. Using the quick solver instead of world solver in the .world file helps a lot. Unfortunately, in this case, merging my OCEANWATER-399 branch in ow_simulator did not help any more. It has smaller arm torque limits that added stability in other test cases, but it did not help this situation. Anyway, I see this as a problem unrelated to this ticket.

Super cool digging, by the way!

ow_dynamic_terrain/src/DynamicTerrainBase.h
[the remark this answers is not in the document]
mogumbo: I prefer removing the redundant lines. Also, I find it useful to group public, private, and protected methods together as much as possible. It reduces the code size because you need fewer access qualifiers , and it's easier to learn to use a class when you can see all its public methods together in the same place.

ow_dynamic_terrain/src/DynamicTerrainBase.cpp
```diff
@@ -0,0 +1,68 @@
...
+  if (!ros::isInitialized())
+  {
+    gzerr << m_plugin_name + ": ROS not initilized! The plugin won't load" << endl;
+    return;
+  }
+
+  m_node_handle = make_unique<ros::NodeHandle>(m_plugin_name);
+  m_node_handle->setCallbackQueue(&m_callback_queue);
+
+  auto on_modify_terrain_circle = [this](const modify_terrain_circle::ConstPtr& msg) {
```
mogumbo, line 18: This is a good use for auto.

ow_dynamic_terrain/src/DynamicTerrainBase.cpp
```diff
@@ -0,0 +1,68 @@
...
+  });
+
+  gzlog << m_plugin_name << ": successfully loaded!" << endl;
+}
+
+template <typename T>
+void DynamicTerrainBase::subscribe(const std::string& topic,
+                                   const boost::function<void(const boost::shared_ptr<T const>&)>& callback)
+{
+  auto topic_fqn = "/" + m_package_name + "/" + topic;
```
mogumbo, line 48: Please use simple type names instead of auto.

ow_dynamic_terrain/src/DynamicTerrainBase.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+gazebo::rendering::Heightmap* DynamicTerrainBase::getHeightmap(gazebo::rendering::ScenePtr scene)
+{
+  if (!scene)
+  {
+    gzerr << m_plugin_name << ": Couldn't acquire scene!" << endl;
+    return nullptr;
+  }
+
+  auto heightmap = scene->GetHeightmap();
```
mogumbo, line 60: Here too. This is a simple type name.

ow_dynamic_terrain/scripts/modify_terrain_scoop_pub.py
```diff
@@ -15,23 +15,24 @@
...
         "/gazebo/link_states", LinkStates, self.handle_link_states)
 
-  def compose_modify_terrain_ellipse_message(self, position, orientation):
+  def compose_modify_terrain_ellipse_message(self, position, orientation, scale=1.0):
     return modify_terrain_ellipse(position=Point(position.x, position.y, position.z),
                                   orientation=orientation,
-                                  outer_radius_a=0.002, outer_radius_b=0.005,
-                                  inner_radius_a=0.001, inner_radius_b=0.001,
-                                  weight=-0.025, merge_method="min")
+                                  outer_radius_a=0.002*scale, outer_radius_b=0.005*scale,
```
mogumbo, line 33: Where do the numbers 0.002, 0.005, and 0.001 come from? Please add a comment to explain the logic here.

ow_dynamic_terrain/scripts/modify_terrain_grinder_pub.py
```diff
@@ -21,15 +21,17 @@ class ModifyTerrainGrinder:
...
+        'ow_dynamic_terrain/modify_terrain_circle/collision', modify_terrain_circle, queue_size=1)
     self.states_sub = rospy.Subscriber(
         "/gazebo/link_states", LinkStates, self.handle_link_states)
 
-  def compose_modify_terrain_circle_message(self, position):
+  def compose_modify_terrain_circle_message(self, position, scale=1.0):
     return modify_terrain_circle(position=Point(position.x, position.y, position.z),
-                                 outer_radius=0.008, inner_radius=0.001,
-                                 weight=-0.2, merge_method="min")
+                                 outer_radius=0.008*scale, inner_radius=0.001*scale,
```
mogumbo, line 33: Where do the numbers 0.008 and 0.001 come from? Please add a comment to explain the logic here. If there is to be a default size for a circle, I would expect it to be a unit circle.

ow_dynamic_terrain/include/memory_ext.h
```diff
@@ -0,0 +1,21 @@
...
+
+// TODO: consider moving the file to a common place
+
+#include <memory>
+
+#if __cplusplus < 201402L  // if less than C++14 is detected add the extension
+
+namespace std
+{
+template <typename T, typename... Args>
```
mogumbo, line 12: Please comment on what these parts do.

ow_lander/urdf/lander.xacro
```diff
@@ -543,15 +543,6 @@
       </plugin>
     </visual>
     <selfCollide>1</selfCollide>
-    <!-- Turn off collisions with the terrain so scoop can go underground. For
-         this to work, the terrain must use a different collide_bitmask. -->
-    <collision>
-      <surface>
```
mogumbo, line 549: Collisions are currently disabled for the hand as well. We should probably also remove that code block, unless there is some reason to keep it that I don't know about.

damianacatanoso:
FYI, I was trying drilling/digging in other directions (use_default=False). I used the grinder to dig a first hole. Then I used the grinder a second time to extend the previous hole. Even if I used only the grinder, I observed the same arm bouncing behavior that we observed at the end of dig linear/dig circular.

ow_dynamic_terrain/include/memory_ext.h
Samahu: addressed

ow_dynamic_terrain/scripts/modify_terrain_grinder_pub.py
Samahu: addressed

ow_dynamic_terrain/include/memory_ext.h
Samahu: C++11 forgot to add a make_unique function similar to make_shared so I had it added for C++11 only compiles

I think it is beneficial to add this in a common place such that  other modules can make use of it

ow_dynamic_terrain/scripts/modify_terrain_scoop_pub.py
Samahu: addressed

ow_dynamic_terrain/src/DynamicTerrainBase.h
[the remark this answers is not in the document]
Samahu: Gazebo applies this consistently through their code and it is not impeding me personally. I have also seen this applied in other frameworks and API. In any case, since there are two votes against one, I have the properties grouped.

ow_lander/urdf/lander.xacro
Samahu: I did try this before actually submitting this PR and it did actually cause slowness to happen

ow_dynamic_terrain/src/DynamicTerrainBase.cpp
Samahu: The simple type rule is really very hard to follow. I don't know which types are simple and which types are not. for example initially we pointed at integers and floats as simple types. Now we considering std::string a simple type and gazebo::rendering::Heightmap* as a simple type.

mogumbo: Yes, those are simple types, although I didn't call them out explicitly in the coding standard. It's easy to type either of those and easy to understand what they are. Using auto here makes the code smaller but harder to understand. I'm not opposed to using namespace gazebo::rendering to shorten things a bit, but some people might disagree with me on that one.

ow_lander/urdf/lander.xacro
mogumbo: Weird. If we're going to keep it, it would be a good idea to amend that comment to explain that is is a performance booster. Later on we should think about disabling collisions between other arm parts and the terrain; maybe we can get even more performance boost.

ow_dynamic_terrain/src/DynamicTerrainBase.cpp
Samahu: I don't agree to this narrative. It is almost always good to use **auto**. In any case, I have amended the PR to declare the type of each variable as I don't wish to delay the PR any further.

ow_dynamic_terrain/src/DynamicTerrainBase.cpp
Samahu: addressed

ow_dynamic_terrain/src/DynamicTerrainBase.cpp
Samahu: Again, I don't agree that the use of **auto** makes the code harder to read and understand (best counter example is python). In any case, this has been addressed.

ow_lander/urdf/lander.xacro
Samahu: addressed

PR #5, opened 2020-08-20
trajectory_publisher node publishes message with guarded_move result when guarded_move is published, for autonomy

damianacatanoso:
Hello!
Ticket: https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-434

Background:
We agreed some time ago that, after publishing guarded_move, there should be a message containing the following information: has ground been found successfully? What are the coordinates of the ground? In which reference frame? Autonomy will use this info to store the ground height at that specific location and feed it to the digging services. 

What I did here:
- I use the existing node "trajectory_publisher" to publish a message on a new topic, at the end of publish_trajectory for guarded_move. 
- the message is defined in ow_lander>msg, called "GuardedMoveResult.msg"
- the topic is called: "/guarded_move_result".

How to test:
- Terminal 1: roslaunch ow europa_terminator_workspace.launch

- Terminal 2: 
a) rosservice call /planning/arm/guarded_move True True 0 0 0 0 0 0 0
b) rosservice call /planning/arm/publish_trajectory True a
 
- Terminal 3, while trajectory is being published: 
a) rostopic list . Check if the topic "/guarded_move_result" is in the list
b) rostopic echo /guarded_move_result . wait until ground is found. The message should show up when it is found.

Questions for you:
- I am having the trajectory_publisher node publishing on a separate topic the guarded_move result. Is that OK or should I create another node?
- Once ground is found, the message with ground info is published only once. Is that OK? If not, how long should I publish it for? Shall I wait for a feedback from autonomy that confirms reception of the message and then publishing?
- @kmdalal, does that message look OK on the autonomy side?
- @Samahu can you give me feedback on how the code is written? Given the existing code, did I choose the proper way of publishing on a new topic?

### DIFF
--- a/ow_lander/msg/GuardedMoveResult.msg
+++ b/ow_lander/msg/GuardedMoveResult.msg
@@ -1,5 +1,3 @@
-bool success   # true if the ground has been successfully found
-float32 X      # ground location coordinates
-float32 Y
-float32 Z
-string frame   # reference frame that ground coordinates refer to
+bool success                 # true if the ground has been successfully found
+geometry_msgs/Point32 point  # X,Y,Z location coordinates
+string frame                 # reference frame that ground coordinates refer to
--- a/ow_lander/msg/GuardedMoveResult.msg
+++ b/ow_lander/msg/GuardedMoveResult.msg
@@ -1,3 +1,3 @@
-bool success                 # true if the ground has been successfully found
-geometry_msgs/Point32 point  # X,Y,Z location coordinates
-string frame                 # reference frame that ground coordinates refer to
+bool success                    # true if the ground has been successfully found
+geometry_msgs/Point32 position  # X,Y,Z location coordinates
+string frame                    # reference frame that ground coordinates refer to
--- a/ow_lander/msg/GuardedMoveResult.msg
+++ b/ow_lander/msg/GuardedMoveResult.msg
@@ -1,0 +1,5 @@
+bool success   # true if the ground has been successfully found
+float32 X      # ground location coordinates
+float32 Y
+float32 Z
+string frame   # reference frame that ground coordinates refer to
--- a/ow_lander/scripts/trajectory_publisher.py
+++ b/ow_lander/scripts/trajectory_publisher.py
@@ -5,8 +5,9 @@
 # this repository.
 
 import rospy
-import tf2
+import tf2_ros
 from std_msgs.msg import Float64
+from geometry_msgs.msg import Point
 from sensor_msgs.msg import JointState
 from ow_lander.srv import *
 import csv
@@ -137,33 +138,25 @@ def talker(req):
   if guarded_move_bool : # If the activity is a guarded_move
 
     # Start publisher for guarded_move result message
-    pubs.append(rospy.Publisher('/guarded_move_result', GuardedMoveResult, queue_size=10))
+    guarded_move_pub = rospy.Publisher('/guarded_move_result', GuardedMoveResult, queue_size=10)
 
     # Start subscriber for the joint_states
     rate = rospy.Rate(int(pub_rate/2)) # Hz
     rospy.Subscriber("/joint_states", JointState, joint_states_cb)
     time.sleep(2)
-    start_guard_delay_acc = 0
     for row in guard_rows[1:]: # Cycles on all the rows except header
       if row[0][0] == '1' : # If the row is a command
         ground = check_for_contact(velocity_array)
         if (ground == 1):
           print "Found ground, stopped motion..."
 
-          listener = tf.TransformListener()
-          listener.waitForTransform("/base_link", "/l_scoop_tip", rospy.Time(0), rospy.Duration(10.0))
-          (trans,rot) = listener.lookupTransform("base_link", "l_scoop_tip", rospy.Time(0))
-          pubs[6].publish(ground, trans[0], trans[1], trans[2], 'base_link')
-
-          # msg = GuardedMoveResult()
-          # msg.success = ground
-          # msg.X = trans[0]
-          # msg.Y = trans[1]
-          # msg.Z = trans[2]
-          # msg.frame = 'base_link'
-          # print(trans)
-          # print(msg)
-          # print(GuardedMoveResult)
+          tfBuffer = tf2_ros.Buffer()
+          listener = tf2_ros.TransformListener(tfBuffer)
+          trans = tfBuffer.lookup_transform("base_link", "l_scoop_tip", rospy.Time(0), rospy.Duration(10.0))
+          Point_X = trans.transform.translation.x
+          Point_Y = trans.transform.translation.y
+          Point_Z = trans.transform.translation.z
+          guarded_move_pub.publish(ground, Point(Point_X, Point_Y, Point_Z), 'base_link')
 
           return True, "Done publishing guarded_move"
 
@@ -172,11 +165,7 @@ def talker(req):
         print "Guarded. Sent %s on joint[0] publisher"%(float("%14s\n"%row[12+0]))
         rate.sleep()
 
-    if (ground != 1):
-      listener = tf.TransformListener()
-      listener.waitForTransform("/base_link", "/l_scoop_tip", rospy.Time(0), rospy.Duration(10.0))
-      (trans,rot) = listener.lookupTransform("base_link", "l_scoop_tip", rospy.Time(0))
-      pubs[6].publish(ground, 0.0, 0.0, 0.0, 'base_link')
+    guarded_move_pub.publish(ground, Point(0.0, 0.0, 0.0), 'base_link')
 
   return True, "Done publishing trajectory"
 
--- a/ow_lander/scripts/trajectory_publisher.py
+++ b/ow_lander/scripts/trajectory_publisher.py
@@ -5,7 +5,7 @@
 # this repository.
 
 import rospy
-import tf
+import tf2
 from std_msgs.msg import Float64
 from sensor_msgs.msg import JointState
 from ow_lander.srv import *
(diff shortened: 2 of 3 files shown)


ow_lander/scripts/trajectory_publisher.py
```diff
@@ -98,30 +100,30 @@ def talker(req):
...
     guard_filename = filename
     prefix = "pre_"
     filename = prefix + filename
     guarded_move_bool = True
 
-  # Reading csv file 
-  with open(filename, 'r') as csvfile: 
-    # creating a csv reader object 
-    csvreader = csv.reader(csvfile) 
+  # Reading csv file
```
mogumbo, line 109: Is your editor modifying whitespace on its own? There are a lot of changes here that aren't really changes. It's better to avoid that if possible.

ow_lander/scripts/trajectory_publisher.py
```diff
@@ -131,24 +133,50 @@ def talker(req):
       print "Sent %s on joint[0] publisher"%(float("%14s\n"%row[12+0]))
       rate.sleep()
 
-        
+
   if guarded_move_bool : # If the activity is a guarded_move
+
+    # Start publisher for guarded_move result message
+    pubs.append(rospy.Publisher('/GUARDED_MOVE_RESULT_W0OOOOHHHHHHHH', GuardedMoveResult, queue_size=10))
```
mogumbo, line 140: Yes, please change this name to something not ridiculous. In general, it's better to make all the changes you know you need to make before starting the code review.

ow_lander/scripts/trajectory_publisher.py
```diff
@@ -131,24 +133,50 @@ def talker(req):
...
     time.sleep(2)
-    start_guard_delay_acc = 0 
+    start_guard_delay_acc = 0
     for row in guard_rows[1:]: # Cycles on all the rows except header
       if row[0][0] == '1' : # If the row is a command
-        ground = check_for_contact(velocity_array)  
-        if (ground == 1):     
+        ground = check_for_contact(velocity_array)
+        if (ground == 1):
             print "Found ground, stopped motion..."
```
mogumbo, line 151: Too much indentation here.

ow_lander/scripts/trajectory_publisher.py
```diff
@@ -131,24 +133,50 @@ def talker(req):
...
             return True, "Done publishing guarded_move"
 
         for x in range(nb_links):
           pubs[x].publish(float("%14s\n"%row[12+x]))
         print "Guarded. Sent %s on joint[0] publisher"%(float("%14s\n"%row[12+0]))
-        rate.sleep()        
+        rate.sleep()
+
+    if (ground != 1):
+        listener = tf.TransformListener()
```
mogumbo, line 176: Too much indentation here too.

mogumbo:
This code looks alright to me. I'm a little concerned about putting special-case code in the trajectory publisher node, but there's not much of it yet. It would probably be better to think about refactoring that after we have converted to actions and after a little more special-case code has accumulated.

ow_lander/scripts/trajectory_publisher.py
damianacatanoso: Mhm, I am not sure how that happened...

ow_lander/scripts/trajectory_publisher.py
damianacatanoso: Yes, I am changing it, I just wanted to start the review soon, before our next meeting, in case you had constructive feedback to share during the meeting

ow_lander/scripts/trajectory_publisher.py
```diff
@@ -131,24 +133,50 @@ def talker(req):
...
+
+            # msg = GuardedMoveResult()
+            # msg.success = ground
+            # msg.X = trans[0]
+            # msg.Y = trans[1]
+            # msg.Z = trans[2]
+            # msg.frame = 'base_link'
+            # print(trans)
+            # print(msg)
+            # print(GuardedMoveResult)
```
Samahu, line 166: Is this for your own debugging? If so no need to push to it, git allows you to commit and parts of the file.

ow_lander/scripts/trajectory_publisher.py
damianacatanoso: No, for some reason this file had mixed indentations before I started working on it. I didn't realize it and kept the existing (wrong indentation) for the modifications I did to the code.

ow_lander/msg/GuardedMoveResult.msg
```diff
@@ -0,0 +1,5 @@
+bool success   # true if the ground has been successfully found
+float32 X      # ground location coordinates
+float32 Y
+float32 Z
```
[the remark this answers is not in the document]
damianacatanoso: Can I use a pre-defined message in the definition of a new message?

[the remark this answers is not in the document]
Samahu: Yes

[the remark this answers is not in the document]
Samahu: Just make sure to add the geometry_msgs to the lander package and you need probably a dependency in CMakeLists.txt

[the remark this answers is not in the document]
mogumbo: That's a good idea to keep things standard. There are a lot of examples of messages including messages in /opt/ros/melodic/share/

ow_lander/scripts/trajectory_publisher.py
```diff
@@ -131,24 +133,50 @@ def talker(req):
       print "Sent %s on joint[0] publisher"%(float("%14s\n"%row[12+0]))
       rate.sleep()
 
-        
+
   if guarded_move_bool : # If the activity is a guarded_move
+
+    # Start publisher for guarded_move result message
+    pubs.append(rospy.Publisher('/guarded_move_result', GuardedMoveResult, queue_size=10))
+
```
Samahu, line 141: I suggest that you move this piece to the top along with the other publishers

ow_lander/scripts/trajectory_publisher.py
```diff
@@ -131,24 +133,50 @@ def talker(req):
...
     for row in guard_rows[1:]: # Cycles on all the rows except header
       if row[0][0] == '1' : # If the row is a command
-        ground = check_for_contact(velocity_array)  
-        if (ground == 1):     
+        ground = check_for_contact(velocity_array)
+        if (ground == 1):
             print "Found ground, stopped motion..."
+
+            listener = tf.TransformListener()
+            listener.waitForTransform("/base_link", "/l_scoop_tip", rospy.Time(0), rospy.Duration(10.0))
```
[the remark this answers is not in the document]
damianacatanoso: As far as I understood, this waits for a maximum of 10 seconds for that transform to be published. If the transform arrives after 1 second, it moves forward, it doesn't wait for 9 more seconds. If the transform is not published within 10 seconds, it moves forward after 10 seconds. I think this is pretty reasonable.

ow_lander/scripts/trajectory_publisher.py
damianacatanoso: I put it there because this particular publisher should exist only if the published trajectory refers to guarded_move, while the other publishers exist for every trajectory.

damianacatanoso:
> I'm unable to test this feature as described:
> 
> * The service /planning/arm/publish_trajectory is not in 'rosservice list'.  I tried running it anyway, from the command line as well as the autonomy node, and it is not available.
> * /guarded_move_result is not in 'rostopic list'.
> 
> I have done a clean build of ow_lander on branch OW-434...

I did some changes. You may try now.

damianacatanoso:
I believe I addressed all your comments in my new commit. You can review again now.

kmdalal:
> > I'm unable to test this feature as described:
> > 
> > * The service /planning/arm/publish_trajectory is not in 'rosservice list'.  I tried running it anyway, from the command line as well as the autonomy node, and it is not available.
> > * /guarded_move_result is not in 'rostopic list'.
> > 
> > I have done a clean build of ow_lander on branch OW-434...
> 
> I did some changes. You may try now.

I retried, and publish_trajectory is available, and it works. 

The new topic shows up in 'rostopic list' only after publish_trajectory is called.  Is this expected?

And, using the service calls in your instructions, it fails:

$ rostopic echo /guarded_move_result
WARNING: topic [/guarded_move_result] does not appear to be published yet
WARNING: no messages received and simulated time is active.
Is /clock being published?
success: False
position: 
  x: 0.0
  y: 0.0
  z: 0.0
frame: "base_link"
---

damianacatanoso:
> > > I'm unable to test this feature as described:
> > > 
> > > * The service /planning/arm/publish_trajectory is not in 'rosservice list'.  I tried running it anyway, from the command line as well as the autonomy node, and it is not available.
> > > * /guarded_move_result is not in 'rostopic list'.
> > > 
> > > I have done a clean build of ow_lander on branch OW-434...
> > 
> > 
> > I did some changes. You may try now.
> 
> I retried, and publish_trajectory is available, and it works.
> 
> The new topic shows up in 'rostopic list' only after publish_trajectory is called. Is this expected?
> 
> And, using the service calls in your instructions, it fails:
> 
> ## $ rostopic echo /guarded_move_result
> WARNING: topic [/guarded_move_result] does not appear to be published yet
> WARNING: no messages received and simulated time is active.
> Is /clock being published?
> success: False
> position:
> x: 0.0
> y: 0.0
> z: 0.0
> frame: "base_link"

Hey Mike, what you described is actually the expected behavior: 
1) the topic is published while publish_trajectory is running, and 
2) no message is published on the topic before the ground is found. The message is published only after the ground is found, if it is found, or after the entire planned trajectory is published, if no ground has been found. 
Would this work on autonomy side? If not, could you describe how you envision this? I can modify it to better fit autonomy's needs!

kmdalal:
Damiana, your specification sounds fine, except:
1) What is a reasonable timeout for finding the ground (I need to implement this in autonomy)?
2) Why did this test fail (return false) and 0,0,0 for ground?

ow_lander/scripts/trajectory_publisher.py
```diff
@@ -131,24 +134,38 @@ def talker(req):
...
-        ground = check_for_contact(velocity_array)  
-        if (ground == 1):     
-            print "Found ground, stopped motion..."
-            return True, "Done publishing guarded_move"
+        ground = check_for_contact(velocity_array)
+        if (ground == 1):
+          print "Found ground, stopped motion..."
+
+          tfBuffer = tf2_ros.Buffer()
+          listener = tf2_ros.TransformListener(tfBuffer)
```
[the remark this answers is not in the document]
mogumbo: Agreed. This is an important change request.

damianacatanoso:
> Damiana, your specification sounds fine, except:
> 
> 1. What is a reasonable timeout for finding the ground (I need to implement this in autonomy)?

Before answering your question, Let me clarify how publish_trajectory works: 
- When calling services relative to the planning of digging or grinding (DigCircular, etc.), the trajectory in joint space is written and stored in a .csv file. publish_trajectory reads the rows of this .csv file, containing the desired joint angle values per time-step, and publishes them so Gazebo can execute this trajectory. 
- If there is no ground, publish_trajectory will go through the entire .csv file, publish all the rows in it, and after publishing the last one, it will output the message "false 0 0 0 base_link".
Now, to answer your question:
- The timeout is equal to the time that publish_trajectory takes to read the entire .csv file. 
- However, this time changes with the inputs of the guarded_move service itself. In particular, the timeout is influenced by: x,y,z coordinates of the starting point, and search_distance. 
- I can look into how to estimate the timeout for each service call, using the scoop velocity and trajectory. Figuring this out can take some time. So my suggestion is: if you can find a way to avoid the need of the timeout variable in the autonomy code, it would be nice. If it turns out to be a priority, I will create a ticket and work on it. Let me know!

> 2. Why did this test fail (return false) and 0,0,0 for ground?
- If you run this test and you got a fail message, maybe you had scoop-terrain collision disabled. This means that the scoop did penetrate the terrain and didn't find anything.

kmdalal:
Thanks for the detailed reply, Damiana.    A rough estimate of time needed would be adequate for a timeout, and I could implement the formula in autonomy if I know it.  It is outside the scope of this ticket.   

How does one enable scoop-terrain collision, and is this relevant or possible without the Gazebo client?

Samahu:
@kmdalal To enable scoop collision: comment out the <collision> section beginning on Line **542** of `ow_simulator/ow_lander/urdf/lander.xacro`.

damianacatanoso:
> Thanks for the detailed reply, Damiana. A rough estimate of time needed would be adequate for a timeout, and I could implement the formula in autonomy if I know it. It is outside the scope of this ticket.
> 
> How does one enable scoop-terrain collision, and is this relevant or possible without the Gazebo client?

Yes, this here is how that block of code looks like when it's commented out:

damianacatanoso:
> Thanks for the detailed reply, Damiana. A rough estimate of time needed would be adequate for a timeout, and I could implement the formula in autonomy if I know it. It is outside the scope of this ticket.
> 
> How does one enable scoop-terrain collision, and is this relevant or possible without the Gazebo client?

Ok, I'll create a ticket for that!

PR #6, opened 2020-08-25
Oceanwater 452 use atacama terrain

mogumbo:


Samahu:
Why the lander is not near the center of the map?
image
Is this intended?

mogumbo:
Yes, I put it there to make more room for the camera to take pictures in and around the workspace and for the arm to work. The arm probably can't reach all the way to the edge if the lander is in the center, so it probably isn't that important.

PR #7, opened 2020-08-25
Added contribution instructions.

mogumbo:
If these instructions are acceptable I think I should add them to each repo, customizing the link to bug fixes and feature requests for each repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,9 +40,13 @@ license or copyright restrictions. See our [license](LICENSE.txt) for more
 details.
 
 If you wish to contribute code or a bug fix please:
-- fork this repository,
-- make changes to your fork, and
-- create a pull request against this repository.
+- Create your own fork of this repository. In the upper-right corner of the
+[ow_simulator front page](https://github.com/nasa/ow_simulator) click `Fork`.
+Your fork will be called `<your_username>/ow_simulator`.
+- In your newly forked repository, create a branch with an appropriate name for
+your feature or bug fix.
+- Make changes to your new branch.
+- Create a pull request against the `master` branch of `nasa/ow_simulator`.
 
 ## Citation
 Coming soon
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -31,10 +31,18 @@ setting up the rest of the OceanWATERS repositories:
...
-Please review [current bugs and features requests](https://github.com/nasa/ow_simulator/issues) before submitting a new one.
+Contributions must be the original work of the contributor with no conflicting
+license or copyright restrictions. See our [license](LICENSE.txt) for more
+details.
 
-If you wish to contribute code or a bug fix please follow our [Contribution Guidlines](todo).  
+If you wish to contribute code or a bug fix please:
+- fork this repository,
+- make changes to your fork, and
+- create a pull request against this repository.
```
kmdalal, line 45: This doesn't seem completely clear.  This would be more clear to me:
* Fork a new branch from the 'master' branch of this repository,
* Make changes to your branch, and
* Create a pull request for the new branch.

Assuming this is what's meant.

mogumbo: I think a person outside our group needs to fork the entire repository before creating a new branch on their own fork of the repository. But creating a new branch might also be good practice and something we want to add to this.

kmdalal: Until now, I did not know anything about _forking_ a repo vs cloning it.  It's something I'll need to study more -- a few short descriptions on the net didn't make it clear.  I think we should describe a more detailed workflow, that makes a clear case for forking, and probably includes branches.   The workflow should be reasonable for both us and contributors, and hopefully avoid more tricky operations such as rebasing.

mogumbo: I made some refinements. Tell me what you think now....  The workflow is different for people on the OceanWATERS team because we can push directly to the original NASA repositories; these instructions target external contributors. I'm not a git expert, but I don't know how rebasing would be involved in this.

kmdalal: Much better, thanks.  I'll need to better understand forking and the different workflow, but I think this looks good for a first version.

PR #8, opened 2020-09-02
Tuned PID, tried to eliminate jerking motion

damianacatanoso:
**Jira ticket:**
https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-410

**Problem:**
- Arm jerking motion, which appears during some of the maneuvers. 

**Background:**
- We thought the arm jerking motion was caused by wrong tuning of PID. While tuning PID I saw that oscillations appeared in the published trajectory, which is stored in the trajectory .csv file after planning. The trajectory in the .csv file shouldn't have oscillations. So I investigated why this is happening.
- It turns out that: MoveIt!, to calculate the trajectory in the .csv file, uses these "fake controllers". The fake controller takes the plan, calculated by the given planner (RRTconnect for instance) and creates a sequence of points that the arm should follow to fulfill the plan. These points are stored in the .csv. There are three different types of fake controllers, among which we need to use the one called "interpolate". Interpolated has a frequency that needs to be set, that we didn't set. The frequency tells how many points per second we want to have in the trajectory file. Tuning this frequency allows to smooth the trajectory, eliminating the oscillations. 

**Approach:**
- Re-tuned PID gains
- Changed MoveIt! fake controller frequency value to 2.5 Hz
- Reduced speed for a specific part of the motion in guarded_move that was subject to vibrations

**Some comments:**
- Sometimes the arm still experiences vibrations. However, it's not as often as before and they are much smaller and imperceptible 
-  Anjan is looking into how to bypass the .csv file to make the planning faster. If he finds a way to do that, the fake controller will most probably not be called. Since the fake controller is causing the problem, I believe the arm will no longer experience this kind of jerk motion.

**Test:**
- Please run the services belonging to the Digging Procedure, in the given order, followed by publish_trajectory. Look at the arm and check for vibrations/jerking motion. Mike, you can look at the solid arm in rviz.

Questions for you:
- Do you see the jerk motion? Where?

### DIFF
--- a/ow_lander/config/fake_controllers.yaml
+++ b/ow_lander/config/fake_controllers.yaml
@@ -1,5 +1,7 @@
+rate: 2.5 
 controller_list:
   - name: fake_arm_controller
+    type: interpolate
     joints:
       - j_shou_yaw
       - j_shou_pitch
--- a/ow_lander/config/lander_control.yaml
+++ b/ow_lander/config/lander_control.yaml
@@ -1,25 +1,25 @@
 # Publish all joint states -----------------------------------
 joint_state_controller:
   type: joint_state_controller/JointStateController
-  publish_rate: 50
+  publish_rate: 100
 
 # Position Controllers ---------------------------------------
 shou_yaw_position_controller:
   type: effort_controllers/JointPositionController
   joint: j_shou_yaw
-  pid: {p: 400.0, i: 10.0, d: 100.0}
+  pid: {p: 1000.0, i: 10.0, d: 300.0}
 shou_pitch_position_controller:
   type: effort_controllers/JointPositionController
   joint: j_shou_pitch
-  pid: {p: 500.0, i: 0.2, d: 15.0}
+  pid: {p: 2000.0, i: 0.2, d: 1500.0}
 prox_pitch_position_controller:
   type: effort_controllers/JointPositionController
   joint: j_prox_pitch
-  pid: {p: 2000.0, i: 0.2, d: 100.0}
+  pid: {p: 1100.0, i: 2.0, d: 100.0}
 dist_pitch_position_controller:
   type: effort_controllers/JointPositionController
   joint: j_dist_pitch
-  pid: {p: 200.0, i: 0.06, d: 10.0}
+  pid: {p: 700.0, i: 0.06, d: 20.0}
 hand_yaw_position_controller:
   type: effort_controllers/JointPositionController
   joint: j_hand_yaw
--- a/ow_lander/config/lander_control.yaml
+++ b/ow_lander/config/lander_control.yaml
@@ -11,7 +11,7 @@ shou_yaw_position_controller:
 shou_pitch_position_controller:
   type: effort_controllers/JointPositionController
   joint: j_shou_pitch
-  pid: {p: 2000.0, i: 0.2, d: 1500.0}
+  pid: {p: 2000.0, i: 0.2, d: 150.0}
 prox_pitch_position_controller:
   type: effort_controllers/JointPositionController
   joint: j_prox_pitch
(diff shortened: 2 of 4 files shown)


Samahu (changes requested):
This change didn't work for me.
* arm collides with lander leg again 
[arm collides with lander leg again](https://youtu.be/twdDhzRyLDQ)

* arm improperly falls to ground in dig circular
[arm improperly falls to ground in dig circular](https://youtu.be/3UCIx2IYwx8)

* dig linear is not better; arm collides with lander leg and goes prematurely above ground during the dig motion
[dig linear is not better](https://youtu.be/fTxYCMCw8yk)

damianacatanoso:
> This change didn't work for me.
> 
> * arm collides with lander leg again
>   [arm collides with lander leg again](https://youtu.be/twdDhzRyLDQ)
> * arm improperly falls to ground in dig circular
>   [arm improperly falls to ground in dig circular](https://youtu.be/3UCIx2IYwx8)
> * dig linear is not better; arm collides with lander leg and goes prematurely above ground during the dig motion
>   [dig linear is not better](https://youtu.be/fTxYCMCw8yk)

This looks terrible, I am looking into it, thanks for the feedback. Just as a note, the digging/drilling services are called after the "unstow" service.

kmdalal (approved):
I tried GuardedMove, DigCircular, and DigLinear (once each) and saw no jerky movements like there were before.  Nice work!

Samahu:
I didn't know that you expect to perform unstow first. So I did another test with dig circular after unstow, see the video below.

This addresses the arm collision with lander leg but I think the drop of the arm hand is still noticeable.
[end effector seem to drop](https://youtu.be/4CrUlV__udA)

kmdalal:
After testing and approving this PR, I tried another Guarded Move after enabling scoop/terrain collision (by commenting OUT the 'collision' element in lander.xacro), and I got a slight arm shake at the end of the movement, that did not stop.  Less severe than what I used to see, but noticable.

damianacatanoso:
> > This change didn't work for me.
> > 
> > * arm collides with lander leg again
> >   [arm collides with lander leg again](https://youtu.be/twdDhzRyLDQ)
> > * arm improperly falls to ground in dig circular
> >   [arm improperly falls to ground in dig circular](https://youtu.be/3UCIx2IYwx8)
> > * dig linear is not better; arm collides with lander leg and goes prematurely above ground during the dig motion
> >   [dig linear is not better](https://youtu.be/fTxYCMCw8yk)
> 
> This looks terrible, I am looking into it, thanks for the feedback. Just as a note, the digging/drilling services are called after the "unstow" service.

I realized there was a typo in one of the gains. Shou_pitch's "d" was 1500, while the correct value is 150.

damianacatanoso:
Could you try again now?

mogumbo (approved):

Samahu:
Hi Damiana,

I did a quick test with your latest against the dig linear. It looks better than before. However, there is still notice the arm shakes half way through the digging:

[small but somewhat noticeable shake](https://youtu.be/UdgBykaueTg)

Is this behavior expected? If so I think you may go ahead and merge

damianacatanoso:
> Hi Damiana,
> 
> I did a quick test with your latest against the dig linear. It looks better than before. However, there is still notice the arm shakes half way through the digging:
> 
> [small but somewhat noticeable shake](https://youtu.be/UdgBykaueTg)
> 
> Is this behavior expected? If so I think you may go ahead and merge

I think we should accept this at the moment. We are planning to discard the linear trenching for the first software release so it shouldn't be a big deal. Once we find a way to bypass the fake controller in the trajectory publisher, I'll tune the gains better. If do it now, we'll have vibrations in other services for the reasons I explained yesterday.

Samahu (approved):

PR #11, opened 2020-09-09
Modified dig_linear, changed default terrain height

damianacatanoso:
1) Did some modifications to dig_linear because it wouldn't go to the target depth. 
2) Reduced default depth because doing experiments with guarded_move for Ussama's review I realized the given default terrain depth was too high.

I am selecting only Terry as reviewer because these are small changes.

To test, plan and publish dig_linear. The scoop should go underground. I suggest to disable collision for the scoop to avoid problems in visualizing this caused by dynamic terrain. Thanks!

### DIFF
--- a/ow_lander/scripts/activity_full_digging_traj.py
+++ b/ow_lander/scripts/activity_full_digging_traj.py
@@ -132,12 +132,14 @@ def dig_linear(move_arm,move_limbs,args):
   #rotate dist pith to pre-trenching position.
   change_joint_value(move_arm, constants.J_DIST_PITCH, -math.pi/2)
 
-  ## Once aligned to trench goal, place hand above trench middle point
-  z_start = ground_position + constants.SCOOP_OFFSET - depth
-  go_to_Z_coordinate(move_limbs, x_start, y_start, z_start)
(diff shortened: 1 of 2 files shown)


ow_lander/scripts/activity_full_digging_traj.py
```diff
@@ -132,12 +132,14 @@ def dig_linear(move_arm,move_limbs,args):
   #rotate dist pith to pre-trenching position.
   change_joint_value(move_arm, constants.J_DIST_PITCH, -math.pi/2)
 
-  ## Once aligned to trench goal, place hand above trench middle point
-  z_start = ground_position + constants.SCOOP_OFFSET - depth
-  go_to_Z_coordinate(move_limbs, x_start, y_start, z_start)
+  ## Once aligned to trench goal, place hand above the desired start point
+  alpha = math.atan2(constants.WRIST_SCOOP_PARAL,constants.WRIST_SCOOP_PERP)
```
mogumbo, line 136: Try to be consistent with your whitespace. I find things to be more readable if there are spaces after commas and spaces around mathematical operators. You do it both ways here.

mogumbo (approved):

PR #12, opened 2020-09-10
Refine digging collision and deformation

Samahu:
## Summary of Changes:
+ Ensure the same digging values works for all terrains
+ Grinder only affects 1/3 of the value
+ Enable collision for scoop and grinder

## Verify
```bash
```
or
```bash
```
and then perform the grind operation followed by a scoop. Note that scooping now collides with the terrain.

## Objectives
* Check that the grind operation affects all the terrain in same manner (in particular atacama and terminator_workspace).
* Check that the Scoop operation no longer updates the collision.
* Check that the Scoop operation only updates the terrain visuals.
* Any suggestions or improvements that can be made to make this work better.

Linked Issue: https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-471

### DIFF
--- a/ow_lander/urdf/lander.xacro
+++ b/ow_lander/urdf/lander.xacro
@@ -543,6 +543,13 @@
       </plugin>
     </visual>
     <selfCollide>1</selfCollide>
+    <collision>
+      <surface>
+        <contact>
+          <collide_bitmask>0x0001</collide_bitmask>
+        </contact>
+      </surface>
+    </collision>
   </gazebo>
 
   <!-- link that we can reference in Gazebo to get altitude of tip
@@ -642,7 +649,7 @@
     <collision>
       <surface>
         <contact>
-          <collide_bitmask>0x0002</collide_bitmask>
+          <collide_bitmask>0x0001</collide_bitmask>
         </contact>
       </surface>
     </collision>
--- a/ow_lander/urdf/lander.xacro
+++ b/ow_lander/urdf/lander.xacro
@@ -546,7 +546,7 @@
     <collision>
       <surface>
         <contact>
-          <collide_bitmask>0x0001</collide_bitmask>
+          <collide_bitmask>0x0001</collide_bitmask> <!-- Enable collision with the terrain by matching the bitmask -->
         </contact>
       </surface>
     </collision>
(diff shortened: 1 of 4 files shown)


ow_lander/urdf/lander.xacro
```diff
@@ -642,7 +649,7 @@
     <collision>
       <surface>
         <contact>
-          <collide_bitmask>0x0002</collide_bitmask>
+          <collide_bitmask>0x0001</collide_bitmask>
```
mogumbo, line 652: You could just leave this block out to have collisions with terrain. It would probably be good to at least comment that this is enabling collisions with terrain.

ow_lander/urdf/lander.xacro
```diff
@@ -543,6 +543,13 @@
       </plugin>
     </visual>
     <selfCollide>1</selfCollide>
+    <collision>
```
mogumbo, line 546: You could just leave this block out to have collisions with terrain. It would probably be good to at least comment that this is enabling collisions with terrain.

mogumbo (approved):

ow_lander/urdf/lander.xacro
```diff
@@ -543,6 +543,13 @@
       </plugin>
     </visual>
     <selfCollide>1</selfCollide>
+    <collision>
+      <surface>
+        <contact>
+          <collide_bitmask>0x0001</collide_bitmask>
```
Samahu, line 549:
```suggestion
          <collide_bitmask>0x0001</collide_bitmask> <!-- Enable collision with the terrain by matching the bitmask -->
```

ow_lander/urdf/lander.xacro
Samahu, line 652:
```suggestion
          <collide_bitmask>0x0001</collide_bitmask> <!-- Enable collision with the terrain by matching the bitmask -->

```

ow_lander/urdf/lander.xacro
Samahu: I figured it would be much easier this rather than looking for the block in code history to find how to disable or enable collision with the terrain.

damianacatanoso (approved):
I tested dig_circular and grind. Dig_circular behaves correctly, the trench shape is the expected one. Grind doesn't behave as expected. Running the grind service with use_default=True, the terrain doesn't show changed, it behaves as the grinder-terrain collision was turned off. Increasing the "depth" parameter, input of "compose_modify_terrain_circle_message" in line 64 of modify_terrain_grinder_pub.py, to 0.1, I observe a shallow trench (see attached figure).
image
I called Ussama to spot the problem and we weren't able to reproduce this issue on his machine. It seems something isolated to my machine. So I will approve this request for now and leave this comment for the record.

PR #13, opened 2020-09-10
Detect ground based on multiple readings

Samahu:
This improves things a bit but I wouldn't call it robust
Not planning to merge this unless no better solution is found by next week


damianacatanoso:
I think you can close this PR, since your new working solution is merged.

Samahu:
> I think you can close this PR, since your new working solution is merged.

You are right, I will close it

PR #15, opened 2020-09-16
Renovated grind service

damianacatanoso:
Hi everyone, this PR is about the "grind" service and its new modifications. I addressed the following:
-  Added the new parameter "radial" to the service message. If radial==True, the grinder moves along a direction parallel to the arm; if radial==False, the grinder moves along a direction perpendicular to the arm. In both cases, the starting point is x,y.
-  Adjusted the depth calculation, including measured dimensions of the grinder.
-  Changed the motion of the grinder from single pass to double pass, that allows for a wider trench
-  Removed unnecessary arm resets at the beginning and end of the service
-  In ow_autonomy, updated the plan ReferenceMission1 to account for grind modifications

I have been experiencing issues visualizing the terrain deformation after grinding. For this reason, I haven't had a chance to check if the grind service, when called with radial=False, provides a proper trench for the "perpendicular to radial" scooping. So I would be glad to have a feedback on that too, as well as some help in solving this problem.

Test:
- Checkout the branch both for ow_simulator and ow_autonomy, do a clean build;
- Please, give me a feedback on how the grinding and digging are performed in both cases: radial=True and radial=False. To do this, call the following in sequence. Remember to call and publish Unstow first.
- rosservice call /planning/arm/grind False 1.65 0.0 0.1 0.6 True -0.155 False
- rosservice call /planning/arm/publish_trajectory True a
- rosservice call -- /planning/arm/dig_circular False 1.65 0.0 0.1 True -0.155 False
- rosservice call /planning/arm/publish_trajectory True a

- and then do the same with the third from the last parameter equal to False.

- Also, please run the ReferenceMission1 plan and give me feedback:
- In one terminal launch europa_terminator_workspace and in another terminal launch the plan/autonomy node:
$ roslaunch ow_autonomy autonomy_node.launch plan:=ReferenceMission1.plx


Samahu:
> Should open another PR for the corresponding branch on ow_autonomy? Is there a way to include that branch in this PR?

I recommend that open another PR on ow_autonomy to be able to refer to changes in the other branch.
However, in the PR description on ow_autonomy I would only add a link to this PR. Just to indicate that this is the main PR and we would use it for discussions that are not tied to a given file.

Samahu:
I did test all the scenario as indicated and all performed successfully. Before I can give a good feedback I need to understand the better between radial and non-radial movements (maybe when I start reviewing the code I will get the idea).

At one of the grinder digging movement I noticed that the grinder was not perpendicular to the surface (as it used to be under default settings). Is that expected/intentional?

damianacatanoso:
> I did test all the scenario as indicated and all performed successfully. Before I can give a good feedback I need to understand the better between radial and non-radial movements (maybe when I start reviewing the code I will get the idea).
> 
> At one of the grinder digging movement I noticed that the grinder was not perpendicular to the surface (as it used to be under default settings). Is that expected/intentional?

That's not intentional but expected. I've tried to impose constraints on the grinder, to keep it vertical, but the planner is not able to satisfy the constraint for some reason. This would need more time to investigate, that we don't have at the moment. Do you think we can accept this for now?

mogumbo (changes requested):
The only real problem I see with this code is doing a grind with default settings. The move gives an error before it finishes. This should be fixed such that a grind and dig_circular with default settings will work together.

I tried @damianacatanoso's instructions on both atacama_y1a and terminator_workspace. The results are the same with each. I also tried with depth values of 0.01, 0.1, and 0.2. The grinder barely touches the ground at 0.01. At 0.1, the grinder affects the collision model but not the visual model. At 0.2, both models are deformed. It would be good if the visual model were affected at at depth of 0.1, but that is a separate issue.

After grinding a trench of depth 0.2 and then doing a corresponding dig_circular at depth 0.2, the scoop removes some material from the lip of the far end of the trench. I expected the scoop to collide with this, but I guess the collision trench is not only deeper than the visual trench but also wider. Again, this is a separate issue.

Samahu:
> Do you think we can accept this for now?

I don't think I can answer this, however, if it is not possible to keep the grinder straight up while it is acting on the terrain then I need to adjust the limits when the terrain gets updated by the grinder.

damianacatanoso:
> > Do you think we can accept this for now?
> 
> I don't think I can answer this, however, if it is not possible to keep the grinder straight up while it is acting on the terrain then I need to adjust the limits when the terrain gets updated by the grinder.

Now that we pushed the code freeze to Tuesday, I am working on this problem. I'll update you!

ow_lander/scripts/activity_grind.py
```diff
@@ -27,32 +54,59 @@ def plan_cartesian_path_lin(move_arm, length):
 
 def grind(move_arm, move_limbs, args):
```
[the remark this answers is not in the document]
damianacatanoso: Created ticket for this: https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-484

ow_lander/scripts/activity_grind.py
```diff
@@ -7,16 +7,43 @@
...
+    depth = req.depth
+    length = req.length
+    radial = req.radial
+    ground_position = req.ground_position
+    delete_prev_traj = req.delete_prev_traj
+
+  return [req.use_defaults, x_start, y_start, depth, length, radial, ground_position, delete_prev_traj]
+
+
+def plan_cartesian_path(move_arm, length, alpha, radial):
```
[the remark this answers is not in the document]
damianacatanoso: Created ticket for this: https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-484

ow_lander/srv/Grind.srv
```diff
@@ -3,6 +3,7 @@ float32 x               # X coordinate of grinding starting point
 float32 y               # Y coordinate of grinding starting point
 float32 depth           # Desired depth
 float32 length          # Desired length
+bool radial             # If True, trenching is along radial direction. If False, perpendicular to radial
```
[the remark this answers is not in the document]
damianacatanoso: Changed radial with "parallel" in the new commit.

ow_lander/scripts/constants.py
```diff
@@ -21,9 +21,9 @@
 
 # Distance between scoop center of mass and lower blade
 SCOOP_HEIGHT = 0.076
-GRINDER_HEIGHT = 3*SCOOP_HEIGHT
+GRINDER_HEIGHT = 0.31
```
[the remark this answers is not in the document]
damianacatanoso: This parameter is no longer used, I eliminated it in the new commit.

ow_lander/scripts/activity_grind.py
```diff
@@ -27,32 +54,59 @@ def plan_cartesian_path_lin(move_arm, length):
...
+
 
-  # exiting terrain - along +Z
-  z_start = ground_position + constants.GRINDER_OFFSET - depth
-  go_to_Z_coordinate(move_arm, x_start, y_start, z_start)
+  # exiting terrain
+  joint_goal = move_arm.get_current_joint_values()
+  change_joint_value(move_arm, constants.J_SHOU_PITCH, joint_goal[1]+0.6)
 
+  move_arm.set_planner_id('RRTconnect')
```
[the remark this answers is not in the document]
damianacatanoso: Because otherwise the planner will keep using RRTstar for the entire session, until the sim is killed. However, this line is no longer required in the new commit.

Samahu:
Some how it seems that your latest changes may have altered RVIZ view, and I don't get to see the "Displays" window when I try to expand it, see attached image:

image

Samahu:
Just test the scenario non-Parallel Grind operation and I have noticed that the scoop goes inside the lander leg, see video:
[image](https://youtu.be/gEZcyPoLu_U)

damianacatanoso:
> Some how it seems that your latest changes may have altered RVIZ view, and I don't get to see the "Displays" window when I try to expand it, see attached image:
> 
> image

ok, let me change the Rviz fi

> Just test the scenario non-Parallel Grind operation and I have noticed that the scoop goes inside the lander leg, see linked video:
> 
> [image](https://youtu.be/gEZcyPoLu_U)

Grind service is always called after Unstow, sorry, I forgot to mention it in the instructions.

mogumbo:
Glad to see "radial" changed to "parallel." I don't understand why the grinder has to be a revolute joint instead of fixed. Can you please explain that?

damianacatanoso:
> Just test the scenario non-Parallel Grind operation and I have noticed that the scoop goes inside the lander leg, see linked video:
> 
> [image](https://youtu.be/gEZcyPoLu_U)

I updated the PR test instructions.

Samahu (changes requested):
Damiana, during a planning for non-parallel grind operation from an unstowed pose, the hand twists back wards for no clear reason. See the linked video:

[image](https://youtu.be/IytOorLmMf8)

I think this has to be understood and fixed before proceeding with this PR. It wastes time and energy and looks very odd to the observer.

Looking into this.

kmdalal:
I have tested this branch extensively over the past week, both from ow_autonomy (OCEANWATER-429 branch, in review) and through rqt.  I get the same behaviors already described above.  I have emailed Damiana with more details.

Samahu:
> Damiana, during a planning for non-parallel grind operation from an unstowed pose, the hand twists back wards for no clear reason. See the linked video:
> 
> [image](https://youtu.be/IytOorLmMf8)
> 
> I think this has to be understood and fixed before proceeding with this PR. It wastes time and energy and looks very odd to the observer.

Alternatively, if these changes are not considered critical we can approve this PR and address this issue (and any other) with hotfixes after we branch for the release today.

ow_lander/scripts/activity_full_digging_traj.py
```diff
@@ -36,7 +36,7 @@ def arg_parsing_lin(req):
     # Default trenching values
```
kmdalal, line 36: This comment applies to default values in _all_ lander services.  I have not understood why default values are provided at this low level at all.   I don't think there should be any default values at this level, nor any "use_defaults" argument (which I do not expose in autonomy).   I realize this would be a pervasive change and I'm not suggesting it for now.

If we _do_ provide defaults in services, it would be nice to know what they mean -- this should be in both code comments and user documentation.

Samahu: I agree

Samahu: Although I don't think there is any meaning beyond these default values. I think they were defined merely to help with testing. Such that the caller wouldn't have to remember these values every time. Clearly there are better ways to deal with this.

ow_lander/scripts/activity_grind.py
```diff
@@ -7,52 +7,123 @@
...
+    parallel = req.parallel
+    ground_position = req.ground_position
+    delete_prev_traj = req.delete_prev_traj
+
+  return [req.use_defaults, x_start, y_start, depth, length, parallel, ground_position, delete_prev_traj]
+
+
+def plan_cartesian_path(move_group, length, alpha, parallel):
+
+  if parallel==False:
```
kmdalal, line 38: Small style comment: for boolean variables such as this one, 'parallel' and '! parallel' are more succinct than 'parallel==True' and 'parallel==False'

ow_lander/scripts/activity_grind.py
```diff
@@ -7,52 +7,123 @@
...
                                0.0)         # jump threshold
+
   return plan, fraction
 
-def grind(move_arm, move_limbs, args):
+def grind(move_arm, move_limbs, move_grinder, args):
 
   x_start = args[1]
-  y_start = args[2]
+  y_start = args[2]
```
kmdalal, line 57: Small style comment: multi-assignment is a compact alternative: x_start, y_start = args[1], args[2]

ow_lander/scripts/activity_grind.py
```diff
@@ -7,52 +7,123 @@
...
-  # rotate hand
-  change_joint_value(move_arm, constants.J_HAND_YAW, -2*math.pi/3)
+  goal_pose = move_grinder.get_current_pose().pose
+  goal_pose.position.x = x_start
+  goal_pose.position.y = y_start
+  goal_pose.position.z = 0.25 
+  goal_pose.orientation.x = 0.70616885803
+  goal_pose.orientation.y = 0.0303977418722
+  goal_pose.orientation.z = -0.706723318474
+  goal_pose.orientation.w = 0.0307192507001
```
kmdalal, line 90: Literal numbers used in code are sometimes termed "magic numbers".  It's good to either abstract them with named constants, or add a comment explaining why these values were chosen.

ow_lander/scripts/activity_full_digging_traj.py
Samahu: btw, there is already a jira ticket for getting rid of **use_defaults**, https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-447.

Additionally, I have create another related ticket https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-483 to get rid of the delete_prev_proj our API.

I can implement this within a day but I am not sure if this of a priority to us at the current moment.

ow_lander/scripts/constants.py
```diff
@@ -11,19 +11,19 @@
...
 SCOOP_OFFSET = 0.215
-GRINDER_OFFSET = 0.8
+GRINDER_OFFSET = 0.16
 
 # Distance between scoop center of mass and lower blade
 SCOOP_HEIGHT = 0.076
-GRINDER_HEIGHT = 3*SCOOP_HEIGHT
 
-DEFAULT_GROUND_HEIGHT = -0.135
+DEFAULT_GROUND_HEIGHT = -0.155
```
kmdalal, line 26: If this default applies to a particular terrain, or region of terrain, is the code safe?  See my earlier comment cautioning against specifying defaults in services.

damianacatanoso:
> Glad to see "radial" changed to "parallel." I don't understand why the grinder has to be a revolute joint instead of fixed. Can you please explain that?
To plan movements of end effectors, MoveIt uses the so called “groups”. As of now, we have three motion groups: antenna, limbs and arm. For the arm motion planning, limbs and arm are of interest. The limbs group has the hand as end effector, while the arm group has the scoop as end effector. Until now, the grinder has been considered an “extension” of the hand with a fixed joint. Imposing a motion to the grinder was achieved imposing a motion on the hand. Considering the hand+grinder as a rigid body inhibits certain movements of the grinder, which can be achieved through 1) allowing the grinder to bend, as it was happening in the previous commit, or 2) allowing relative rotation between the hand finger and the grinder body. I tried imposing an orientation constraint on the grinder to keep it vertical but I found that constraints can be imposed only on end effectors, which in our case is the hand (limbs) or scoop (arm), otherwise the planner ignores them. So the right thing to do was to create a new motion group, called “grinder”, that has the grinder as end effector, such that I can directly command motion on the grinder and, if needed, impose constraints as well. After creating the new group “grinder”, keeping the j_grinder joint type fixed, I needed to impose Cartesian (linear) motion on the end effector (grinder). Cartesian motion is achieved providing the end effector with a series of pose (position+orientation) points that the grinder has to go through. This series of poses is the xyz trajectory of the center of mass + the orientation of the grinder. The orientation shouldn’t change during the linear motion so the orientation part of the pose is the same, while the position part of the points follows the desired straight line. With a grinder fixed joint, the planner wouldn’t find a solution and the desired linear motion wasn’t possible. At this point my high-level thought was: we want the grinder to stay vertical during the linear motion, which means that we are adding a constraint to the grinder, which means that we are removing one degree of freedom to the system. If we force the grinder to go linear, the lack of this degree of freedom will appear as the “bending”. If we force the grinder to stay vertical, the lack of degree of freedom translates into impossibility of performing clean linear motion. So the solution was to give that degree of freedom back to the system, allowing for rotation between the grinder and the hand, which means introducing a revolute joint.

ow_lander/srv/DigCircular.srv
```diff
@@ -2,7 +2,7 @@ bool use_defaults       # Set to true to use the default trench location
 float32 x               # Trench starting point - X coordinate
 float32 y               # Trench starting point - Y coordinate
 float32 depth           # Desired depth
-bool radial             # If True, trenching is along radial direction. If False, perpendicular to radial
+bool parallel           # If True, resulting trench is parallel to arm. If False, perpendicular to arm
```
kmdalal, line 5: Thanks!

ow_lander/srv/Grind.srv
```diff
@@ -3,6 +3,7 @@ float32 x               # X coordinate of grinding starting point
 float32 y               # Y coordinate of grinding starting point
 float32 depth           # Desired depth
 float32 length          # Desired length
+bool parallel           # If True, resulting trench is parallel to arm. If False, perpendicular to arm
 float32 ground_position # Z coordinate of ground level in base_link frame
 bool delete_prev_traj   # Set to true to cleanup ~/.ros
```
kmdalal, line 8: This strikes me as a dangerous parameter, one that should be replaced with a safer way to manage ~/.ros and trajectories.   Recall the bug we had in which a service call that passed true here inadvertently deleted another service's trajectory that was still needed.   Please add a new Jira to address this, unless it already exists.

ow_lander/scripts/constants.py
damianacatanoso: This is the terrain height found by guarded_move when running it with use_defaults=True in europa_terminator_workspace. I am not sure how safe it is to use it for other locations in europa_terminator, it might be OK since the terrain is not very irregular. It's definitely not safe to use it in atacama_y1a, since the terrain is very irregular there. However, autonomy shouldn't run the grinding service with use_defaults=True. Autonomy should know the ground height in that specific location and terrain, and use it to call the service.

ow_lander/srv/Grind.srv
Samahu: I think you are referring to the **delete_prev_traj**. This actually what happened to me, I thought it would delete only the previous trajectory instead it seems to delete all previous trajectories. Hence, I have added the ticket [API Improvements: Remove delete_prev_traj parameter from all of the services](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-483)

damianacatanoso: Here it is: https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-486

ow_lander/scripts/activity_grind.py
damianacatanoso: You are right! These numbers describe a specific pose (position+orientation) that I want the grinder to have in order to penetrate the terrain vertically and not diagonally as it was before. What do you think is best: comments or putting them in constants.py ?

ow_lander/include/ow_lander/lander_joints.h
```diff
@@ -18,21 +18,22 @@ typedef enum {
...
-  "j_ant_pan", "j_ant_tilt"
+  "j_ant_pan", "j_ant_tilt", "j_grinder"
 };
 
 // These names are good for displaying to humans
 const std::vector<std::string> joint_display_names {
   "Shoulder Yaw", "Shoulder Pitch", "Proximal Pitch", "Distal Pitch", "Hand Yaw", "Scoop Yaw",
-  "Antenna Pan", "Antenna Tilt"
+  "Antenna Pan", "Antenna Tilt", "Grinder Yaw"
 };
```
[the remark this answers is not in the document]
damianacatanoso: OK!

ow_lander/scripts/activity_grind.py
kmdalal: If they are only in this file, comments.  If used by other files, in constants.py.

mogumbo (approved):

Samahu (approved):

kmdalal (approved):

PR #16, opened 2020-09-16
Working solution for ground detection

Samahu:
## The approach
The approach is fairly straight forward, it is based off monitoring the slight bouncing of the arm that occurs when the arm touches the ground. This worked fine for all of the configurations that I have tested and even multiple times per session. I have tested this implementation against the **quick** solver and **world** solver, both worked fine.  
To make sure the solution works even if no bouncing does not occur (due to a give lander/arm physics setup), I am planning to add solution based of the trend of _scoop z deltas_ as it moves downwards rather than using single delta (I think I will address this as a separate PR).

## Summary of Changes
* Removed the peak detection approach
* Implemented a detection based on the arm bouncing off ground
  - WIP: Provide a configurable implementation that uses the trend of delta values rather than a single delta value

## Verify
Start by launch the simulation against **the terminator workspace** 
```bash
```
or the **atacama terrain**
```bash
```

Perform planning for the arm **guarded move** and publish the trajectory

## Objectives 
* Verify that the arm is being detected around the proper altitude
* Try different configurations of the simulation that you think may affect the performance
  - for example try changing the trajectory publish rate, physics solver, joint limits, ... etc
* Review the code and provide suggestions

__Linked Jira Issues:__
* https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-461
* https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-472


ow_lander/scripts/trajectory_publisher.py
```diff
@@ -18,27 +19,39 @@
...
+# over a short term of position z value it would probably make the implementation
+# more robust against an arm physics configuration where the end effector doesn't
+#  bounce off the ground at all.
 
-lag = 80
-threshold = 6.3
-influence = 0.5
-peak_detection = PeakDetectionRT(lag=lag, threshold=threshold, influence=influence)
+# constanst
+GROUND_DETECTION_THRESHOLD = 0.001  # TODO: tune this value further to improve the detection
```
mogumbo, line 28: This should probably factor in timestep in case we want to run the sim at different speeds in the future. For example GROUND_DETECTION_THRESHOLD = 0.5 * timestep

This effectively makes the detection based on velocity. Someone correct me if I'm wrong: I thought we switched away from velocity because it's an indirect measurement we can only estimate in real-life using joint states. That said, I'm fine with "cheating" and using ground truth velocity if its the only way forward in the short term.

mogumbo (changes requested):
This tests well for me. Can we safely speed up the ground search while we're at it by changing activity_guarded_move.py Line 107? @damianacatanoso, is this safe? I think you have experience with it.

ow_lander/scripts/trajectory_publisher.py
Samahu: This solution is already effectively based on velocity since I am basing my decision on the delta of two positions.

With regard to multiplying by a timestep that should be fine however I think the proper way is to have monitor the trend of velocities.

_Note_: there is no cheating in this solution, I can use tf2 to read the change in position over the cycle and the solution wouldn't change much, but gazebo/link_states gives already provides me with a stream of joints positions it would much more efficient implementation this way.

Again, no cheating in this approach.

Samahu: It is more efficient than using tf2 because I don't have to pull the joint values.


The only place this would be cheating is if we want the solution to work outside of the simulation i.e on hardware robot.

mogumbo: This solution is based on distance, not velocity. Changing the timestep of the sim could affect this solution. I hope I didn't offend too much with the word cheating. I only mean to say it's using ground truth info that a real robot wouldn't have.

Samahu: I did address the cheating topic (not responding because of feeling offended). Before implementing I thought of using **tf2** but resorted to using **gazebo/link_states** mainly favoring to subscribe to the available stream of values rather than having to pull these values out of tf2.

   

> This solution is based on distance, not velocity.
the computed delta is between two readings taken at two different times, the system clock is not factored but that's a different thing.

Samahu: In any case I will provide an implementation based off tf2 (while keeping the gazebo/link_states until I fully understand the motives)

mogumbo: Actually, I'm assuming this script runs once per gazebo timestep. If it's running on a separate clock then it might not matter.

Samahu: I did not say that factoring time delta wouldn't be useful. I was just trying to explain that what I have here is technically velocity. I haven't put enough thought into the GROUND_DETECTION_THRESHOLD value.

Samahu:
I have implemented the same solution using **TF**. I did verify that the solution works both ways (**TF** or **gazebo/link_states**) on my machine, however, the optimal GROUND_DETECTION_THRESHOLD value will be different for each method. Maybe factoring in the time delta as @mogumbo suggested would allow us to use the same threshold value for both methods (after all time delta is much larger for the **TF** method than the **gazebo/link_state** method.

Now that the **TF** is implemented (it is the default method) I will work on tuning GROUND_DETECTION_THRESHOLD constant.

ow_lander/scripts/trajectory_publisher.py
Samahu: addressed

damianacatanoso:
> This tests well for me. Can we safely speed up the ground search while we're at it by changing activity_guarded_move.py Line 107? @damianacatanoso, is this safe? I think you have experience with it.

I tested it removing the scaling factor and it works very well. So, @Samahu, I am pretty confident we can remove it.

damianacatanoso (approved):
- As said in one of the inline comments, line 170 in activity_guarded_move.py can be eliminated. 
- Ground detection is much more accurate than it was before. Running some experiments I founds that, in europa_terminator_workspace, the terrain height is around -0.155 and this corresponds to the z value published on guarded_move_result using this branch. Great work!

anjan1984 (approved):
Works great for me

mogumbo (approved):

PR #17, opened 2020-09-16
Simons errors

mogumbo:
After communicating with an external user, he found his Ubuntu 18 system was defaulting to python 3 instead of python 2. I made changes to scripts to ask for python 2 explicitly. He and I both tested this branch successfully.


kmdalal (approved):

Samahu (approved):
Again, I think it is more of a user environment management, but I see no harm.

PR #19, opened 2020-09-21
Oceanwater 475 publish rul

mollocon44:


ow_power_system/src/PowerCSV.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+
+/* Function to load pre-generated values as a vector of vectors of doubles
+ * Returns std::vector<std::vector<std::double>>
+ */
+std::vector<std::vector<double>> powerCSV::loadCSV(const std::string& filename){
+  ifstream power_csv(filename);
+  if (power_csv.fail()) {
+    cout << "Unable to open data file" << endl;
+  }
```
kmdalal, line 20: Shouldn't the function return here?

ow_power_system/src/PowerCSV.cpp
```diff
@@ -0,0 +1,57 @@
...
+  vector<vector<double>> values;
```
kmdalal, line 21: I'm not sure how 'vector' works here without the 'std::' prefix, like you use everywhere above, plus in the header file.  I strongly suggest putting a 'using std::vector' at the top of this file for clarity, and drop the 'std::' from vector everywhere for consistency.

ow_power_system/src/PowerCSV.cpp
```diff
@@ -0,0 +1,57 @@
+//PowerCSV.cpp
```
mogumbo, line 1: This line is unnecessary. You can remove it.

ow/launch/power_demo.launch
```diff
@@ -0,0 +1,83 @@
+<?xml version="1.0"?>
```
mogumbo, line 1: Does this file need to exist? It looks like it's almost the same thing as common.launch. If you need to have a demo, can it be a small launch file that includes common.launch?

ow_power_system/include/PowerCSV.h
```diff
@@ -0,0 +1,19 @@
+//PowerCSV.h
```
mogumbo, line 1: This line is also redundant. Please remove it.

ow_power_system/src/power_system_node.cpp
```diff
@@ -17,27 +24,112 @@ int main(int argc, char* argv[]) {
...
+  
+  //Load power values csv
+  string csv_path,csv_file;
+  csv_file = "/data/onewatt.csv";
+  //FILE OPTIONS: onewatt.csv, eightwatt.csv, sixteenwatt.csv
+  csv_path = ros::package::getPath("ow_power_system");
+  csv_path += csv_file;
+
+  auto power_csv = loadCSV(csv_path);
+  cout << "Power system CSV loaded!";
```
kmdalal, line 42: Please remove any C++ stream output messages that are just for your development or debugging.  If the message is important for users (probably not this one), use ROS output (ROS_INFO, ROS_WARN, ROS_ERROR).   You can also use ROS_DEBUG to keep your debugging messages in.

ow_power_system/src/power_system_node.cpp
```diff
@@ -17,27 +24,112 @@ int main(int argc, char* argv[]) {
...
+
+
+/* Function to load pre-generated values as a vector of vectors of doubles
+ * Returns std::vector<std::vector<std::double>>
+ */
+std::vector<std::vector<double>> loadCSV(const std::string& filename){
+  ifstream power_csv(filename);
+  if (power_csv.fail()) {
+    cout << "Unable to open data file" << endl;
+  }
```
kmdalal, line 98: Shouldn't function return here?

ow_power_system/src/power_system_node.cpp
```diff
@@ -17,27 +24,112 @@ int main(int argc, char* argv[]) {
...
   }
 
   return 0;
 }
+
+
+/* Function to load pre-generated values as a vector of vectors of doubles
+ * Returns std::vector<std::vector<std::double>>
+ */
+std::vector<std::vector<double>> loadCSV(const std::string& filename){
```
mogumbo, line 94: This function is duplicate code from PowerCSV.cpp. Please remove one of them.

kmdalal, line 94: Though I didn't compare them carefully, this function appears to be duplicated in PowerCSV.cpp.  Can you remove one, or refactor if they are slightly different?  (Refactoring not essential for this release).

mogumbo (changes requested):

kmdalal (changes requested):
See my inline code comments that request changes.

ow/launch/power_demo.launch
mollocon44: Sorry, didn't realize I was tracking that one. I just made it so I could run without gazebo without accidentally changing common.launch in melodic-devel

ow_power_system/src/PowerCSV.cpp
mollocon44: Yes. Added solution s.t. it should just publish zeros for all values in that case.

ow_power_system/include/PowerCSV.h
mollocon44: File removed.

ow_power_system/src/PowerCSV.cpp
mollocon44: File removed.

ow_power_system/src/power_system_node.cpp
mollocon44: PowerCSV.h, PowerCSV.cpp removed.

ow_power_system/src/power_system_node.cpp
mollocon44: Added. Returns zero values to be published by node and adds ROS_WARN message.

ow_power_system/src/power_system_node.cpp
mollocon44: Done.

mollocon44 (commented):
Comments addressed. Removed unnecessary files and changed output to ROS_DEBUG or ROS_WARN

ow_power_system/src/power_system_node.cpp
```diff
@@ -17,27 +25,117 @@ int main(int argc, char* argv[]) {
...
+  csv_file = "/data/onewatt.csv";
+  //FILE OPTIONS: onewatt.csv, eightwatt.csv, sixteenwatt.csv
+  csv_path = ros::package::getPath("ow_power_system");
+  csv_path += csv_file;
+
+  auto power_csv = loadCSV(csv_path);
+  
+  //Define our publication rate and initialize our time to 0
+  double power_update_rate;  
+  if (csv_file=="/data/onewatt.csv"){
```
Samahu, line 46: When this condition is ever false?

ow_power_system/src/power_system_node.cpp
```diff
@@ -17,27 +25,117 @@ int main(int argc, char* argv[]) {
 
   //Construct our power draw listener
   //TODO: Check out latching
```
mogumbo, line 27: If there is work to do please make a Jira ticket for it.

ow_power_system/src/power_system_node.cpp
```diff
@@ -17,27 +25,117 @@ int main(int argc, char* argv[]) {
...
+  auto power_csv = loadCSV(csv_path);
+  
+  //Define our publication rate and initialize our time to 0
+  double power_update_rate;  
+  if (csv_file=="/data/onewatt.csv"){
+    power_update_rate = 0.10; //csv is at 10Hz intervals
+  } else {
+    power_update_rate = 1.0; // 1 Hz default
+  }
+  int t_step= 1;//starting at 1 not zero to skip csv headings
```
Samahu, line 51: based on loadCSV code, I think this t_step should be set to zero

```suggestion
  int t_step= 0;
```

ow_power_system/src/power_system_node.cpp
```diff
@@ -17,27 +25,117 @@ int main(int argc, char* argv[]) {
...
 
-    soc_msg.data=0.0; //zero for testing purposes, will be calculated later
+  int power_lifetime = power_csv.size();
+  while (ros::ok()) {
+    if (t_step < power_lifetime) {
+      soc_msg.data=power_csv[t_step][voltage_i];
+      rul_msg.data=power_csv[t_step][rul_i];
+    } //should keep publishing last voltage value if sim time exceeds the length of the csv
+    
+    //soc_msg.data=t;
```
mogumbo, line 74: This looks like more work to do, so maybe make another Jira ticket.

mogumbo (approved):
I added a couple more comments that you can handle as you see fit, but I think it's okay to merge.

ow_power_system/src/power_system_node.cpp
```diff
@@ -17,27 +25,117 @@ int main(int argc, char* argv[]) {
...
+ * Returns std::vector<std::vector<std::double>>
+ */
+vector<vector<double>> loadCSV(const std::string& filename){
+  ifstream power_csv(filename);
+  if (power_csv.fail()) {
+    ROS_WARN("Loading power csv file has failed. power_system_node will publish zeros.");
+    vector<double> zero_row = {0.0,0.0,0.0};
+    vector<vector<double>> zeros;
+    zeros.push_back(zero_row);
+    zeros.push_back(zero_row);
```
Samahu, line 101: there reason you are adding two **zero_row** here is because **t_step** needlessly set to 1.

ow_power_system/src/power_system_node.cpp
mollocon44: There are three different options for csv files and two are at different rates.

ow_power_system/src/power_system_node.cpp
mollocon44: There already is for the power listener, I'll update the description to include checking out latching.

ow_power_system/src/power_system_node.cpp
mollocon44: Nope, I did do that! I'll get rid of the comment.

ow_power_system/src/power_system_node.cpp
mollocon44: For all other csvs, there is a header row to skip -- I'll change it in the loadCSV function so I can set t_step to 0 and remove the second zero_row.

mollocon44: Actually, wait, I was looking at my column loading. I'd just have to repeat a line similarly in the section that actually matters, so I'd prefer to keep it the way it is. It's meant to be a temporary solution anyway, so it shouldn't really matter - do you feel strongly that it has to be done exactly this way?

ow_power_system/src/power_system_node.cpp
Samahu: but one need to change the code to try with different csv. Shouldn't the file csv file be passed as a **ros::param**? This way the user don't need to recompile your code each time they want to try a different option.

ow_power_system/src/power_system_node.cpp
Samahu: Nope, it is not a big deal. I just saw that your loadCSV function skips the csv header line and then adds samples.

kmdalal (approved):

ow_power_system/src/power_system_node.cpp
mollocon44: Ah thanks, I'm with you now - just fixed it.

PR #21, opened 2020-09-22
Threshold ground detection based on velocity trend

Samahu:
## Summary of changes
Significantly improved ground detection by tracking the **trending value** of the arm movement speed and use it to decide if the arm reached ground or not. 

## Verify
Launch the simulation against the europa_terminator_workspace and atacama_y1a
```bash
```
or
```bash
```

Use __ros service caller__ or your favorite method to perform the invoked guarded detection command.

Observe whether the ground detection takes place at the instant the scoop touches the ground. This should perform better than before as I was able to lower the threshold value with the latest changes.

**Note:** Even though I compute the arm movement in 3d but the current criterion to detect ground use only on the z values for decision making. The reason I didn't try this is I would require some time to create these scenarios and test them. Nonetheless, the current implementation may actually be able to detect the guarded move in 3d as the largest trending value will be the z value. If you are able to test that easily please do so and let me know if this works or not.

## Objectives
* Make sure the code runs successfully 
* run the guarded move command multiple times in one session against the current configuration on melodic-devel, also you may try the following configurations:
  - europa_terminator_workspace
  - atacma_y1a
  - world solver instead of quick
  - increase or decrease the trajectory publish rate as the arm descends

If you can think of any other scenario please try it and report any failure and I can work on addressing that.

**Linked Issues:**
* https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-481 (current issue)
* https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-485 (follow up issue)

### DIFF
--- a/ow_lander/scripts/ground_detection.py
+++ b/ow_lander/scripts/ground_detection.py
@@ -54,7 +54,8 @@ class GroundDetector:
 
     if self._check_ground_method == self._gazebo_link_states_method:
       rospy.Subscriber("/gazebo/link_states", LinkStates, self._handle_link_states)
-      self._query_ground_position_method = lambda: self._tf_lookup_position(rospy.Duration(10))
+      self._query_ground_position_method = lambda: self._tf_lookup_position(
+          rospy.Duration(10))
     else:
       self._query_ground_position_method = lambda: self._last_position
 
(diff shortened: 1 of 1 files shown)


mogumbo (approved):

anjan1984 (approved):
Works great for me!

PR #24, opened 2020-09-25
power_csv_file and power_update_rate both configurable

Samahu:
## Summary of Changes
* The power_system_node has two configurable parameters `power_draw_csv_path` and `power_update_rate`
*  Both these values need to be set otherwise default values will be used
* The parameters can be set through command line arguments during the launch of the simulation
  - When using command line args for the power_draw only the name of the csv file is needed
* power_update_rate is now expressed in Hz rather than seconds
* README.md file updated to reflect these changes and some formatting improvements

~_Note:  In README.md, the "Update interval" line for **onewatt** option was wrongly stating that the update interval is 10 seconds, in reality it is 0.1 seconds_~ WRONG  

## Verify
launch the simulation
```bash
```

Pay attention to power_system_node messages and verify that the right parameter is picked
  
Test out different values which you can pass through the command line, for example
```bash
roslaunch ow atacama_y1a.launch power_draw_csv_file:=eightwatt.csv
```

**Linked Issue:** https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-494


Samahu:
@mollocon44 @chetankul I think at least one of you has to review this PR. Note that this PR targets release-6 so it has to be merged by tomorrow.

ow_power_system/src/power_system_node.cpp
```diff
@@ -29,22 +29,41 @@ int main(int argc, char* argv[]) {
...
   }
-  int t_step= 0;
+  else
+  {
+    // Validated the parameter
+    if (power_update_rate == 0)
+    {
+      power_update_rate = 10;
+      ROS_WARN_NAMED("power_system_node", "power_update_rate param was set to zero! Using default value: 10 Hz");
+    }
```
Samahu, line 58: Maybe this should be an error, I am not sure.
Or instead we can consider that setting a value of zero means using the highest rate possible (by convention) .

ow/launch/common.launch
```diff
@@ -6,6 +6,8 @@
   <arg name="gazebo" default="true"/>
   <arg name="gzclient" default="true"/>  <!-- Only has effect if gazebo = true -->
   <arg name="record_bag" default="false"/>
+  <arg name="power_draw_csv_file" default="onewatt.csv"/>  <!-- options: onewatt.csv, eightwatt.csv, sixteenwatt.csv -->
+  <arg name="power_update_rate" default="10"/>  <!-- Use 10 Hz for onewatt.csv, 1 Hz for the two other options -->
```
mollocon44, line 10: The 10 Hz was the incorrect one, not the 10s. onewatt.csv should be publishing at 10s intervals. I'll fix both here and in the readme.

ow_power_system/src/power_system_node.cpp
mollocon44: It looks like Rate uses the exact frequency, right? http://docs.ros.org/diamondback/api/rostime/html/rate_8cpp_source.html
We want onewatt.csv to update every 10 seconds and the others to update every 1 second.

mollocon44 (commented):
Made changes to default rate: onewatt.csv should have a 10s (0.1Hz) update rate.

mollocon44:
Fixed the default value. power_update_rate potentially shouldn't be settable, due to the fact that the csvs are tied to a time interval. This will be useful for the battery model though, so I'm leaving it

mogumbo (changes requested):
I used a fresh build with this branch checked out alongside releases/release-6-0 branches in the other repos. No matter what launch line I use (even one with an invalid csv filename), I always get this in the console:

> * /power_system_node/power_draw_csv_path: /home/twelsh/work...
> * /power_system_node/power_update_rate: 0.1

And the power_system_node topics output the contents of onewatt.csv at 10 second intervals.

I don't understand the need for power_update_rate since the output rate is defined in the csv files.

Samahu:
> Made changes to default rate: onewatt.csv should have a 10s (0.1Hz) update rate.

I think there was a bit of confusion on my part with regard to ros::Rate .. I thought it takes seconds .. sorry about that

Samahu:
> I don't understand the need for power_update_rate since the output rate is defined in the csv files.

The rate is not embedded into the csv file, it doesn't make sense to have them in the csv as they are a single value needed on initializing. They were hard coded into the code based on the filename.

Samahu:
> Note: In README.md, the "Update interval" line for **onewatt** option was wrongly stating that the update interval is 10 seconds, in reality it is 0.1 seconds

also apparently this change that I made to README.md was wrong. I'll fix it

Samahu:
> I used a fresh build with this branch checked out alongside releases/release-6-0 branches in the other repos. No matter what launch line I use (even one with an invalid csv filename), I always get this in the console:
> 
> > * /power_system_node/power_draw_csv_path: /home/twelsh/work...
> > * /power_system_node/power_update_rate: 0.1
> 

Good catch! That's because I forgot to propagate to power arguments from common.launch to the simulation startup launch files.

mogumbo:
> > I don't understand the need for power_update_rate since the output rate is defined in the csv files.
> 
> The rate is not embedded into the csv file, it doesn't make sense to have them in the csv as they are a single value needed on initializing. They were hard coded into the code based on the filename.

The first column in the files is time, which implies a rate. For example, onewatt.csv contains time values 0, 10, 20, 30, etc. So if we output those rows they should output once every 10 seconds or a rate of 0.1 Hz. So if you specify a different rate, what is the expected behavior? Will there be interpolation between rows to get the right values for the new times you are outputting?

I updated and did a fresh build. roslaunch ow atacama_y1a.launch power_draw_csv_file:=eightwatt.csv power_update_rate:=5 now outputs 0.0 for both power topics.

Samahu:
> The first column in the files is time, which implies a rate. For example, onewatt.csv contains time values 0, 10, 20, 30, etc. So if we output those rows they should output once every 10 seconds or a rate of 0.1 Hz. So if you specify a different rate, what is the expected behavior? Will there be interpolation between rows to get the right values for the new times you are outputting?

Although I get the picture .. this wasn't part of the initial implementation. The code originally used hard coded values and I made it configurable.

Samahu:
> I updated and did a fresh build. roslaunch ow atacama_y1a.launch power_draw_csv_file:=eightwatt.csv power_update_rate:=5 now outputs 0.0 for both power topics.

This suggests that it is not able to find any of the csv files. It is working fine on my machine .. can't produce it.

ow_power_system/src/power_system_node.cpp
```diff
@@ -29,22 +29,42 @@ int main(int argc, char* argv[]) {
   ros::Publisher RUL_pub = n.advertise<std_msgs::Float64>("remaining_useful_life",1000);
   
   //Load power values csv
-  string csv_path,csv_file;
-  csv_file = "/data/onewatt.csv";
-  //FILE OPTIONS: onewatt.csv, eightwatt.csv, sixteenwatt.csv
-  csv_path = ros::package::getPath("ow_power_system");
-  csv_path += csv_file;
+  string csv_path;
+  auto csv_path_param_exist = n.param("power_draw_csv_path", csv_path,
```
kmdalal, line 33: A more descriptive variable name than 'n' would be nice, even as short as 'nh' for node handle.

ow_power_system/src/power_system_node.cpp
```diff
@@ -29,22 +29,42 @@ int main(int argc, char* argv[]) {
...
+    ros::package::getPath("ow_power_system") + "/data/onewatt.csv");
```
kmdalal, line 34: Would be good to abstract "/data/onewatt.csv" with a variable.

ow_power_system/src/power_system_node.cpp
```diff
@@ -29,22 +29,42 @@ int main(int argc, char* argv[]) {
...
-  
-  //Define our publication rate and initialize our time to 0
-  double power_update_rate;  
-  if (csv_file=="/data/onewatt.csv"){
-    power_update_rate = 0.10; //csv is at 10Hz intervals
-  } else {
-    power_update_rate = 1.0; // 1 Hz default
+
+  // Retrieve our publication rate expressed in Hz
+  double power_update_rate = csv_path.find("onewatt") != string::npos ? 0.1 : 1;  // if onewatt default to 0.1 Hz otherwise 1 Hz
```
kmdalal, line 46: I don't understand why "onewatt" is used here, irregardless of the power_draw_csv_path parameter.

ow_power_system/src/power_system_node.cpp
```diff
@@ -29,22 +29,42 @@ int main(int argc, char* argv[]) {
...
+    ROS_WARN_NAMED("power_system_node", "Overriding the default power_update_rate!");
+    // Validated the parameter
+    if (power_update_rate == 0)
+    {
+      ROS_ERROR_NAMED("power_system_node", "power_update_rate param was set to zero! passed value ignored.");
+    }
+    else
+    {
+      power_update_rate = power_update_rate_override;
+    }
```
kmdalal, line 61: I'm confused by what's going on here.  The user can override the default power_update_rate, but only if it was NOT set to 0?

ow_power_system/src/power_system_node.cpp
kmdalal: In my build of this branch, there is no "onewatt.csv" or any other .csv file, in the ow/devel tree.  Is this a build problem?

Samahu: Note sure, It is under **ow_power_system/data**.

ow_power_system/src/power_system_node.cpp
Samahu: Well, this is a hard coded logic .. if "onewatt" is used then the default publish rate is 0.1 Hz, 1 for others.

ow_power_system/src/power_system_node.cpp
Samahu: I will address shortly

ow_power_system/src/power_system_node.cpp
Samahu: if it is set to zero it might result in a division by zero .. see the ros::Rate constructor

ow_power_system/src/power_system_node.cpp
Samahu: addressed

mogumbo (approved):

kmdalal (approved):

PR #26, opened 2020-09-28
Oceanwater 487 energy efficient planner

damianacatanoso:
**ABOUT THIS BRANCH**
While reviewing the renovated grinder service, @Samahu noticed that the arm would make an unnecessary motion before the grinder penetrates the terrain. Here is the video: https://www.youtube.com/watch?v=IytOorLmMf8 . This raised a question on whether the planner is actually optimizing the motion for something in particular. I did some research and discovered the following.

- In OceanWATERS we use two planners: RRTconnect and RRTstar. RRT connect is not an optimal planner. The basic idea behind RRTconnect is to grow two trees, one from the start and one from the goal, and attempt to connect them.
- RRTstar is an optimal planner. For RRTstar, the following optimization objectives are available: PathLengthOptimizationObjective (Default), MechanicalWorkOptimizationObjective, MaximizeMinClearanceObjective, StateCostIntegralObjective, MinimaxObjective. 
- We are interested in minimizing the mechanical work, so I introduced in the ompl_planning.yaml a line to specify that. 

Now that the optimization objective is specified, I removed from activity_grinder.py that intermediate motion that I had to introduce to avoid the arm twisting. I tested multiple times and the twisting doesn't happen. 

**TEST**
- Please plan "Unstow". 
- Then plan the grinding service with use_defaults=True. You don't need to publish any trajectory because this was a problem at the planning level.
- Plan "Unstow" again.
- Plan "Grind" with x=1.65, y=0, depth=0.1, length=0.6, parallel=False, ground_position=-0.155
Please report if you see any twisting, thanks!

### DIFF
--- a/ow_lander/scripts/activity_grind.py
+++ b/ow_lander/scripts/activity_grind.py
@@ -65,19 +65,6 @@ def grind(move_arm, move_limbs, move_grinder, args):
   h = math.sqrt( pow(y_start-constants.Y_SHOU,2) + pow(x_start-constants.X_SHOU,2) )
   l = constants.Y_SHOU - constants.HAND_Y_OFFSET
   beta = math.asin (l/h)
-
-  joint_goal = move_arm.get_current_joint_values()
-  joint_goal[constants.J_DIST_PITCH] = 0
-  joint_goal[constants.J_HAND_YAW] = 4*math.pi/3
-  joint_goal[constants.J_PROX_PITCH] = -math.pi/2
-  joint_goal[constants.J_SHOU_PITCH] = math.pi/2
-  joint_goal[constants.J_SHOU_YAW] = alpha + beta
-  # If out of joint range, abort
-  if (is_shou_yaw_goal_in_range(joint_goal)==False):
-    return False
-  move_arm.go(joint_goal, wait=True)
-  move_arm.stop()
-
   alpha = alpha+beta
   
   if parallel:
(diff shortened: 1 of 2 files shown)


mogumbo (approved):

ow_lander/scripts/activity_grind.py
```diff
@@ -65,19 +65,6 @@ def grind(move_arm, move_limbs, move_grinder, args):
...
-  joint_goal[constants.J_HAND_YAW] = 4*math.pi/3
-  joint_goal[constants.J_PROX_PITCH] = -math.pi/2
-  joint_goal[constants.J_SHOU_PITCH] = math.pi/2
-  joint_goal[constants.J_SHOU_YAW] = alpha + beta
-  # If out of joint range, abort
-  if (is_shou_yaw_goal_in_range(joint_goal)==False):
-    return False
-  move_arm.go(joint_goal, wait=True)
-  move_arm.stop()
-
```
Samahu, line 80: What these lines removed as part of this PR?

Samahu: In the test procedure you specify to test the _stow_ but this part touches the _grind_ command

Samahu: I saw that you put the reason for removing these lines in the PR description

Samahu (approved):
Works as expected

PR #27, opened 2020-10-06
Fixed scoop-terrain instability, updated joints effort limit and friction 

damianacatanoso:
This PR covers two Jira tickets: 
- https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-470
- https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-499

I used the same branch and same PR for both because: 1) both tickets require small changes and 2) the values of the parameters that I changed for the second ticket depends on the values of the parameters set in the first ticket.

**1) Joint compliance**
The aim of this ticket was to investigate if it's possible to add flexibility (or compliance) to the joints as an attempt of solving the scoop bouncing instability that manifests after guarded_move finds the ground. After doing research on the many parameters that can be set on the robot urdf and the world files, I found that it is possible to stabilize the interaction between two links using the kp and kd parameters, here: http://gazebosim.org/tutorials?tut=physics_params&cat=physics
So I looked for a way to apply the same concept to the scoop-environment interaction and I found that defining those parameters in the urdf file, instead of the world file, under the <gazebo> section of the scoop link, the kp and kd values can be assigned specifically to the scoop interacting with the terrain. Please note that the scoop-terrain contact is now stable for both quick and world solver.

**2) Assign real values for maximum torque and friction**
While doing the research above I realized that we probably never did a proper research on real values of the parameters that appear in the joint limits section of lander.xacro, such as maximum torque, maximum velocity, maximum and minimum angle. So I did it and documented my findings here: https://babelfish.arc.nasa.gov/confluence/display/OCEANWATERS/Arm+Dynamics . I kept our current velocity limits because, in real missions, movements are too slow. I only changed the maximum available torque according to the collected data and I introduced a static friction torque equal to 10% of the maximum available torque.

**TEST**
- Launch europa_terminator_workspace.launch 
- In new terminal: 
$ rosservice call /planning/arm/guarded_move True False 0 0 0 0 0 0 0
- Then:
$ rosservice call /planning/arm/publish_trajectory True a
- before the scoop touches the ground, in new terminal enter: 
$ rostopic echo guarded_move_result
- When the scoop touches the ground you should see 1) no oscillations/vibrations and 2) the echo of the message containing the ground coordinates.

- Do the same in atacama_y1a.
- Do the same, changing type of solver to "world": in ow_europa>worlds>terminator_workspace.world or atacama_y1a.launch (depending on which sim you are launching), line 21, replace "quick" with "world".
- If you want, you can try the above using different guarded_move parameters, such as:
$ rosservice call /planning/arm/guarded_move False False 1.8 0.2 0.5 1 1 1 0.7
- You can also try to run ReferenceMission1.plx and verify that the instability at the end of guarded_move disappeared:
$ roslaunch ow_autonomy autonomy_node.launch plan:=ReferenceMission1.plx

- Let me know if everything works fine for you, thanks!


kmdalal (approved):
Tested with the default solver on europa_terminator_workspace and atacama, as instructed.   And with ReferenceMission1.  Looked great.   I don't have time to test with other solvers --  someone else should.

Samahu (approved):

mogumbo:
Holy physics stability, Batman! Nice work. If I understand correctly, you have added friction/damping to joints and added spring/damping behavior to the scoop-terrain collision, and both of these give us better stability. But there is no compliance added to the arm. Please tell me if I have this wrong.

Is it possible or advantageous to add kp/kd values globally? If not, should we add kp/kd to other links, such as the grinder and hand?

Should we switch to the world solver and attempt to lower the physics update rate for better performance? Or should that be part of a new ticket?

damianacatanoso:
> Holy physics stability, Batman! Nice work. If I understand correctly, you have added friction/damping to joints and added spring/damping behavior to the scoop-terrain collision, and both of these give us better stability. But there is no compliance added to the arm. Please tell me if I have this wrong.
> 
The damping on the joints was already there. I only added a little static friction (5% of total available torque) to the joints to make the sim more realistic. Adding damping and friction to joints does not affect physics stability between scoop and terrain. According to my research there is no way of adding compliance to joints/links, so no compliance has been added. What affects this stability is the tuning of the two parameters kp and kd, which, as you said, add spring/damping behavior to the scoop-terrain collision. 
> Is it possible or advantageous to add kp/kd values globally? If not, should we add kp/kd to other links, such as the grinder and hand?
> 
Yes, it is possible to add these values globally, directly in the .world file. I remember trying and the sim was working fine. I didn't test all the possible arm motions with the global values though. I think, if the other interactions are fine, there is no reason of adding those parameters, they might create problems in the future.  I agree that we should set kp/kd for the grinder as well.
> Should we switch to the world solver and attempt to lower the physics update rate for better performance? Or should that be part of a new ticket?
We should definitely try that! The right values of kp and kd might be a function of the timestep. So I would to do it in this ticket/PR. Can you suggest a lower bound for feasible timestep?

mogumbo:
> We should definitely try that! The right values of kp and kd might be a function of the timestep. So I would to do it in this ticket/PR. Can you suggest an upper bound for feasible timestep?

The world files currently set physics at 400 Hz. Try for 100, but don't be surprised if you can't get it that low. 200 would probably give us performance that is quite good.

damianacatanoso:
> The world files currently set physics at 400 Hz. Try for 100, but don't be surprised if you can't get it that low. 200 would probably give us performance that is quite good.
>
The best I could do is 160. If you want to take a look at it, pull OCEANWATER-470_Joints_compliance in both ow_simulator and ow_europa.

mogumbo:
I have been testing with a timestep of 0.005, which is a little more conservative and safer than 0.00625. Everything seems stable with the world solver, but there is some instability with the quick solver. I think that's okay since the quick solver isn't supposed to be as good and we would prefer to not use it.

Running the autonomy demo, guarded_move failed to find the ground twice on the atacama terrain and dig_circular stopped before the scoop left the ground on terminator_workspace. I don't know that these problems are related to this PR, but we definitely have some refinement to do.

damianacatanoso:
It looks like I will need more work on the stability issues when lowering the update rate and increasing the time step in the world file. So I'll close this PR, merge commit 2b84b7fba3c8ee3f5418bad437a1bb695e5907b5 of this branch to melodic-devel and continue the stability work on a separate ticket/branch.

PR #31, opened 2020-10-16
Feature/oceanwater 485 dynamically infer ground detection threshold

Samahu:
## Summary of Changes
* This change gets rid of having to set a fixed threshold. This change addresses an issue with the current implementation fails to detect ground if arm speed is reduced significantly.
* Code Formatting & Docstring

## Verfity
```bash
roslaunch ow atacama_y1a.launch (or europa_terminator_workspace.launch)
```

**Tests**:
- Then invoke guarded move planning followed by publish_trajectory under default settings (run multiple times in same session)
- Repeat the process after reducing joint speed constraints to *0.01* for all of arm joints in the `lander.xacro` file
 
>__Note:__ This is designed  to work against a mainly downwards movement
 
__Linked Issue(s):__ https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-485

### DIFF
--- a/ow_lander/scripts/ground_detection.py
+++ b/ow_lander/scripts/ground_detection.py
@@ -78,7 +78,7 @@ class GroundDetector:
 
   def _check_condition(self, new_position):
     """
-    :type new_position: class 'geometry_msgs.msg._Vector3.Vector3
+    :type new_position: geometry_msgs.msg._Vector3.Vector3
     """
     if self._last_position is None:
       self._last_position = np.array(
--- a/ow_lander/scripts/ground_detection.py
+++ b/ow_lander/scripts/ground_detection.py
@@ -76,7 +76,7 @@ class GroundDetector:
     self._ground_detected = False
     self._trending_velocity = SlidingWindow(5, np.mean)
     self._dynamic_threshold = None
-    self._threshold_tolerance = 0.015 # TODO: infer tolerance dynamically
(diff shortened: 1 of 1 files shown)


kmdalal:
In my initial test (branch unmodified) GuardedMove was successful, but the scoop was "trembling" at the end -- I have not seen this in weeks.  I assume this is an unrelated issue.

I just did one run, didn't have time for more.

After changing velocity limits to 0.01, and waiting for the very slow operations to complete, everything was successful, and there was no trembling at the end.  

The actual guarded move did not reach as far as the ghost arm -- see attached pic.  However, I've seen this often and assume it is also an unrelated issue.

lander

kmdalal (approved):

Samahu:
> The actual guarded move did not reach as far as the ghost arm -- see attached pic. However, I've seen this often and assume it is also an unrelated issue

I presume you are running the simulation without the gazebo client. of course it won't match .. the ghost arm is the planned position .. The actual arm won't match the ghost arm since the scoop hits the ground before reaching the planned goal.

With regard to the trembling .. make sure you have update ow_europa to _melodic-devel_ and pulled to the latest

damianacatanoso (approved):
Works for me. I observe a violent bouncing motion which should be solved after merging https://github.com/nasa/ow_simulator/pull/27 . I didn't observe false ground detection in this branch so I approve.

PR #35, opened 2020-11-10
Added kp, kd to grinder contact, reduced scoop rotation during dig_ci…

damianacatanoso:
Ticket: https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-454

Problem: when dig_circular is called with parallel=False, at the end of the hand rotation the grinder would 1) collide with the terrain before it visually touches the ground and 2) bounce.

Solution: part 2) of problem above is solved introducing kp and kd for the grinder, as it's been done for the scoop. part 1) is solved replacing the mesh with a series of co-axial cylinders only in the *collision* model.

Approaches tried for part 1) of the problem:
- I re-drew in SolidWorks the grinder model, saved the STL using the highest number of triangles (resolution), replaced the old one with this new model. Nothing changed.
- Replaced the grinder with the scoop model and the problem disappears. The scoop mesh collides with the ground when there is visual contact.
- Replaced grinder with cylinder and the problem disappears. Here is a video: https://www.youtube.com/watch?v=NAZGdhlEyT8

Test: 
- launch europa_terminator_workspace 
- $ rosservice call /planning/arm/grind -- False 2.0 0.0 0.10 0.4 False -0.155 False
- publish trajectory
- $ rosservice call /planning/arm/dig_circular -- False 2.0 0.02 0.1 False -0.155 False
- publish trajectory
- check that, at the end of dig_circular, the grinder doesn't bounce anymore and it actually touches the terrain. 

I did this test setting the depth of "modify_terrain_circle_message" in modify_terrain_grinder_pub.py to 0.16 to observe that shape of the trench. The shape of the trench doesn't change. I did also play around with the Message Publisher in rqt, getting the grinder to approach the terrain from the sides, and checked if the grinder touches the terrain when collision is detected. And it does!

### DIFF
--- a/ow_lander/scripts/activity_full_digging_traj.py
+++ b/ow_lander/scripts/activity_full_digging_traj.py
@@ -211,7 +211,7 @@ def dig_circular(move_arm, move_limbs, args):
     go_to_Z_coordinate(move_limbs, x_start, y_start, z_start)
 
     # Rotate hand perpendicular to arm direction
-    change_joint_value(move_arm, constants.J_HAND_YAW, -0.29*math.pi)
+    change_joint_value(move_arm, constants.J_HAND_YAW, -0.23*math.pi)
 
   else:
     # Rotate hand so scoop is in middle point
--- a/ow_lander/scripts/activity_full_digging_traj.py
+++ b/ow_lander/scripts/activity_full_digging_traj.py
@@ -211,7 +211,7 @@ def dig_circular(move_arm, move_limbs, args):
     go_to_Z_coordinate(move_limbs, x_start, y_start, z_start)
 
     # Rotate hand perpendicular to arm direction
-    change_joint_value(move_arm, constants.J_HAND_YAW, -0.23*math.pi)
+    change_joint_value(move_arm, constants.J_HAND_YAW, -0.29*math.pi)
 
   else:
     # Rotate hand so scoop is in middle point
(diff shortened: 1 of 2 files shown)


Samahu:
Did you try going with a lower number of triangles instead of increasing the number of triangles (Typically collision libraries work better with lower number of triangles for mesh collider)?

damianacatanoso:
> Did you try going with a lower number of triangles instead of increasing the number of triangles (Typically collision libraries work better with lower number of triangles for mesh collider)?

No, I didn't, but I should try that. I still don't have my smartcard so I cannot connect remotely to the machine with SolidWorks, which is at Ames. I am not sure when my badge will be ready, so I am not sure when I'll be able to do this. Can we keep this PR open until then? Or is it better to close this and open another ticket?

Samahu:
> > Did you try going with a lower number of triangles instead of increasing the number of triangles (Typically collision libraries work better with lower number of triangles for mesh collider)?
> 
> No, I didn't, but I should try that. I still don't have my smartcard so I cannot connect remotely to the machine with SolidWorks, which is at Ames. I am not sure when my badge will be ready, so I am not sure when I'll be able to do this. Can we keep this PR open until then? Or is it better to close this and open another ticket?

If you just want to edit the STL to reduce the poly count, you can probably use **blender** to do that.

damianacatanoso:
Hi everyone! I just pushed a working solution for this issue and I updated the description and testing instructions. Please read again and test. Thanks for your advice!

mogumbo (approved):
It works for me. I thought you had to have separate links for each collider. Glad to see that I was wrong and you are able to use multiple collision tags under one link.

Samahu (approved):

PR #37, opened 2020-11-16
Reduced joints damping. For scoop collision: removed kp, reduced kd, …

damianacatanoso:
**Ticket:** 
https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-513

**Aim of this work:** 
Speed up simulation --> lower the real time update rate and increase the max step size to keep the real time factor equal to 1. 

**What I did and learnt:**
Decreasing the real_time_update_rate introduces instability in the physics. A proper tuning of collision/contact related parameters can make the system stable. These parameters are defined in lander.xacro and are: damping of the arm joints, kp and kd, maximum number of contacts. In this PR, I lower the real_time_update_rate from 400 to 160 Hz, which translates into an increase of the max_time_step from 0.0025 to 0.00625 s. I do this in all the four worlds in ow_europa. To compensate for instability I decreased the "damping" factor for the arm joints from 50 to 10. Even though, after decreasing the damping, the simulation was stable, I noticed that this error message would appear on terminal while there was contact between the scoop and the terrain (in ReferenceMission1, between the end of guarded move and the beginning of grind):
`ODE Message 3: LCP internal error, s <= 0`
I did some research and this is what I found out:
- ODE, the physics engine, solves a Linear Complementarity Problem. When the contact force between two objects is too large (high impact speed, high mass, etc.) ODE cannot solve the problem. In the past we saw this error after guarded_move would find the ground and bounce.
- Another source of this error can be a too large number of contact points for a specific contact (scoop-ground for instance)
- When getting this error, the sim doesn't crash. From my observation, I think this is what happens: the solver throws the error because it doesn't find a solution for that specific time step, so the arm freezes temporarily until a new solution is found. This results in a less smooth motion, which is anyway almost imperceptible.

This is how I got rid of this error:
- forcing the scoop to have a single point of contact with the terrain
- removing kp and decreasing kd for the scoop-terrain contact

I tried to decrease the real_time_update_rate below 160Hz but I got too much instability: the lander would explode right after launching the sim. 

**Test:**
- launch europa_terminator_workspace or atacama
- $ roslaunch ow_autonomy autonomy_node.launch plan:=ReferenceMission1.plx
- check that the sim is visually stable and check that there's no ODE error shows up. Let me know how it goes. 
- General thoughts/feedback is appreciated, thanks!

### DIFF
--- a/ow_lander/urdf/lander.xacro
+++ b/ow_lander/urdf/lander.xacro
@@ -544,12 +544,11 @@
     </visual>
     <selfCollide>1</selfCollide>
     <collision>
-      <max_contacts>1</max_contacts>
       <surface>
         <contact>
           <collide_bitmask>0x0001</collide_bitmask> <!-- Enable collision with the terrain by matching the bitmask -->
           <ode>
-            <kd>5e11</kd>
+            <kd>1e12</kd>
           </ode>
         </contact>
       </surface>
--- a/ow_lander/urdf/lander.xacro
+++ b/ow_lander/urdf/lander.xacro
@@ -164,7 +164,7 @@
       link="l_shou" />
     <axis
       xyz="0 0 -1" />
-    <dynamics damping="50.0" friction="2.6"/>
+    <dynamics damping="10.0" friction="2.6"/>
     <limit effort="70.9" velocity="0.15" lower="-1.8" upper="1.8"/>
   </joint>
   <transmission name="shou_yaw_transmission">
@@ -240,7 +240,7 @@
       link="l_prox" />
     <axis
       xyz="0 0 -1" />
-    <dynamics damping="50.0" friction="4.6"/>
+    <dynamics damping="10.0" friction="4.6"/>
     <limit effort="101.0" velocity="0.2" lower="0.05" upper="2.2"/>
   </joint>
   <transmission name="shou_pitch_transmission">
@@ -316,7 +316,7 @@
       link="l_dist" />
     <axis
       xyz="0 0 -1" />
-    <dynamics damping="50.0" friction="2.7"/>
+    <dynamics damping="10.0" friction="2.7"/>
     <limit effort="73.0" velocity="0.2"/>
   </joint>
   <transmission name="prox_pitch_transmission">
@@ -392,7 +392,7 @@
       link="l_wrist" />
     <axis
       xyz="0 0 1" />
-    <dynamics damping="50.0" friction="0.3"/>
+    <dynamics damping="10.0" friction="0.3"/>
     <limit effort="25.7" velocity="0.2" />
   </joint>
   <transmission name="dist_pitch_transmission">
@@ -480,7 +480,7 @@
       link="l_hand" />
     <axis
       xyz="0 0 -1" />
-    <dynamics damping="50.0" friction="0.3"/>
+    <dynamics damping="10.0" friction="0.3"/>
     <limit effort="25.7" velocity="0.2" />
   </joint>
   <transmission name="hand_yaw_transmission">
@@ -544,12 +544,12 @@
     </visual>
     <selfCollide>1</selfCollide>
     <collision>
+      <max_contacts>1</max_contacts>
       <surface>
         <contact>
           <collide_bitmask>0x0001</collide_bitmask> <!-- Enable collision with the terrain by matching the bitmask -->
           <ode>
-            <kp>1e+15</kp>
-            <kd>1e+13</kd>
+            <kd>5e11</kd>
           </ode>
         </contact>
       </surface>
@@ -586,7 +586,7 @@
       link="l_scoop" />
     <axis
       xyz="0 0 -1" />
-    <dynamics damping="50.0" friction="0.5"/>
+    <dynamics damping="10.0" friction="0.5"/>
     <limit effort="20.0" velocity="0.2" />
   </joint>
   <transmission name="scoop_yaw_transmission">


mogumbo:
Just to clarify for other reviewers: decreasing the `real_time_update_rate` does not introduce instability. Increasing the `max_step_size` is what causes the instability. `real_time_update_rate` is just a dial that lets you run the sim equal to, faster, or slower than real-time.

Also, don't forget to checkout the companion branch on ow_europa before running tests.

ow_lander/urdf/lander.xacro
```diff
@@ -544,12 +544,12 @@
...
     <selfCollide>1</selfCollide>
     <collision>
+      <max_contacts>1</max_contacts>
       <surface>
         <contact>
           <collide_bitmask>0x0001</collide_bitmask> <!-- Enable collision with the terrain by matching the bitmask -->
           <ode>
-            <kp>1e+15</kp>
-            <kd>1e+13</kd>
+            <kd>5e11</kd>
```
mogumbo, line 552: removing kp just reverts it to the default value. Out of curiosity, what is the default value?

ow_lander/urdf/lander.xacro
```diff
@@ -544,12 +544,12 @@
     </visual>
     <selfCollide>1</selfCollide>
     <collision>
+      <max_contacts>1</max_contacts>
```
mogumbo, line 547: I'm concerned about this because it will might result in unrealistic collisions. How is the one contact point chosen--Is it just the contact point where penetration is deepest? And what if a very different contact point is chosen on the next frame--couldn't that cause jerking or popping motions?

ow_lander/urdf/lander.xacro
damianacatanoso: default kp=1e+12 , kd=1 . More info here: http://sdformat.org/spec?ver=1.6&elem=collision#contact_ode

ow_lander/urdf/lander.xacro
damianacatanoso: I was hesitant as well and I wanted to hear your opinion. From testing, the scoop-terrain interaction doesn't seem to change after reducing the max_contacts: I didn't observe any jerking/popping. I am not sure how the contact point is chosen. When setting max_contacts to a number greater than 1, the ODE error shows up. If we decide to run the sim at timestep=0.005 s I might be able to increase the contact points without getting the error.

mogumbo: I would leave this for a separate ticket as it should be tested in more situations before settling on a solution. The LCP errors are annoying but don't seem to cause any real damage. Please make a ticket for suppressing LCP errors and include two possible solutions: 1) max_contacts, or 2) https://github.com/osrf/gazebo/blob/6fd426b3949c4ca73fa126cde68f5cc4a59522eb/gazebo/physics/ode/ODEPhysics_TEST.cc#L336
Or there may be other solutions we have not discovered.

damianacatanoso: Done!

mogumbo (approved):

kmdalal (approved):
Everything looked fine to me.  The detailed writeup in this PR strikes me as perhaps its most valuable contribution.   Would it make sense to put some of this information in the User Guide, say in a Troubleshooting section or Appendix?

damianacatanoso:
> Everything looked fine to me. The detailed writeup in this PR strikes me as perhaps its most valuable contribution. Would it make sense to put some of this information in the User Guide, say in a Troubleshooting section or Appendix?

Yes, I think it would, but not right now: the work is not mature enough to give a complete explanation. I am in the process of understanding two interconnected processes: 1) Increase the time step and tune proper parameters to keep the simulation stable, and 2) increase the real time factor for sim speed up. I figured out the first but not the second yet. When the second will be understood, I'll write a paragraph on this. In the meantime I am storing my "notes", including this PR description, here: https://babelfish.arc.nasa.gov/confluence/display/OCEANWATERS/Simulation+Speed .

Samahu:
> * forcing the scoop to have a single point of contact with the terrain

It is not clear to me how you forced the scoop the scoop to have a single point of contact using the changes in the PR and the linked one, can you elaborate?

Samahu:
> Just to clarify for other reviewers: decreasing the `real_time_update_rate` does not introduce instability. Increasing the `max_step_size` is what causes the instability. `real_time_update_rate` is just a dial that lets you run the sim equal to, faster, or slower than real-time.

I am personally aware of what a physics engine and how it is typically implemented.

damianacatanoso:
> > * forcing the scoop to have a single point of contact with the terrain
> 
> It is not clear to me how you forced the scoop the scoop to have a single point of contact using the changes in the PR and the linked one, can you elaborate?

Sure! Look at commit d563a079[...]. In lander.xacro, under <l_scoop>,<gazebo>,<collision>, you'll see: <max_contacts>1</max_contacts> .
This forces the scoop-gazebo contact have just one point. If you look at Terry's comments history you'll see that we decided to remove this line. Let me know if this clarifies.

Samahu (approved):
I haven't observed any new problems based on the limited amount of testing that I carried out.

Samahu:
> Sure! Look at commit [d563a07](https://github.com/nasa/ow_simulator/commit/d563a079fa752d8f5acac2f47f9e4639cca7db6b)[...]. In lander.xacro, under l_scoop>gazebo>collision , you'll see: <max_contacts>1</max_contacts> .
> This forces the scoop-gazebo contact have just one point. If you look at Terry's comments history you'll see that we decided to remove this line. Let me know if this clarifies.

The history was collapsed thus I didn't go through it

PR #38, opened 2020-11-17
Document PLEXIL version needed for installation

kmdalal:
This is just a documentation change, for the corresponding pull request in ow_autonomy: https://github.com/nasa/ow_autonomy/pull/11

### DIFF
--- a/oceanwaters/doc/setup_dev_env.md
+++ b/oceanwaters/doc/setup_dev_env.md
@@ -60,13 +60,9 @@ of PLEXIL.  On rare occasions the latest version of this branch won't be
 compatible with OceanWATERS and you will need to check out a particular commit.
 We plan to simplify the workflow needed to keep these two software systems compatible.
 
-* UNTIL FURTHER NOTICE (i.e. the absence of this step), check out an older
-  commit of PLEXIL that supports OceanWATERS.  The newest version of the
-  releases/plexil-4 branch will not work.
-
-```
-git checkout 51813f1
-```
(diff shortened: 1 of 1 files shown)


lanseafood (approved):
👍🏻

kmdalal:
NOTE:  path_planning_commander.py was inadvertently included in the PR branch (OCEANWATER-496).  This is because this branch was forked from master and not melodic-devel, and master is not merged into melodic-devel.    It can be ignored, as it's already a released change.

ow_lander/scripts/path_planning_commander.py
```diff
@@ -33,9 +33,9 @@ def __init__(self):
     robot = moveit_commander.RobotCommander()
     scene = moveit_commander.PlanningSceneInterface()
 
-    move_arm = moveit_commander.MoveGroupCommander("arm")
-    move_limbs = moveit_commander.MoveGroupCommander("limbs")
-    move_grinder = moveit_commander.MoveGroupCommander("grinder")
+    move_arm = moveit_commander.MoveGroupCommander("arm", wait_for_servers=15.0)
```
mogumbo, line 36: This is confusing me. The wait_for_servers pieces are not in melodic-devel, which implies they were either added as a hotfix to master but not melodic-devel or they were included in master at release time and later removed from melodic-devel. If the latter, we should not include this change.

kmdalal: After the merges (master -> melodic-devel -> OW-496) this diff is gone and everything looks as expected!

mogumbo (approved):

PR #42, opened 2020-12-04
Oceanwater 552 faults remapping

lanseafood:
image
New Faults structure
/og_joint_states has original data from gazebo, pristine
/joint_states is now published by faults node, and relevant subscribers listen to it. 

What the node and topic connection looks like now
image


Samahu:
Is dynamic_reconfigure.client sort of a binary file?

ow_faults/src/FaultInjector.cpp
```diff
@@ -12,8 +12,8 @@ using namespace ow_lander;
 
 FaultInjector::FaultInjector(ros::NodeHandle node_handle)
 {
-  m_joint_state_sub = node_handle.subscribe("/joint_states", 10, &FaultInjector::jointStateCb, this);
-  m_joint_state_pub = node_handle.advertise<sensor_msgs::JointState>("/faults/joint_states", 10);
+  m_joint_state_sub = node_handle.subscribe("/og_joint_states", 10, &FaultInjector::jointStateCb, this);
+  m_joint_state_pub = node_handle.advertise<sensor_msgs::JointState>("/joint_states", 10); // 'new' joint states topic where fault data is now publishes on
```
Samahu, line 16: per ROS CppStyle max line length is 120 characters

ow_faults/include/ow_faults/FaultInjector.h
```diff
@@ -32,6 +32,7 @@ class FaultInjector
   // Output /faults/joint_states, a modified version of /joint_states, injecting
   // simple message faults that don't need to be simulated at their source.
   void jointStateCb(const sensor_msgs::JointStateConstPtr& msg);
+  void testFxn(const sensor_msgs::JointStateConstPtr& msg);
```
Samahu, line 35: I understand that's this is a test but I think you can come up with a more descriptive name.
What is Fxn anyway!

ow_testbed/launch/spawn.launch
```diff
@@ -25,6 +25,7 @@
   <!-- Convert joint states from Gazebo to tf-tree for rviz -->
   <node name="robot_state_publisher" pkg="robot_state_publisher" type="robot_state_publisher" >
     <param name="publish_frequency" value="30" />
+    <remap from="/joint_states" to="/faults2/joint_states" />
```
Samahu, line 28: This file is not used by the simulation I would suggest you undo this change to stay on what matters.

ow_faults/src/FaultTestSub.cpp
```diff
@@ -0,0 +1,65 @@
...
+   * object go out of scope, this callback will automatically be unsubscribed from
+   * this topic.
+   *
+   * The second parameter to the subscribe() function is the size of the message
+   * queue.  If messages are arriving faster than they are being processed, this
+   * is the number of messages that will be buffered up before beginning to throw
+   * away the oldest ones.
+   */
+// %Tag(SUBSCRIBER)%
+  ros::Subscriber sub = n.subscribe("/joint_states", 10, chatterCallback);
```
Samahu, line 51: Use proper name for the callback function

ow_faults/src/FaultTestSub.cpp
```diff
@@ -0,0 +1,65 @@
...
+   * called chatterCallback.  subscribe() returns a Subscriber object that you
+   * must hold on to until you want to unsubscribe.  When all copies of the Subscriber
+   * object go out of scope, this callback will automatically be unsubscribed from
+   * this topic.
+   *
+   * The second parameter to the subscribe() function is the size of the message
+   * queue.  If messages are arriving faster than they are being processed, this
+   * is the number of messages that will be buffered up before beginning to throw
+   * away the oldest ones.
+   */
```
Samahu, line 49: Remove tutorial comments ..
```suggestion
  /**
   * The subscribe() call is how you tell ROS that you want to receive messages
   * on a given topic.  This invokes a call to the ROS
   * master node, which keeps a registry of who is publishing and who
   * is subscribing.  Messages are passed to a callback function, here
   * called chatterCallback.  subscribe() returns a Subscriber object that you
   * must hold on to until you want to unsubscribe.  When all copies of the Subscriber
   * object go out of scope, this callback will automatically be unsubscribed from
   * this topic.
   *
   * The second parameter to the subscribe() function is the size of the message
   * queue.  If messages are arriving faster than they are being processed, this
   * is the number of messages that will be buffered up before beginning to throw
   * away the oldest ones.
   */
```

Samahu (changes requested):
This seems too early for review

ow_faults/include/ow_faults/FaultInjector.h
lanseafood: fxn is shorthand for function. And yea I can remove this, I think we should delete

ow_faults/src/FaultInjector.cpp
lanseafood: will adjust

ow_faults/src/FaultTestSub.cpp
lanseafood: i'll remove this file, I think it's not relevant anymore

kmdalal:
Lanssie, is there a convenient way to test this, and if so could you describe how?

ow_faults/include/ow_faults/FaultInjector.h
kmdalal: Never seen that one, and it confused me too.   "Fn" has been a standard abbreviation for ages.  I would naturally read "testFn" as "test function".

ow/launch/common.launch
```diff
@@ -4,6 +4,7 @@
   <arg name="world_name" default="$(find ow_europa)/worlds/terminator.world"/>
   <arg name="arm_sim_enable" default="false" />
   <arg name="gazebo" default="true"/>
+  <arg name="extra_gazebo_args" value="joint_states:=/og_joint_states" />
```
kmdalal, line 7: Assuming 'og' means original, 'orig' is much more understandable.

lanseafood:
> Is dynamic_reconfigure.client sort of a binary file?

I think so? I figured any binaries would get caught in a .gitignore but maybe this one did not.

ow_faults/src/FaultInjector.cpp
kmdalal: line too long

ow_faults/src/FaultTestSub.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * This tutorial demonstrates simple receipt of messages over the ROS system.
+ */
+// %Tag(CALLBACK)%
+void chatterCallback(const std_msgs::String::ConstPtr& msg)
+{
+  ROS_INFO("I heard: [%s]", msg->data.c_str());
+}
+// %EndTag(CALLBACK)%
```
kmdalal: I'm not familiar with the Tag/EndTag comments.  What are they?

ow_testbed/launch/spawn.launch
kmdalal: What is '/faults2' for?

lanseafood: I think that was another test function! I'll double check and remove those

Samahu: @lanseafood this is outdated .. we are commenting on a piece of code that has been already removed

kmdalal:
Repeating a question from last week, is this change testable, and how?

lanseafood:
> Repeating a question from last week, is this change testable, and how?

Hey @kmdalal, sorry I missed your question earlier! I believe all you need to do is check out the branch and roslaunch. If you look at the node graph in rviz, you should be able to see the relinking. Since we're not injecting any false values at the start, things should be performing as usual. 

If you rostopic echo /joint_states, you'll see the nominal state. If you enable any fault from rqt and look at the rostopic echo /joint_states, you'll also be able to see that reflected in the telemetry.

ow_faults/src/FaultInjector.cpp
```diff
@@ -12,8 +12,9 @@ using namespace ow_lander;
 
 FaultInjector::FaultInjector(ros::NodeHandle node_handle)
 {
-  m_joint_state_sub = node_handle.subscribe("/joint_states", 10, &FaultInjector::jointStateCb, this);
-  m_joint_state_pub = node_handle.advertise<sensor_msgs::JointState>("/faults/joint_states", 10);
+  m_joint_state_sub = node_handle.subscribe("/orig_joint_states", 10, &FaultInjector::jointStateCb, this);
```
mogumbo, line 15: Please make it /original/joint_states. This is more consistent with other ROS topic naming/categorization. And there's really no need to abbreviate "original."

mogumbo (changes requested):

kmdalal:
For some reason in rqt, the Faults section of the dynamic_reconfigure plugin is gone -- there used to be checkboxes to set each fault parameter.   However, I can set faults just fine using dynamic_reconfigure on the command line.

mogumbo:
>  the Faults section of the dynamic_reconfigure plugin is gone

My faults section is still there. Try rebuilding the workspace.

kmdalal (approved):

Samahu:
@lanseafood I maybe missing something but to me this code is not complete unless you forward the data you receive through the remapped topic _/original/joint_stats_ to the original topic _/joint_states_ (unaltered for the time being).

lanseafood:
> @lanseafood I maybe missing something but to me this code is not complete unless you forward the data you receive through the remapped topic _/original/joint_stats_ to the original topic _/joint_states_ (unaltered for the time being).
That is currently the case, the orig data is being unaltered to joint states so long as you don't check a faults.

Samahu:
So I have enabled the faults for some of the joints and called a give arm activity. The lander within RVIZ stays intact (as I would expect since the joints are kind of disabled by the faults), however, within gazebo the arm still performs the activity as if no faults were introduced.. That tells me we were not able to hide the original _joint_states_ from gazebo itself. Can you check if there is a way to make the potentially manipulated _/joint_states_ be the source for gazebo.

mogumbo:
> So I have enabled the faults for some of the joints and called a give arm activity. The lander within RVIZ stays intact (as I would expect since the joints are kind of disabled by the faults), however, within gazebo the arm still performs the activity as if no faults were introduced.. That tells me we were not able to hide the original _joint_states_ from gazebo itself. Can you check if there is a way to make the potentially manipulated _/joint_states_ be the source for gazebo.

Gazebo should show ground truth (what is actually happening), while rviz should show what it thinks is happening based on available ROS topics. As far as I can tell, that's what is happening. Gazebo has its own internal messaging system and we're not affecting it at all by playing around with ROS topics. It might be possible to affect it with a gazebo plugin. Using Gazebo's Topic Visualization from the Window pulldown menu, I don't see any internal topics that show joint states.

mogumbo:
I tested a bit. Sometimes injecting arm faults causes roslaunch ow_autonomy autonomy_node.launch to fail at planning the grinding, so it stows the arm instead. I haven't been able to get it to fail at guarded_move. Most importantly, I haven't gotten it to crash the sim or cause any such non-real-world behavior. So I'm okay with merging this once all comments are addressed.

ow_faults/src/FaultInjector.cpp
Samahu: I agree with Terry. I would even **suggest** (not a strong opinion) to use `/_original/joint_states` if that's possible to indicate that this is topic is for private use. But I don't know if prefixing a topic name with underscore is permissible.

lanseafood:
some output examples
ant pan encoder failure:
* looks like we get a warning on console that the antennae is having the expected error, rvix and gazebo still moving
image

shoulder yaw encoder failure
*looks like we also get a warning on console as well here. rviz and gazebo all still move.
image

lanseafood:
> So I have enabled the faults for some of the joints and called a give arm activity. The lander within RVIZ stays intact (as I would expect since the joints are kind of disabled by the faults), however, within gazebo the arm still performs the activity as if no faults were introduced.. That tells me we were not able to hide the original _joint_states_ from gazebo itself. Can you check if there is a way to make the potentially manipulated _/joint_states_ be the source for gazebo.

@Samahu  i believe there is a future ticket to work on https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-553?filter=19030. I think the scope of this ticket is just remapping joint_states

ow_faults/src/FaultInjector.cpp
lanseafood: will do!

mogumbo (approved):

Samahu (approved):
Good work :+1:

PR #43, opened 2020-12-09
Increase real time factor - Test

damianacatanoso:
Hi everyone, I am opening this PR because I am having an issue with ground detection and I was hoping if you could run a test and report if you have the same issue. NOTE: this branch is not ready for merging yet so please don't focus on the code. The issue is that when the sim runs as fast as it can (update_rate set to 0), all services work fine except guarded_move, for which I consistently get failure in detection. Ussama tested this branch and he gets successful detection most of the times. 

TEST:
- in terminator_workspace.world set update rate to 0, as follows:
`<real_time_update_rate>0</real_time_update_rate>`
- launch europa_terminator_workspace.launch . The sim now should run as fast as it can
- plan guarded_move
- publish guarded_move
- quickly (you should do this fast, before the scoop touches the ground) in another terminal enter: 
`rostopic echo guarded_move_result`
- Check if it successfully finds the ground
- Repeat last 3 steps 10 times and report how many times you got successful ground detection.

Thank you so much and let me know if you have any questions!

Feedback from Ussama:
1) It is not always easy to produce  the missed or the premature ground detection problem on my machine .. most of the time I need to try 6 or 7 times till I produce one.
2) The missed ground detection happens when the initial 5 samples fluctuate significantly in their values resulting in a larger stdev.. which raises the bar for the trending value to break the threshold. The question is why this stronger variation at the 5 initial samples is taking place when the speed is enabled? I have no clue currently. You may try to increase the minimum number of sample to 6 or 7 to see if that helps. Note that this would slightly delay the ground detection.
3) The premature ground detection happens less frequent and can also be attributed to insufficient number of samples used to produce the dynamic threshold values (mean+tolerance). Increasing the number to 6 or 7 applies
- I did test with 6 and still resulted in missed detection
- I did try 7 too and although it does seem to improve things but eventually I got a missed detection after ~10 attempts
@Samahu, you don't need to test again, I added you to this PR so you can follow the discussion.

### DIFF
--- a/ow/launch/atacama_y1a.launch
+++ b/ow/launch/atacama_y1a.launch
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <launch>
-
+  <arg name="use_sim_time" value="true"/>
   <arg name="arm_sim_enable" default="false"/>
   <arg name="gazebo" default="true"/>
   <arg name="gzclient" default="true"/>
--- a/ow_dynamic_terrain/scripts/modify_terrain_grinder_pub.py
+++ b/ow_dynamic_terrain/scripts/modify_terrain_grinder_pub.py
@@ -7,7 +7,6 @@ and issuing a corresponding modify_terrain_* message to update the terrain.
 """
 
 import rospy
-import time
 from math import degrees
 import numpy as np
 from geometry_msgs.msg import Point
@@ -19,6 +18,7 @@ from ow_dynamic_terrain.msg import modify_terrain_circle
 class ModifyTerrainGrinder:
 
   def __init__(self, *args):
+    rospy.set_param('gazebo/use_sim_time', True)
     rospy.init_node("modify_terrain_grinder_pub", anonymous=True)
     self.last_translation = np.zeros(3)
     self.pub_visual = rospy.Publisher(
--- a/ow_ephemeris/scripts/sunlight_values_for_shaders.py
+++ b/ow_ephemeris/scripts/sunlight_values_for_shaders.py
@@ -36,6 +36,7 @@ class SunValues:
 
     # Ignore missing transforms--there will probably be a few missing at startup
     try:
+      self.listener.waitForTransform(self.reference_body, self.sun, rospy.Time(0), rospy.Duration(20.0))
       (trans,rot) = self.listener.lookupTransform(self.reference_body, self.sun, rospy.Time(0))
     except:
       return
@@ -63,6 +64,7 @@ class SunValues:
 
 
 def main():
+  rospy.set_param('gazebo/use_sim_time', True)
   rospy.init_node("sunlight_values")
 
   sun_values = SunValues()
--- a/ow_lander/launch/fake_moveit_controller_manager.launch
+++ b/ow_lander/launch/fake_moveit_controller_manager.launch
@@ -1,5 +1,7 @@
 <launch>
 
+  <arg name="use_sim_time" value="true"/>
+
   <!-- Set the param that trajectory_execution_manager needs to find the controller plugin -->
   <param name="moveit_controller_manager" value="moveit_fake_controller_manager/MoveItFakeControllerManager"/>
 
--- a/ow_lander/launch/gazebo.launch
+++ b/ow_lander/launch/gazebo.launch
@@ -1,5 +1,7 @@
 <launch>
 
+  <arg name="use_sim_time" value="true"/>
+
   <include file="$(find gazebo_ros)/launch/empty_world.launch" />
 
   <!-- Load urdf -->
--- a/ow_lander/launch/move_group.launch
+++ b/ow_lander/launch/move_group.launch
@@ -1,6 +1,8 @@
 <launch>
-
+  <!-- <arg name="use_sim_time" value="true"/> -->
+  <arg name="use_sim_time" value="true"/>
   <include file="$(find ow_lander)/launch/planning_context.launch" />
+
 
   <!-- GDB Debug Option -->
   <arg name="debug" default="false" />
--- a/ow_lander/scripts/ground_detection.py
+++ b/ow_lander/scripts/ground_detection.py
@@ -4,7 +4,7 @@
 # Research and Simulation can be found in README.md in the root directory of
 # this repository.
 
-import time
+# import time
 import numpy as np
 import rospy
 import tf2_ros
@@ -83,16 +83,20 @@ class GroundDetector:
     if self._last_position is None:
       self._last_position = np.array(
           [new_position.x, new_position.y, new_position.z])
-      self._last_time = time.time()
+      self._last_time = rospy.get_time()
       return False
     current_position = np.array(
         [new_position.x, new_position.y, new_position.z])
-    current_time = time.time()
+    current_time = rospy.get_time()
+    # print (current_time) #This is rostime, published by /clock
     delta_d = current_position - self._last_position
+    # print (delta_d)
     delta_t = current_time - self._last_time
+    # print (delta_t)
     self._last_position, self._last_time = current_position, current_time
     velocity_z = delta_d[2] / delta_t
     self._trending_velocity.append(velocity_z)
+    print(self._dynamic_threshold)
     if self._dynamic_threshold is None:
       if self._trending_velocity.valid:
         self._dynamic_threshold = self._trending_velocity.value
@@ -137,6 +141,7 @@ class GroundDetector:
     return t.transform.translation
 
   def _tf_2_method(self):
+    # print (self._tf_lookup_position().z)
 
     return self._check_condition(self._tf_lookup_position())
 
@@ -155,4 +160,5 @@ class GroundDetector:
     :returns: True if ground was detected, False otherwise
     :rtype: boolean
     """
+    # print(rospy.get_time())
     return self._check_ground_method()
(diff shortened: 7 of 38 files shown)


Samahu:
I just want to add one more note that I have noticed during my testing of this branch. The last part of the guarded move i.e when the arm starts descending downwards the move seems to be happening as  bursts of movements which can contribute to the problem of missing ground detection. The same behavior is observable on the _melodic-devel_ branch so this behavior is not something that has been introduced by this branch. I am not sure at what stage this change happened but I earlier in the project these bursts were not that much observable.

You can clearly see the difference on the OCEANWATER-518 branch.

ow_dynamic_terrain/scripts/modify_terrain_grinder_pub.py
```diff
@@ -19,6 +18,7 @@
 class ModifyTerrainGrinder:
 
   def __init__(self, *args):
+    rospy.set_param('gazebo/use_sim_time', True)
```
Samahu, line 21: Is this really a parameter that is consumed by gazebo? Do you have any reference on this?

ow_ephemeris/scripts/sunlight_values_for_shaders.py
```diff
@@ -36,6 +36,7 @@ def computeIntensity(self):
 
     # Ignore missing transforms--there will probably be a few missing at startup
     try:
+      self.listener.waitForTransform(self.reference_body, self.sun, rospy.Time(0), rospy.Duration(20.0))
```
Samahu, line 39: Why did you have to add this line? Couldn't you just add a duration(20.0) param to the lookupTransform on next statement?

ow_lander/launch/fake_moveit_controller_manager.launch
```diff
@@ -1,5 +1,7 @@
 <launch>
 
+  <arg name="use_sim_time" value="true"/>
```
Samahu, line 3: Really has no effect

ow_lander/launch/gazebo.launch
```diff
@@ -1,5 +1,7 @@
 <launch>
 
+  <arg name="use_sim_time" value="true"/>
```
Samahu, line 3: Same as others .. this has no effect

ow_lander/launch/move_group.launch
```diff
@@ -1,7 +1,9 @@
 <launch>
-
+  <!-- <arg name="use_sim_time" value="true"/> -->
+  <arg name="use_sim_time" value="true"/>
```
Samahu, line 3: most likely has no effect

ow_lander/scripts/ground_detection.py
```diff
@@ -83,16 +83,20 @@ def _check_condition(self, new_position):
...
       self._last_position = np.array(
           [new_position.x, new_position.y, new_position.z])
-      self._last_time = time.time()
+      self._last_time = rospy.get_time()
       return False
     current_position = np.array(
         [new_position.x, new_position.y, new_position.z])
-    current_time = time.time()
+    current_time = rospy.get_time()
+    # print (current_time) #This is rostime, published by /clock
```
Samahu, line 91: Remove debug statements
```suggestion
```

ow_lander/urdf/lander.xacro
```diff
@@ -548,10 +548,12 @@
         <contact>
           <collide_bitmask>0x0001</collide_bitmask> <!-- Enable collision with the terrain by matching the bitmask -->
           <ode>
-            <kd>1e12</kd>
+            <kd>5e12</kd>
+            <!-- <kp>1e20</kp> -->
           </ode>
         </contact>
       </surface>
+      <max_contacts>1</max_contacts>
```
Samahu, line 556: I thought that we discussed that reducing the number of contacts to 1 even though hides the LCP error message but is unlikely good for the accuracy of the results. Do you have new findings on this?

Samahu (commented):
I had a bunch of comments mostly has to do with the use of **use_sim_time**. I don't think we need to spread the parameter on every node .. I am pretty certain that it is a server parameter that needs to be set once for the entire session.

mogumbo:
> I had a bunch of comments mostly has to do with the use of **use_sim_time**. I don't think we need to spread the parameter on every node .. I am pretty certain that it is a server parameter that needs to be set once for the entire session.

Correct. Setting use_sim_time to true when launching gazebo causes it to be set on ROS's global parameter server. Every node should use it. ROS2 behaves different and requires you to set it on every node individually. It's painful and OSRF is getting plenty of pressure to make it easier on ROS2 as well.

ow/launch/atacama_y1a.launch
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <launch>
-
+  <arg name="use_sim_time" value="true"/>
```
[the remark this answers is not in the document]
damianacatanoso: Yes I agree, I was going to look into this later though, once more urgent problems are solved..

mogumbo:
I won't look at the code yet since you say it's not ready. I got True for 7 out of 10 ground detections. I only saw one ground detection that appeared to be instantaneous. In the other 9 runs, the scoop scooted across the ground as it was trying to descend deeper. In the other runs the scoop scooted across the ground a little or a lot. In a couple of the True runs, the scoop scooted almost as far as the False runs.

It would be interesting to see if the smoother motion from OCEANWATER-518 affects behavior. You could try to merge it into your branch and run this test again. Of course, don't push that merge since OCEANWATER-518 isn't finished yet.

Samahu:
> I won't look at the code yet since you say it's not ready. I got True for 7 out of 10 ground detections. I only saw one ground detection that appeared to be instantaneous. In the other 9 runs, the scoop scooted across the ground as it was trying to descend deeper. In the other runs the scoop scooted across the ground a little or a lot. In a couple of the True runs, the scoop scooted almost as far as the False runs.
> 
> It would be interesting to see if the smoother motion from OCEANWATER-518 affects behavior. You could try to merge it into your branch and run this test again. Of course, don't push that merge since OCEANWATER-518 isn't finished yet.

@damianacatanoso If you want to try what you currently have against OCEANWATER-518 as Terry suggested, you would need to re-enable the 'trajectory_publisher' node which runs the ground detection code.

Samahu:
> I just want to add one more note that I have noticed during my testing of this branch. The last part of the guarded move i.e when the arm starts descending downwards the move seems to be happening as bursts of movements which can contribute to the problem of missing ground detection. The same behavior is observable on the _melodic-devel_ branch so this behavior is not something that has been introduced by this branch. I am not sure at what stage this change happened but I earlier in the project these bursts were not that much observable.
> 
> You can clearly see the difference on the OCEANWATER-518 branch.

A video that shows the bursted moves during "guraded move": https://youtu.be/eVLpBiXo7sA

Samahu:
> @damianacatanoso If you want to try what you currently have against OCEANWATER-518 as Terry suggested, you would need to re-enable the 'trajectory_publisher' node which runs the ground detection code.

I did review yesterday what needs to be done to enable ground_detection on OCEANWATER-518 branch, and unfortunately it is not that trivial. You won't be able to check how changes work on OCEANWATER-518 with ground detection enable unless the [OCEANWATER-540 | Factor out ground detection functionality out of trajectory_publisher](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-540) is completed. Let me know if you are blocked by this.

ow_dynamic_terrain/scripts/modify_terrain_grinder_pub.py
damianacatanoso: I wrote about this here: https://babelfish.arc.nasa.gov/confluence/display/OCEANWATERS/Increase+Real+Time+Factor 

Here is the reference:  http://wiki.ros.org/Clock

ow_lander/urdf/lander.xacro
damianacatanoso: As I said, this PR wasn't meant for a code review.. I was just testing if things improved reducing the number of contacts.. I am not planning to use this in the final version of this branch. I'll push the final version soon.

damianacatanoso:
@Samahu and @mogumbo, thank you very much for your feedback. As discussed with Ussama yesterday, at this point, it doesn't make sense to debug this branch with the old controller or even merging an incomplete trajectory controller. I'll just put this work in standby and use my time to help Ussama finish the trajectory controller. I'll clean this branch and rename it using the "archive" tag. When the trajectory controller will be complete, I'll resume this work, merging the final trajectory controller branch and debugging it. If I don't manage to get to it before I leave, the next person can refer to the documentation I wrote on Confluence and the branch itself.

ow_dynamic_terrain/scripts/modify_terrain_grinder_pub.py
Samahu: Damiana, my question was specifically on the use of **"gazebo/use_sim_time"**. Is this an actual topic that Gazebo makes use? I couldn't see any info on this topic besides your own confluence doc.

damianacatanoso: Here is a $ rosparam list:
image

Samahu: This parameter is not found on _melodic-devel branch_. If you are on the _increase_real_time_factor branch_ then this param would just show up because we are introducing it to the simulation as part of this statement `rospy.set_param('gazebo/use_sim_time', True)` (here and in other python nodes).

PR #41, opened 2020-12-04
Faults remapping

lanseafood:
image
New Faults structure
/og_joint_states has original data from gazebo, pristine
/joint_states is now published by faults node, and relevant subscribers listen to it. 

What the node and topic connection looks like now
image

PR #45, opened 2020-12-10
OceanWaters 554 Faults Messaging System

lanseafood:
## Linked Issues:
| Jira Ticket 🎟️   | [Oceanwater-554](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-554) |
| ----------- | ----------- |
| EPIC ⚡| [Oceanwater-551](https://babelfish.arc.nasa.gov/jira/projects/OCEANWATER/issues/OCEANWATER-551?filter=reportedbyme) |
| Github :octocat:  | #41 ✔️ |

## Summary of Changes

Faults node hs two new publishers, the system faults and arm faults. Two new msg types were also created for these topics.
image

Publishing Messages:
System Faults topic with no faults selected
image

System Faults topic with  faults selected
image

Arm Faults topic with faults slected
image

## Test
To test this, simply run a launch file, and rostopic echo the system_faults_status or the arm_faults_status. With no faults selected, they should have value = 0 (no fault). When any faults are selected from rqt, they should return 4 (arm execution fault) and 2 (hardware fault) specifically. Autonomy should listen to these topics for info. Right now the ant faults also result in a 4, but that's because I haven't gotten a chance to ask Anna what the behavior should be if there are multiple faults that occur.


ow_faults/src/FaultInjector.cpp
```diff
@@ -26,6 +32,20 @@ void FaultInjector::faultsConfigCb(ow_faults::FaultsConfig& faults, uint32_t lev
   m_faults = faults;
 }
 
+void FaultInjector::setSystemFaultMessage(ow_faults::SystemFaults& msg) {
+  // for now only arm execution errors
+  msg.header.stamp = ros::Time::now();
+  msg.header.frame_id = '/world';
```
kmdalal, line 38: Should be "/world".  Single quotes are for characters, and this produced a compiler warning for me.

ow_faults/src/FaultInjector.cpp
```diff
@@ -26,6 +32,20 @@ void FaultInjector::faultsConfigCb(ow_faults::FaultsConfig& faults, uint32_t lev
...
+  msg.value = 4;
+}
+
+void FaultInjector::setArmFaultMessage(ow_faults::ArmFaults& msg) {
+  // for now only arm execution errors
+  msg.header.stamp = ros::Time::now();
+  msg.header.frame_id = '/world';
```
kmdalal, line 45: Same as previous comment.

ow_faults/include/ow_faults/FaultInjector.h
```diff
@@ -47,6 +53,9 @@ class FaultInjector
   ros::Subscriber m_joint_state_sub;
   ros::Publisher m_joint_state_pub;
 
+  ros::Publisher fault_status_pub;
+  ros::Publisher arm_fault_status_pub;
```
kmdalal, line 57: Prefix these with "m_" like the other data members -- it's one of our coding guidelines (on Confluence).

ow_faults/src/FaultInjector.cpp
lanseafood: good catch, I was wondering why that was the case. :D will fix

ow_faults/src/FaultInjector.cpp
```diff
@@ -26,6 +32,20 @@ void FaultInjector::faultsConfigCb(ow_faults::FaultsConfig& faults, uint32_t lev
   m_faults = faults;
 }
 
+void FaultInjector::setSystemFaultMessage(ow_faults::SystemFaults& msg) {
+  // for now only arm execution errors
+  msg.header.stamp = ros::Time::now();
+  msg.header.frame_id = '/world';
+  msg.value = 4;
```
mogumbo, line 39: Since you call setSystemFaultMessage from multiple faults, this value can be affected more than once. Shouldn't it be msg.value |= 0x4; ? And should 0x4 be defined with a group of other bitmasks with descriptive names?

ow_faults/src/FaultInjector.cpp
```diff
@@ -26,6 +32,20 @@ void FaultInjector::faultsConfigCb(ow_faults::FaultsConfig& faults, uint32_t lev
...
+}
+
+void FaultInjector::setArmFaultMessage(ow_faults::ArmFaults& msg) {
+  // for now only arm execution errors
+  msg.header.stamp = ros::Time::now();
+  msg.header.frame_id = '/world';
+  msg.value = 1;
```
mogumbo, line 46: Same as previous comment.

mogumbo (changes requested):
Your solution differs a bit from the Jira ticket. For example, you have created messages but not services. The code or the Jira ticket should probably be modified.

ow_faults/src/FaultInjector.cpp
```diff
@@ -26,6 +32,20 @@ void FaultInjector::faultsConfigCb(ow_faults::FaultsConfig& faults, uint32_t lev
   m_faults = faults;
 }
 
+void FaultInjector::setSystemFaultMessage(ow_faults::SystemFaults& msg) {
```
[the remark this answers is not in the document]
lanseafood: will pluralize all

ow_faults/src/FaultInjector.cpp
lanseafood: Hey @mogumbo , here I was following the convention set by JPL. In their messages, the name in question is always 'value'. AFAIK Anna mentioned that these were all int values, which changed depending on the type of error that occured. Last we spoke, I brought up what happens if more than 1 type of error occurs and she said they hadn't thought through that use case yet. I'll ask her but given her OOO status right now, I'm not sure when she'll be back so I left it for now. Let me know if you want to still move away from their convention or stay for now.

lanseafood: To clarify, once we can determine how they want to update the System Fault message, I would adjust this function to determine what the value should be.

lanseafood:
> Your solution differs a bit from the Jira ticket. For example, you have created messages but not services. The code or the Jira ticket should probably be modified.

Ah yes, typo on my end in the ticket. It should say messages not service! I've updated that :) https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-554

ow_faults/msg/ArmFaults.msg
```diff
@@ -0,0 +1,12 @@
...
+Header header
+uint32 value
+# None = 0
+# Hardware = 1
+# Trajectory Generation = 2
+# Collision = 3
+# Estop = 4
+# Position Limit = 5
+# Torque Limit = 6
+# Velocity Lmit = 7
```
kmdalal, line 11: typo Lmit

kmdalal: And a more substantial question: are any other than Hardware implemented?  I get that one when I fault an encoder.

ow_faults/src/FaultInjector.cpp
mogumbo: It appears system fault status values are all power of two like a regular bitmask, but they aren't being used like a bitmask. Arm fault values are not defined like a bitmask at all, but they should be in case more than one fault occurs simultaneously. To me this all appears to have a broken design and we should resolved it before we merge this branch.

I think the values you have commented in the .msg files should be put into enums somewhere instead of being comments. And my strong preference is to resolve the design with Anna before we call this finished.

Ideally, msg.value would use the type of some typedef'ed enum, but that is probably too inconvenient or impossible to do in a ROS message.

ow_faults/msg/ArmFaults.msg
lanseafood: Nope, only hardware right now. JPL hasn't really implemented others beyond hardware and we also decided to focus first on hardware anyways.

ow_faults/src/FaultInjector.cpp
lanseafood: Yes, I think there just hasn't been any work so far that's yet justified that discussion so now is the time to figure it out. I can send an email to Anna to discuss this and hopefully, we can get a response back from her before EOY.

kmdalal: Indeed, Lanssie is following what Anna defined, as I'm working with the same message files, which are in the owlat_msgs package.   It's not clear to me whether the arm faults are disjoint (in which case the increasing numbers are fine) or whether they can occur together (in which case a bitmask makes more sense).  I agree we should get this straightened out with Anna when she is available -- emailing her now is a good idea.   

I'd be fine with merging this to enable continued development on our end, if the possible redefinition of the fault values won't create a headache later.

lanseafood: Sent and cc'd everyone on my email to Anna. Also updated the code to have Enums, so I think we have resovled what we can. With respect to bitmask, we can see what Anna says about how they want to resolve multiple faults in the messages.

ow_faults/include/ow_faults/FaultInjector.h
```diff
@@ -28,11 +30,21 @@ class FaultInjector
 
   void faultsConfigCb(ow_faults::FaultsConfig& faults, uint32_t level);
 
+  enum Nominal {None=0};
+
+  enum ArmFaults {Hardware=1, TrajectoryGeneration=2, Collision=3, Estop=4, PositionLimit=5, TorqueLimit=6, VelocityLimit=7, NoForceData=8};
```
mogumbo, line 35: Please spread these enums onto multiple lines for readability.

lanseafood: done!

mogumbo (approved):

ow_faults/src/FaultInjector.cpp
```diff
@@ -26,6 +29,20 @@ void FaultInjector::faultsConfigCb(ow_faults::FaultsConfig& faults, uint32_t lev
...
+  msg.value = value; //should be 4 for now
+}
+
+void FaultInjector::setArmFaultMessage(ow_faults::ArmFaults& msg, int value) {
+  // for now only arm execution errors
+  msg.header.stamp = ros::Time::now();
+  msg.header.frame_id = "/world";
+  msg.value = value; //should be 1 for now
+}
+
```
kmdalal, line 45: Even in the comments, use the new enum names instead of 1 and 4.

lanseafood: done!

kmdalal (approved):

PR #49, opened 2020-12-14
Resumed PositionController for ant_pan and tilt

damianacatanoso:
Ticket: https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-559

There are two antenna joints: pan and tilt. Both of them don't need planning and are moved independently. So there's no need to switch to the JointTrajectoryController. In this branch I resumed the JointPositionController only for the antenna. 

Test:
- launch europa_terminator_workspace.launch
- in rqt>Control and Viz>Message Publisher click to expand the /ant_pan_position_controller/command topic. Tick the box. Change rate to 100. In the 'expression' field write float numbers such as: 0.5, 0.8, -0.5, one at a time. Observe antenna reaching the goal positions in Gazebo.
- make sure your ow_autonomy is in the OW-518 branch.
- run ReferenceMission1 : roslaunch ow_autonomy autonomy_node.launch plan:="ReferenceMission1.plx"
- observe antenna motion (you'll get an error when Unstow is called, that's normal). 
- Let me know if the antenna moves as expected please

### DIFF
--- a/ow_lander/config/ros_controllers.yaml
+++ b/ow_lander/config/ros_controllers.yaml
@@ -55,22 +55,14 @@ controller_list:
   #     - j_dist_pitch
   #     - j_hand_yaw
   #     - j_grinder
-antenna_controller:
-  type: effort_controllers/JointTrajectoryController
-  joints:
-    - j_ant_pan
-    - j_ant_tilt
-  gains:
-    j_ant_pan:
-      p: 100
-      d: 1
-      i: 1
-      i_clamp: 1
-    j_ant_tilt:
-      p: 100
-      d: 1
-      i: 1
-      i_clamp: 1
+ant_pan_position_controller:
+  type: effort_controllers/JointPositionController
+  joint: j_ant_pan
+  pid: {p: 100.0, i: 0.01, d: 15.0}
+ant_tilt_position_controller:
+  type: effort_controllers/JointPositionController
+  joint: j_ant_tilt
+  pid: {p: 550.0, i: 0.01, d: 70.0}
 arm_controller:
   type: effort_controllers/JointTrajectoryController
   joints:
--- a/ow_lander/config/ros_controllers.yaml
+++ b/ow_lander/config/ros_controllers.yaml
@@ -6,7 +6,6 @@ moveit_sim_hw_interface:
 generic_hw_control_loop:
   loop_hz: 300
   cycle_time_error_threshold: 0.01
-
 # Settings for ros_control hardware interface
 hardware_interface:
   joints:
@@ -20,6 +19,11 @@ hardware_interface:
     - j_grinder
     - j_scoop_yaw
   sim_control_mode: 1  # 0: position, 1: velocity
+# Publish all joint states
+# Creates the /joint_states topic necessary in ROS
+joint_state_controller:
+  type: joint_state_controller/JointStateController
+  publish_rate: 50
 controller_list:
   - name: arm_controller
     action_ns: follow_joint_trajectory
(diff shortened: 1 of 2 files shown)


Samahu (changes requested):
You need to revise and brake the the other definition of **antenna_controller** on line 28 into two controllers as you did to the definition on line 58.

damianacatanoso:
> You need to revise and brake the the other definition of **antenna_controller** on line 28 into two controllers as you did to the definition on line 58.

Can you point it out in the code?

Samahu:
> > You need to revise and brake the the other definition of **antenna_controller** on line 28 into two controllers as you did to the definition on line 58.
> 
> Can you point it out in the code?

GitHub PR system doesn't allow me to reference lines of code away from the code changes. I was referring to line 28 of the file `ow_lander/config/ros_controllers.yaml` (sorry I didn't specify this initially).

damianacatanoso:
> > > You need to revise and brake the the other definition of **antenna_controller** on line 28 into two controllers as you did to the definition on line 58.
> > 
> > 
> > Can you point it out in the code?
> 
> GitHub PR system doesn't allow me to reference lines of code away from the code changes. I was referring to line 28 of the file `ow_lander/config/ros_controllers.yaml` (sorry I didn't specify this initially).

Addressed, thanks!

mogumbo (approved):

Samahu:
@damianacatanoso It seems that the _follow_joint_trajectory_ should have a matching _JointTrajectoryController_ for things to play nicely. Since we are not going to use a JointTrajectoryController for the **antenna tilt and pan joints** - at least for the current release - we need to remove the definition of both `ant_pan_position_controller/follow_joint_trajectory` and `ant_tilt_position_controller/follow_joint_trajectory`.

damianacatanoso:
> @damianacatanoso It seems that the _follow_joint_trajectory_ should have a matching _JointTrajectoryController_ for things to play nicely. Since we are not going to use a JointTrajectoryController for the **antenna tilt and pan joints** - at least for the current release - we need to remove the definition of both `ant_pan_position_controller/follow_joint_trajectory` and `ant_tilt_position_controller/follow_joint_trajectory`.

Addressed.

ow_lander/config/ros_controllers.yaml
```diff
@@ -211,4 +178,15 @@ grinder_controller:
       p: *kp_default
       d: *kd_default
       i: *ki_default
-      i_clamp: *ki_clamp_default
\ No newline at end of file
+      i_clamp: *ki_clamp_default
+joint_state_controller:
+  type: joint_state_controller/JointStateController
+  publish_rate: 300
```
Samahu, line 184: Did you change the publish_rate intentionally? what is the reason for this change?

damianacatanoso: Oh I put it equal to the control loop frequency, but maybe they don't have to be equal. What's the reason why we choose 50?

Samahu: That's the default value set by **moveit setup_assistant**

Samahu: So I am not very in favor of changing as this may have other side effects.

Samahu: I also recommend keeping the block position in the same place as before. The reason if we every run setup assistant again then it wouldn't change as something that has change.

damianacatanoso: ok

Samahu (approved):
Thanks for working this out!

PR #50, opened 2020-12-15
Commented and added new values to JointTrajectoryController constraints

damianacatanoso:
Ticket: https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-529

The aim of this branch is to 1) understand the "constraints" field in ros_controller.yaml file, 2) eliminate the error (see below) caused by these constraints and 3) set reasonable constraints values. Error: 
[ WARN] [1605820189.718456935, 1916524839.670000076]: Controller arm_controller failed with error GOAL_TOLERANCE_VIOLATED:
[ WARN] [1605820189.718527851, 1916524839.670000076]: Controller handle arm_controller reports status ABORTED

1) From here: http://wiki.ros.org/joint_trajectory_controller#Controller_configuration_examples : 
- _constraints/goal_time_ (double, default: 0.0)
If the timestamp of the goal trajectory point is t, then following the trajectory succeeds if it reaches the goal within t +/- goal_time, and aborts otherwise.
- _constraints/stopped_velocity_tolerance_ (double, default: 0.01)
Velocity to be considered approximately equal to zero.
- _constraints/<joint>/goal_ (double, default: 0.0)
Position tolerance for a particular joint to reach the goal. When the joint is within goal_position +/- goal_tolerance, than the trajectory can succeed.
- _constraints/<joint>/trajectory_ (double, default: 0.0)
Position tolerance for a particular joint throughout the trajectory. If the joint position ever falls outside trajectory_position +/- tolerance, then the trajectory aborts.

2) The constraints node was placed hierarchically under "gains", while it should be under "arm_controller". I suspect that the given values were not read, so the default values were applied. In particular, the goal_time had a 0.0 default value and this was the constraint violated. So I relocated the "constraint" field and gave a larger value to the goal_time. 

3) I assigned reasonable values to the constraint variables. Also, I assigned the previous PID values because p=16000 is way too much. NOTE: I didn't tune the PID (I'll do this here: https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-528), I just wrote the values that currently are in melodic-devel.

TEST:
- europa_terminator_workspace.launch
- Run these services one by one: /arm/unstow, /arm/deliver_sample, /arm/stow , and check if you get the error message on main terminal. Thanks!


ow_lander/config/ros_controllers.yaml
```diff
@@ -80,47 +80,54 @@ arm_controller:
     - j_dist_pitch
     - j_hand_yaw
     - j_scoop_yaw
+  constraints:
+    # If the timestamp of the goal trajectory point is t, then following the trajectory 
+    # succeeds if it reaches the goal within t +/- goal_time, and aborts otherwise.
+    goal_time: 50.0
```
Samahu, line 86: Based on your comment this is a tolerance value and I presume it is in seconds; What is the reason you picked such a high value. Ultimately if trajectory execution takes time substantially longer than expected then we want to controller to abort the execution.

Samahu (commented):
@damianacatanoso Please explain the reason you picked 50 seconds for `goal_time` constraint.

ow_lander/config/ros_controllers.yaml
Samahu: ?

Samahu (approved):
Great Job! :rocket:

ow_lander/config/ros_controllers.yaml
damianacatanoso: Just copying-pasting from our conversation, for the record: "So I think I opted for a high value just to be conservative in special cases. For instance sometimes it happens that the scoop, for example while digging, gets momentarily stuck in the trench and then eventually it comes out. 50s is not that long, it's not even one minute. I don't think we really need to have a strict time constraint."

Samahu: Yes, it is not a big deal, maybe once we have more experience with it we can revise the values again.

PR #56, opened 2020-12-22
set pid values once for all controllers +

Samahu:
## Summary of Changes
This is mainly a quick refactor of the `ros_controllers.yaml` to avoid setting different PID values for the same joint in different controllers.
* Set pid values once - of the same joint - for all controllers
* Reference only joints that belong to the controller in constraints section

## Test
```bash
roslaunch ow atacama_y1a.launch # if this starts then all good
```

### DIFF
--- a/ow_lander/config/ros_controllers.yaml
+++ b/ow_lander/config/ros_controllers.yaml
@@ -90,29 +90,29 @@ arm_controller:
   stop_trajectory_duration: 1.0
   gains:
     j_shou_yaw: 
-      p: 1000.0
-      d: 300.0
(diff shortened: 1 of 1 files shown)


Samahu:
@damianacatanoso I found that multiplying the `P` element values by 10 helps reduce the issue we are observing on OCEANWATER-518.

damianacatanoso (approved):

damianacatanoso:
> @damianacatanoso I found that multiplying the `P` element values by 10 helps reduce the issue we are observing on OCEANWATER-518.

Can you push so I can test?

PR #58, opened 2020-12-23
Invoke dig_circular using arm_controller + remove all refs to limbs_controllers

Samahu:
## Linked Issues:
* [OCEANWATER-570 | Use arm_controller for performing the dig_circular activity](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-570)
* [OCEANWATER-569 | RRTConnect: Unable to sample any valid states for goal tree ](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-569)

## Summary of Changes
* Invoke dig_circular using arm_controller
* Remove all refs to limbs_controllers

## Test
First start the simulation
```bash
```
Then invoke the dig_circular service (use_defaults: true, parallel: false/true)
```bash
roservice call /arm/dig_circular ...
```
Check for any errors and any deviations in expected behavior.
Check if this resolves issue **OCEANWATER-569**

## TODOs
@damianacatanoso I think the values set to dip down the arm before it does the rotation needs to be revised

### DIFF
--- a/ow_lander/config/fake_controllers.yaml
+++ b/ow_lander/config/fake_controllers.yaml
@@ -11,13 +11,6 @@ controller_list:
       - j_dist_pitch
       - j_hand_yaw
       - j_scoop_yaw
-  - name: fake_limbs_controller
-    joints:
-      - j_shou_yaw
-      - j_shou_pitch
-      - j_prox_pitch
-      - j_dist_pitch
-      - j_hand_yaw
   - name: fake_grinder_controller
     joints:
       - j_shou_yaw
--- a/ow_lander/config/kinematics.yaml
+++ b/ow_lander/config/kinematics.yaml
@@ -6,10 +6,6 @@ antenna:
   kinematics_solver: kdl_kinematics_plugin/KDLKinematicsPlugin
   kinematics_solver_search_resolution: 0.005
   kinematics_solver_timeout: 0.005
-limbs:
-  kinematics_solver: kdl_kinematics_plugin/KDLKinematicsPlugin
-  kinematics_solver_search_resolution: 0.005
-  kinematics_solver_timeout: 0.005
 grinder:
   kinematics_solver: kdl_kinematics_plugin/KDLKinematicsPlugin
   kinematics_solver_search_resolution: 0.005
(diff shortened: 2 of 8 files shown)


kmdalal (approved):
Did several circular digs (using defaults, varying parallel only).  Did not see the RTCONNECT error from before, nor any warnings or errors at all during the operation.  Nice!

damianacatanoso:
I was working on this as well, and I got to successfully remore limbs controller with parallel=true. However, with parallel=false the resulting movement is not correct. The hand bends. Try to run dig_circular this way: 
$ rosservice call /arm/dig_circular -- False 2.0 0 0.01 False -0.155 
and observe the hand motion when it approaches the terrain: that's not supposed to happen. This is the problem I was trying to solve yesterday and today. Let me know if you find a solution. It seems the Inverse Kinematics (IK) can be solved only for the tip_link specified in the lander.srdf and not for other links. As a consequence, no orientation constraints can be applied to a link that is not end effector (l_hand in our case). So I tried to momentarily change the end effector of the arm group using move_group.set_end_effector_link() but no success, I still get an IK error. Some time ago I tried to perform the circular scooping motion using a series of cartesian goal points (waypoints) but it didn't work, the arm would do weird motions between waypoints.

Samahu:
> I was working on this as well, and I got to successfully remove limbs controller with parallel=true. However, with parallel=false the resulting movement is not correct. The hand bends. Try to run dig_circular this way:
> $ rosservice call /arm/dig_circular -- False 2.0 0 0.01 False -0.155
> and observe the hand motion when it approaches the terrain: that's not supposed to happen. This is the problem I was trying to solve yesterday and today. Let me know if you find a solution. It seems the Inverse Kinematics (IK) can be solved only for the tip_link specified in the lander.srdf and not for other links. As a consequence, no orientation constraints can be applied to a link that is not end effector (l_hand in our case). So I tried to momentarily change the end effector of the arm group using move_group.set_end_effector_link() but no success, I still get an IK error. Some time ago I tried to perform the circular scooping motion using a series of cartesian goal points (waypoints) sampled from a half circle (the desired trajectory) but it didn't work, the arm would do weird motions between waypoints.

I see the problem .. unfortunately I wasn't much testing with parallel=False .. and when I do I had the use_defaults=True which internally overrides that flag. I have few things that I want to try.

Samahu:
Closing this PR for now, this is technically an optimization to the current code which can be addressed within next release

PR #59, opened 2020-12-24
Use cartesian path planning to position the arm for dig_circular

Samahu:
## Summary of Changes
This addresses the issue [OCEANWATER-569 | RRTConnect: Unable to sample any valid states for goal tree ](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-569) by using to Cartesian path planning on the _limbs_controller_ when positioning the arm for digging within a dig_circular activity.

## Test
First start the simulation
```bash
```
Then invoke the dig_circular service with the following configurations:
* (use_defaults: true)
* (use_defaults: false, x:1.65 +/- 0.2, y: 0 +/-0.5, z: 0 +/- 0.2, parallel: false/true, ground_position: -0.155)
```bash
roservice call /arm/dig_circular ...
```
Check for any errors and any deviations in expected behavior.
Check if this resolves issue **OCEANWATER-569**

### DIFF
--- a/ow_lander/scripts/activity_full_digging_traj.py
+++ b/ow_lander/scripts/activity_full_digging_traj.py
@@ -32,6 +32,24 @@ def go_to_Z_coordinate(move_group, x_start, y_start, z_start):
   move_group.clear_pose_targets()
 
 
+def go_to_Z_cartesian(move_group, z_start):
+  """
+  :type move_group: class 'moveit_commander.move_group.MoveGroupCommander'
+  :type z_start: float
+  """
+  wpose = move_group.get_current_pose().pose
+  wpose.position.z = z_start
+  waypoints = [copy.deepcopy(wpose)]
+  plan, _ = move_group.compute_cartesian_path(
+      waypoints,
+      0.01,
+      0.0)
+  if len(plan.joint_trajectory.points) == 0:  # If no plan found, abort
+    return False
+  move_group.execute(plan)
+  move_group.stop()
+
+
 def change_joint_value(move_group, joint_index, target_value):
   """
   :type move_group: class 'moveit_commander.move_group.MoveGroupCommander' 
@@ -210,32 +228,25 @@ def dig_circular(move_arm, move_limbs, args, controller_switcher):
     return False
 
   if not parallel:
-
     # Once aligned to trench goal, place hand above trench middle point
     z_start = ground_position + constants.R_PARALLEL_FALSE - depth
     controller_switcher('limbs_controller', 'arm_controller')
-    go_to_Z_coordinate(move_limbs, x_start, y_start, z_start)
+    go_to_Z_cartesian(move_limbs, z_start)
     controller_switcher('arm_controller', 'limbs_controller')
-
     # Rotate hand perpendicular to arm direction
     change_joint_value(move_arm, constants.J_HAND_YAW, -0.29*math.pi)
-
   else:
     # Rotate hand so scoop is in middle point
     change_joint_value(move_arm, constants.J_HAND_YAW, 0.0)
-
     # Rotate scoop
     change_joint_value(move_arm, constants.J_SCOOP_YAW, math.pi/2)
-
     # Rotate dist so scoop is back
     change_joint_value(move_arm, constants.J_DIST_PITCH, -19.0/54.0*math.pi)
-
     # Once aligned to trench goal, place hand above trench middle point
     z_start = ground_position + constants.R_PARALLEL_FALSE - depth
     controller_switcher('limbs_controller', 'arm_controller')
-    go_to_Z_coordinate(move_limbs, x_start, y_start, z_start)
+    go_to_Z_cartesian(move_limbs, z_start)
     controller_switcher('arm_controller', 'limbs_controller')
-
     # Rotate dist to dig
     joint_goal = move_arm.get_current_joint_values()
     dist_now = joint_goal[3]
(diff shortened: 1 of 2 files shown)


mogumbo (approved):
I don't get any errors during testing. However, I don't understand how these changes make previously unreachable targets reachable. Some comments in the code would be a good idea.

damianacatanoso (commented):
Looks good. I run the service 6 times and I got this error twice:
image

Samahu:
> I don't get any errors during testing. However, I don't understand how these changes make previously unreachable targets reachable. Some comments in the code would be a good idea.

I don't have a firm explanation but I think the problem is within the RRTConnect planner itself that is exposed when switching between different controllers in the middle of an activity. Somehow, using Cartesian path planning avoid this issue (by not having to generate these goal states).

Samahu:
> Looks good. The previous error does not appear anymore. But I run the service 6 times and I got this error twice:
> image

Is the error fatal? meaning the entire activity is interrupted

mogumbo:
> I don't have a firm explanation but I think the problem is within the RRTConnect planner itself that is exposed when switching between different controllers in the middle of an activity. Somehow, using Cartesian path planning avoid this issue (by not having to generate these goal states).

So the problem is difficult to understand and this is apparently a workaround. That would be good stuff to learn from the comments. Otherwise a future developer might say "Why is this here? I'll just remove it."

damianacatanoso:
> Is the error fatal? meaning the entire activity is interrupted
I saw that the arm stopped before reaching the goal position, which probably was when this error message got thrown. After it stopped, when the hand started rotating, the hand bounced a little (very small bounce). Let me see if I can take a video.

damianacatanoso:
> > Is the error fatal? meaning the entire activity is interrupted
> 
> I saw that the arm stopped before reaching the goal position, which probably was when this error message got thrown. After it stopped, when the hand started rotating, the hand bounced a little (very small bounce). Let me see if I can take a video.


https://youtu.be/YCA3KYkb6PE

ow_lander/scripts/activity_full_digging_traj.py
```diff
@@ -32,6 +32,24 @@ def go_to_Z_coordinate(move_group, x_start, y_start, z_start):
   move_group.clear_pose_targets()
 
 
+def go_to_Z_cartesian(move_group, z_start):
+  """
+  :type move_group: class 'moveit_commander.move_group.MoveGroupCommander'
+  :type z_start: float
+  """
+  wpose = move_group.get_current_pose().pose
+  wpose.position.z = z_start
```
damianacatanoso, line 41: To compensate the shift we could have a 3D cartesian motion. Now it's 1D because along Z. We could have a motion along X, Y, Z to make sure that the final coordinates of the end effector are x_start, y_start, z_start.

Samahu: I was trying this actually. I think it is sufficient to set only the X, Z since the hand retreats a bit backward a bit on the X direction.

damianacatanoso: > I was trying this actually. I think it is sufficient to set only the X, Z since the hand retreats a bit backward a bit on the X direction.
This is for the specific case of the arm scooping at the center of the workspace. The offset might change when scooping on the sides, gaining a Y component as well.

damianacatanoso: When trying to give the cartesian planner the 3D goal point (x_star, y_star, z_star), the planner doesn't find any trajectory to get there. This is because the cartesian planner uses "pose" (position+orientation) waypoints and I am telling the planner to move the hand to another position while keeping the same orientation throughout the trajectory, which is not possible with our arm configuration. So the solution is to place the hand on the point (x_star, y_star, z_star) using _go_to_Z_coordinate_, before calling _go_to_Z_cartesian_. I have implemented this on a copy of this branch called OCEANWATER-518_Switch_to_using_effort_controllers_JointTrajectoryController_DAMIANA. In this branch I also: 1) multiplied by 10 the P gains for all the joint controllers, 2) set RRTstar as default planner (apparently default was still RRTconnect). I am not creating a ticket/PR because it addresses the same problem that your ticket/PR addresses. If you like, you can merge mine into yours.

Samahu: You can create a separate PR that would merge into this branch (no jira ticket needed).
This what I was trying to deal with .. the issue when adding x, y to the goal of go_to_Z_cartesian is that limbs move_group_commander fails to generate a proper solution. There are cases where it succeeds and on other occasions it fails (for the exact same goal). I am trying to understand the problem better before settling on a particular a solution.

Samahu: Ultimately, we need to find something reliable by tomorrow.

Samahu: I don't seem to see the branch .. did you forget to publish it?

Samahu:
Closing this PR in favor of https://github.com/nasa/ow_simulator/pull/60 which should completely address issue [OCEANWATER-569 | RRTConnect: Unable to sample any valid states for goal](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-569)

ow_lander/scripts/activity_full_digging_traj.py
damianacatanoso: Sorry, the branch is called "bugfix/rrtconnect_unable_to_sample_any_valid_states_for_goal_tree_DAMIANA". Also, I can explain why it fails.

PR #61, opened 2020-12-29
Oceanwater 544 ensure rrt star planner is set during arm operations

damianacatanoso:
Changed default planner for arm and limbs to RRTstar because:
- it's optimal planner
- optimizes for mechanical work so uses less power
- avoids unnecessary arm motions when reaching a target position

Test:
-  europa_terminator_workspace.launch
- roslaunch ow_autonomy autonomy_node.launch plan:="ReferenceMission1.plx"
- Check that all services run correctly

### DIFF
--- a/ow_lander/config/ompl_planning.yaml
+++ b/ow_lander/config/ompl_planning.yaml
@@ -122,7 +122,7 @@ planner_configs:
     dense_delta_fraction: 0.001  # delta fraction for interface detection. default: 0.001
     max_failures: 5000  # maximum consecutive failure limit. default: 5000
 arm:
-  default_planner_config: RRTConnect
+  default_planner_config: RRTstar
   planner_configs:
     - SBL
     - EST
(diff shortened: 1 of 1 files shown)


damianacatanoso:
@mogumbo I didn't have the error you mentioned because I run the plan with the new PID values (https://github.com/nasa/ow_simulator/pull/62). I don't really see why using RRTstar instead of RRTconnect would case that error. Maybe it was just an occurrence that would have happened with RRTconnect as well. I am not sure, but since we don't have time to investigate this for Release 7, I'll just keep RRTconnect as default planner and use RRTstar only for dig_circular instead. I'll push these changes.

damianacatanoso:
> @mogumbo I didn't have the error you mentioned because I run the plan with the new PID values (#62). I don't really see why using RRTstar instead of RRTconnect would case that error. Maybe it was just an occurrence that would have happened with RRTconnect as well. I am not sure, but since we don't have time to investigate this for Release 7, I'll just keep RRTconnect as default planner and use RRTstar only for dig_circular instead. I'll push these changes.

Actually, I'll include these changes (setting RRTstar as planner for dig_circular) in another PR. I am closing this PR.

PR #62, opened 2020-12-29
Multiplied P gains by 10 for shou,dist,prox,hand joints

damianacatanoso:
This should increase the accuracy of the controller. 
Test:

europa_terminator_workspace.launch
roslaunch ow_autonomy autonomy_node.launch plan:="ReferenceMission1.plx"
Check that all services run correctly

### DIFF
--- a/ow_lander/config/ros_controllers.yaml
+++ b/ow_lander/config/ros_controllers.yaml
@@ -90,7 +90,7 @@ arm_controller:
   stop_trajectory_duration: &traj_stop_duration 0.5
   gains:
     j_shou_yaw: 
-      p: &j_shou_yaw_p 8000.0
+      p: &j_shou_yaw_p 9000.0
       d: &j_shou_yaw_d 300.0
       i: &j_shou_yaw_i 10.0
       i_clamp: &ki_clamp_default 2
--- a/ow_lander/config/ros_controllers.yaml
+++ b/ow_lander/config/ros_controllers.yaml
@@ -105,12 +105,12 @@ arm_controller:
       i: &j_prox_pitch_i 2.0
       i_clamp: 1
     j_dist_pitch:
-      p: &j_dist_pitch_p 7000.0
+      p: &j_dist_pitch_p 1400.0
       d: &j_dist_pitch_d 20.0
       i: &j_dist_pitch_i 0.06
       i_clamp: 1
     j_hand_yaw:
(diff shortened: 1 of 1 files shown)


Samahu:
Getting this error:

image

The issue doesn't impede the execution of the activity but there is a slight shake when the scoop rotates.

damianacatanoso:
> The issue doesn't impede the execution of the activity but there is a slight shake when the scoop rotates.

Addressed issue by changing dist, hand values to the double of their original value. I tested dig_circular with use_default=True and with rosservice call /arm/dig_circular "{use_defaults: false, x: 1.65, y: 0.2, depth: 0.01, parallel: false, ground_position: -0.155}"

Samahu:
I did spend very little time testing this but the new values worked better for me. Also as I have expected the higher P did help eliminate or limit the issue with LCP errors.

The error message about `controller taking too long to execute` still occurs but it is a very valid message.
```bash
[ERROR] [...]: Controller is taking too long to execute trajectory (the expected upper bound for the trajectory execution was 21.444671 seconds). Stopping trajectory.
```
This error happens when the controller is unable to fully follow the planned trajectory.
I have seen this happen when the scoop collides with the terrain and is unable to continue.
On the other hand, on the current release branch (release-7), it does not report any error even though the planned trajectory was not realized, instead it reports success along with lots of LCP error messages.

Unfortunately, even though the API tells that it failed to carry out the planned trajectory we report success anyway, and since this is the case (i.e. our services return success regardless if trajectory success status) the autonomy probably doesn't make good use of this. So at this point I am not sure if we want to address this and to what level.

I am OK with the new values but there is a change in the behavior to be expected (and check if it is  reasonable to us).

Also you may want to change the merge into branch towards the release-7 branch .. since OCEANWATER-518 has been merged already.

Samahu:
@damianacatanoso I did change the merge into branch to **release-7**. Just in case we want to proceed with the merge

Samahu (approved):
Per our discussion, the higher P values seem to improve/resolve the two issues:
* Trajectory start position deviates from current
* Reduce or eliminate LCP error messages
Additionally, this doesn't change the expected behavior for the scoop to be stalled when obstructed by the terrain.

kmdalal (approved):
I haven't tested this, but feel that Ussama's feedback suffices.

PR #63, opened 2020-12-29
Another solution to rrtconnect unable to sample any valid states for goal tree

damianacatanoso:
This is a simple solution to the "RRTConnect unable to sample any valid states for goal tree" problem and the offset that was introduced by adopting the cartesian planner.

Test:
- europa_terminator_workspace.launch
- roslaunch ow_autonomy autonomy_node.launch plan:="ReferenceMission1.plx"
- Check that there's no more offset between the trench resulting from the grind service and the scoop circular trajectory.

### DIFF
--- a/ow_lander/config/ompl_planning.yaml
+++ b/ow_lander/config/ompl_planning.yaml
@@ -122,7 +122,7 @@ planner_configs:
     dense_delta_fraction: 0.001  # delta fraction for interface detection. default: 0.001
     max_failures: 5000  # maximum consecutive failure limit. default: 5000
 arm:
-  default_planner_config: RRTConnect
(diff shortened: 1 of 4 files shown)


Samahu (commented):
I think this approach can work too to avoid the problem we are facing "Unable to sample any valid states ...".

The only take away is it adds another planning segment and increases the activity execution time a bit.

I have observed one error but I am not sure if it totally halts the execution of the activity:

image
 
 We definitely need more testing but I can see that this approach helps  remove error.

damianacatanoso:
This solution has not been used but we'll keep it as a backup. Solution chosen: https://github.com/nasa/ow_simulator/pull/60

PR #67, opened 2021-01-20
Oceanwater 544 ensure RRTstar planner is set during arm operations

damianacatanoso:
## Linked Issues:
| Jira Ticket 🎟️   | [Oceanwater-544](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-544) |
| ----------- | ----------- |
| EPIC ⚡| [Oceanwater-87](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-87) |
| Github :octocat:  | # |

## Summary of Changes
* Changed default planner for arm and limbs move_group to RRTstar
* While testing I noticed some of the PID gains could have been improved, so I adjusted the P gain for yaw pitch and hand
* I did the following tests to verify RRTstar reliability (please keep in mind that grind and deliver_sample were already using RRTstar, so no testing is required for them):
- roslaunch ow_autonomy autonomy_node.launch plan:=ReferenceMission1.plx
- Guarded_move:
rosservice call /arm/guarded_move -- False 1.7 0.3 0.5 0 0 1 0.7
rosservice call /arm/guarded_move -- False 1.7 -0.3 0.5 0 0 1 0.7
rosservice call /arm/guarded_move -- False 1.9 -0.4 0.5 0 0 1 0.7
rosservice call /arm/guarded_move -- False 1.9 0.5 0.5 0 0 1 0.7
rosservice call /arm/guarded_move -- False 1.7 0.0 0.5 1 1 1 0.7
rosservice call /arm/guarded_move -- False 1.7 0.0 0.5 0.5 -0.5 1 0.8
- Dig_circular:
rosservice call /arm/dig_circular -- False 2.0 0.3 0.05 False -0.155
rosservice call /arm/dig_circular -- False 2.0 -0.3 0.05 True -0.155
rosservice call /arm/dig_circular -- False 2.0 0.1 0.05 True -0.155
rosservice call /arm/dig_circular -- False 2.0 -0.1 0.05 False -0.155
rosservice call /arm/dig_circular -- False 2.0 -0.1 0.05 True -0.155
- Dig_linear:
rosservice call /arm/dig_linear -- False 1.6 -0.1 0.05 0.3 -0.155
rosservice call /arm/dig_linear -- False 1.55 -0.3 0.05 0.3 -0.155
rosservice call /arm/dig_linear -- False 1.55 -0.6 0.05 0.2 -0.155
rosservice call /arm/dig_linear -- False 1.3 -0.5 0.05 0.3 -0.155

## Test
* You may repeat the tests above, including ReferenceMission1, and/or try different input values. Before you test dig_circular and dig_linear, I suggest you disable collision between scoop and terrain, here is how: go to lander.xacro, line 549, replace 0x0001 with 0x0002. Because otherwise, since we are not grinding before, the scoop will not penetrate the terrain.

### DIFF
--- a/ow_lander/config/ompl_planning.yaml
+++ b/ow_lander/config/ompl_planning.yaml
@@ -122,7 +122,7 @@ planner_configs:
     dense_delta_fraction: 0.001  # delta fraction for interface detection. default: 0.001
     max_failures: 5000  # maximum consecutive failure limit. default: 5000
 arm:
-  default_planner_config: RRTConnect
+  default_planner_config: RRTstar
   planner_configs:
     - SBL
     - EST
@@ -178,7 +178,7 @@ antenna:
   projection_evaluator: joints(j_ant_pan,j_ant_tilt)
   longest_valid_segment_fraction: 0.005
 limbs:
-  default_planner_config: RRTConnect
+  default_planner_config: RRTstar
   planner_configs:
     - SBL
     - EST
--- a/ow_lander/config/ros_controllers.yaml
+++ b/ow_lander/config/ros_controllers.yaml
@@ -90,7 +90,7 @@ arm_controller:
   stop_trajectory_duration: &traj_stop_duration 0.5
   gains:
     j_shou_yaw: 
-      p: &j_shou_yaw_p 9000.0
+      p: &j_shou_yaw_p 8300.0
       d: &j_shou_yaw_d 300.0
       i: &j_shou_yaw_i 10.0
       i_clamp: &ki_clamp_default 2
@@ -105,13 +105,13 @@ arm_controller:
       i: &j_prox_pitch_i 2.0
       i_clamp: 1
     j_dist_pitch:
-      p: &j_dist_pitch_p 1400.0
+      p: &j_dist_pitch_p 2000.0
       d: &j_dist_pitch_d 20.0
       i: &j_dist_pitch_i 0.06
       i_clamp: 1
     j_hand_yaw:
-      p: &j_hand_yaw_p 200.0
-      d: &j_hand_yaw_d 10.0
+      p: &j_hand_yaw_p 1000.0
+      d: &j_hand_yaw_d 60.0
       i: &j_hand_yaw_i 0.01
       i_clamp: 1
     j_scoop_yaw:
--- a/ow_lander/scripts/activity_deliver_sample.py
+++ b/ow_lander/scripts/activity_deliver_sample.py
@@ -9,6 +9,7 @@ from tf.transformations import euler_from_quaternion
 from moveit_msgs.msg import PositionConstraint
 from geometry_msgs.msg import Quaternion
 from shape_msgs.msg import SolidPrimitive
+import rospy
 
 
 def arg_parsing(req):
@@ -33,6 +34,7 @@ def deliver_sample(move_arm, args):
   :type move_arm: class 'moveit_commander.move_group.MoveGroupCommander'
   :type args: List[bool, float, float, float]
   """
+  default_planner_id = rospy.get_param("/move_group/arm/default_planner_config")
   move_arm.set_planner_id("RRTstar")
 
   x_delivery = args[1]
@@ -102,5 +104,7 @@ def deliver_sample(move_arm, args):
   plan = move_arm.go(wait=True)
   move_arm.stop()
   move_arm.clear_pose_targets()
+
+  move_arm.set_planner_id(default_planner_id)
 
   return True
--- a/ow_lander/scripts/activity_deliver_sample.py
+++ b/ow_lander/scripts/activity_deliver_sample.py
@@ -103,6 +103,4 @@ def deliver_sample(move_arm, args):
   move_arm.stop()
   move_arm.clear_pose_targets()
 
-  move_arm.set_planner_id("RRTconnect")
-
   return True


Samahu:
When executing guarded move using the following params
```bash
rosservice call /arm/guarded_move -- False 1.7 0.0 0.5 1 1 1 0.7
```
The arm doesn't descend in a vertical movement

[
image
](https://youtu.be/SvZ-P9gaQpE)

I am not sure if the same behavior happens even without using RRStar but I don't think this problem relates to a specific planner.

damianacatanoso:
> When executing guarded move using the following params
> 
> ```shell
> rosservice call /arm/guarded_move -- False 1.7 0.0 0.5 1 1 1 0.7
> ```
> 
> The arm doesn't descend in a vertical movement
> 
> [ image ](https://youtu.be/SvZ-P9gaQpE)
> 
> I am not sure if the same behavior happens even without using RRStar but I don't think this problem relates to a specific planner.

This is actually the correct behavior. guarded_move has the parameters (direction_x, direction_y, direction_z) which represent the coordinates of the vector defining the descent direction. In this case these values are (1,1,1) so the descent direction is a line tilted 45degrees with respect to all three axes of base_link frame.

Samahu:
> > When executing guarded move using the following params
> > ```shell
> > rosservice call /arm/guarded_move -- False 1.7 0.0 0.5 1 1 1 0.7
> > ```
> > 
> > 
> > The arm doesn't descend in a vertical movement
> > [ image ](https://youtu.be/SvZ-P9gaQpE)
> > I am not sure if the same behavior happens even without using RRStar but I don't think this problem relates to a specific planner.
> 
> This is actually the correct behavior. guarded_move has the parameters (direction_x, direction_y, direction_z) which represent the coordinates of the vector defining the descent direction. In this case these values are (1,1,1) so the descent direction is a line tilted 45degrees with respect to all three axes of base_link frame.

Yes, I forgot about the direction parameter.

Samahu (changes requested):
This looks plausible based on the tests you had.

However, in `activity_deliver_sample.py` there is a line that switches the current planner for the arm back to *RRTConnect*. This would affect subsequent operations .. meaning initially the arm move_group is using *RRTStar* per the `ompl_planning.yaml` but once a **deliver_sample** movement is executed all subsequent activities performed afterwards would be using *RRTConnect*.

Additionally, since we using *RRTStar* by default I recommend removing the command on line 36 of `activity_deliver_sample.py` where we explicitly set the planner to be *RRTStar*.

damianacatanoso:
> This looks plausible based on the tests you had.
> 
> However, in `activity_deliver_sample.py` there is a line that switches the current planner for the arm back to _RRTConnect_. This would affect subsequent operations .. meaning initially the arm move_group is using _RRTStar_ per the `ompl_planning.yaml` but once a **deliver_sample** movement is executed all subsequent activities performed afterwards would be using _RRTConnect_.
> 
> Additionally, since we using _RRTStar_ by default I recommend removing the command on line 36 of `activity_deliver_sample.py` where we explicitly set the planner to be _RRTStar_.

Good catch! I removed the line where the planner was set back to RRTConnect at the end of sample delivery. However, I kept the set_planner_id to RRTStar at the beginning because Anjan explicitly said Sample Delivery need RRTStar and we should keep that line for safety.

Samahu:
> > This looks plausible based on the tests you had.
> > However, in `activity_deliver_sample.py` there is a line that switches the current planner for the arm back to _RRTConnect_. This would affect subsequent operations .. meaning initially the arm move_group is using _RRTStar_ per the `ompl_planning.yaml` but once a **deliver_sample** movement is executed all subsequent activities performed afterwards would be using _RRTConnect_.
> > Additionally, since we using _RRTStar_ by default I recommend removing the command on line 36 of `activity_deliver_sample.py` where we explicitly set the planner to be _RRTStar_.
> 
> Good catch! I removed the line where the planner was set back to RRTConnect at the end of sample delivery. However, I kept the set_planner_id to RRTStar at the beginning because Anjan explicitly said Sample Delivery needs RRTStar and we should keep that line for safety.

Maybe I am lacking a bit of history on the issue but do we know why the *RRTConnect* planner is not suitable in first place for the deliver sample? I mean *RRTStar* and *RRTConnect* are merely variants of the same algorithm *RRT*. Could the issue been resolved without selecting *RRTStar* specifically.

In any case, if you wish to maintain *RRTStar* for the deliver sample without affecting all subsequent operations, then before explicitly switching to the *RRTStar* you need to save the name of current planner to a variable, do the switch, then restore the previous planner before returning the call. Otherwise, say we did change the config back to use *RRTConnect*, then what is going to happen is all calls will be using *RRTConnect* initially, but once a **deliver_sample** operation is invoked then all other arm movements would suddenly be using *RRTStar*, thus the behavior won't be consistent.

anjan1984:
The constraint for delivery sample is that the scoop needs to be upright (pointing upwards) while transferring the collected sample from the ground to the delivery point. This can be achieved by imposing a constraint on the motion plan. At the end of the dig trench function the scoop is pointing upwards. So it can be safely assumed that an optimal solution does not need to change the orientation of the scoop to find the solution. Non-optimal versions of the plan have shown lots of different movements that does not keep the scoop pointing upwards all the time. RRT* is a solution that solves this problem. However it could be solved by other methods.

damianacatanoso:
> Maybe I am lacking a bit of history on the issue but do we know why the _RRTConnect_ planner is not suitable in first place for the deliver sample? I mean _RRTStar_ and _RRTConnect_ are merely variants of the same algorithm _RRT_. Could the issue been resolved without selecting _RRTStar_ specifically.
> 
> In any case, if you wish to maintain _RRTStar_ for the deliver sample without affecting all subsequent operations, then before explicitly switching to the _RRTStar_ you need to save the name of current planner to a variable, do the switch, then restore the previous planner before returning the call. Otherwise, say we did change the config back to use _RRTConnect_, then what is going to happen is all calls will be using _RRTConnect_ initially, but once a **deliver_sample** operation is invoked then all other arm movements would suddenly be using _RRTStar_, thus the behavior won't be consistent.

That's true, I'll do that. I think (Anjan wrote the sample_delivery code, not me) the reason why it's so important that sample delivery is done with RRTStar is that RRTConnect provides different solutions for the arm movement among different service calls. With RRTConnect, since it's not optimal, the arm would still get to the same goal state but doing weird intermediate movements. While RRTstar always provides the same (sub-)optimal solution. That's the same reason why we use RRTstar for grind.

damianacatanoso:
> The constraint for delivery sample is that the scoop needs to be upright (pointing upwards) while transferring the collected sample from the ground to the delivery point. This can be achieved by imposing a constraint on the motion plan. At the end of the dig trench function the scoop is pointing upwards. So it can be safely assumed that an optimal solution does not need to change the orientation of the scoop to find the solution. Non-optimal versions of the plan have shown lots of different movements that does not keep the scoop pointing upwards all the time. RRT* is a solution that solves this problem. However it could be solved by other methods.

So RRTConnect violates the orientation constraint on the scoop?

anjan1984:
I did not spend enough time on making RRT connect work with the constraint. I found RRT* to be an easier solution.

Samahu:
> The constraint for delivery sample is that the scoop needs to be upright (pointing upwards) while transferring the collected sample from the ground to the delivery point. This can be achieved by imposing a constraint on the motion plan. At the end of the dig trench function the scoop is pointing upwards. So it can be safely assumed that an optimal solution does not need to change the orientation of the scoop to find the solution. Non-optimal versions of the plan have shown lots of different movements that does not keep the scoop pointing upwards all the time. RRT* is a solution that solves this problem. However it could be solved by other methods.

Makes sense. Thanks for clarifying this.

damianacatanoso:
> Maybe I am lacking a bit of history on the issue but do we know why the _RRTConnect_ planner is not suitable in first place for the deliver sample? I mean _RRTStar_ and _RRTConnect_ are merely variants of the same algorithm _RRT_. Could the issue been resolved without selecting _RRTStar_ specifically.
> 
> In any case, if you wish to maintain _RRTStar_ for the deliver sample without affecting all subsequent operations, then before explicitly switching to the _RRTStar_ you need to save the name of current planner to a variable, do the switch, then restore the previous planner before returning the call. Otherwise, say we did change the config back to use _RRTConnect_, then what is going to happen is all calls will be using _RRTConnect_ initially, but once a **deliver_sample** operation is invoked then all other arm movements would suddenly be using _RRTStar_, thus the behavior won't be consistent.

Addressed!

Samahu (approved):
Even though the MoveitGroupCommander interface has a get_planner_id but it returns an empty string first time it is used. So I see why you didn't use though. Very good approach!

damianacatanoso:
> get_planner_id

Thanks :)

PR #74, opened 2021-01-28
Use default version of OpenCV on current platform + Use catkin default c++ standard

Samahu:
## Linked Issues:
| EPIC ⚡ | [OCEANWATER-577 / Support for ROS Noetic](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-577) |
| Jira Ticket 🎟️   | [OCEANWATER-578 / Remove strict requirement on OpenCV 3.0 in ow_dynamic_terrain ](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-578) |
| Jira Ticket 🎟️   | [OCEANWATER-593 / Don't explicitly set a required C++ standard unless absolutely needed ](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-593) |

## Summary of Changes
This is a single iteration at our effort to support **ROS Noetic**, these following changes are necessary and does't break **ROS Melodic** compilation: 
* Use default version of OpenCV on current platform 
* Use catkin default c++ standard

## Test
* Make sure the system still successfully compiles against **ROS Melodic**
* If you happen to have an Ubuntu 20.04 system with all the prerequisites for OceanWATERS then you may verify that ow_dynamic_terrain doesn't fail due to a requiring OpenCV 3.0

### DIFF
--- a/ow_dynamic_terrain/CMakeLists.txt
+++ b/ow_dynamic_terrain/CMakeLists.txt
@@ -17,7 +17,7 @@ find_package(catkin REQUIRED COMPONENTS
   cv_bridge
 )
 
-find_package(OpenCV 3 REQUIRED
+find_package(OpenCV REQUIRED
   core
 )
 
--- a/ow_dynamic_terrain/CMakeLists.txt
+++ b/ow_dynamic_terrain/CMakeLists.txt
@@ -1,10 +1,6 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(ow_dynamic_terrain)
 
-## Compile as C++11, supported in ROS Kinetic and newer
-# add_compile_options(-std=c++11)
-add_compile_options(-std=c++11)
-
 ## Find catkin macros and libraries
 ## if COMPONENTS list like find_package(catkin REQUIRED COMPONENTS xyz)
 ## is used, also find other catkin packages
--- a/ow_ephemeris/CMakeLists.txt
+++ b/ow_ephemeris/CMakeLists.txt
@@ -1,9 +1,6 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(ow_ephemeris)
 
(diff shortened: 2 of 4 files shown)


ow_dynamic_terrain/CMakeLists.txt
```diff
@@ -1,10 +1,6 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(ow_dynamic_terrain)
 
-## Compile as C++11, supported in ROS Kinetic and newer
-# add_compile_options(-std=c++11)
-add_compile_options(-std=c++11)
```
kmdalal, line 6: Just curious: is the removal of this add_compile_options and the others a requirement for Noetic, or an unrelated cleanup?

kmdalal (approved):
I'm unable to test this in Ubuntu 20 anytime soon, but the regression passed.

ow_dynamic_terrain/CMakeLists.txt
Samahu: I am not sure why this was added in first place (probably the default template by catkin adds that) it might have been required at some point. Both **melodic** and **noetic** are currently embracing c++14 standard so having this module requesting c++11 resulted in a set of compile time errors for **noetic** (Ubuntu 20.04 has an upgraded version of the c/c++ compiler and libraries).

mogumbo (approved):
This compiles and runs on Ubuntu 18.04 with melodic. ow_dynamic_terrain doesn't compile on Ubuntu 20.04 with noetic but not because of a missing OpenCV 3.0.

Samahu:
> This compiles and runs on Ubuntu 18.04 with melodic. ow_dynamic_terrain doesn't compile on Ubuntu 20.04 with noetic but not because of a missing OpenCV 3.0.

Yes, there are more problems down the line.

PR #79, opened 2021-02-04
Ow test stop gazebo

lanseafood:
## Linked Issues:
| ----------- | ----------- |
| Github :octocat:  | # |

## Summary of Changes
* 
if arm fault, stops guarded arm from moving
if no arm fault, guarded move can be called
however, does not stop arm mid guarded move, that may require another PR or more drastic changes (like breaking up the waypoints into separate messages instead of 1 condensed message. 

## Test
*


lanseafood:
reopened under correct name, removing now

PR #81, opened 2021-02-04
Oceanwater 602 Faults Stop Arm for  Unstow and Stow

lanseafood:
## Linked Issues:
| Jira Ticket 🎟️   | [Oceanwater-602](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-602) |
| ----------- | ----------- |
| Github :octocat:  | na |

## Summary of Changes
* if arm fault, stops stow and unstow movement from executing
* if no arm fault, the stow and unstow can be called
** the stopping of arms through faults can be called prior to action and also during action

## Test
* call any roslaunch 
* go into rqt and select any arm fault
* open a new terminal and try calling the rosservice for stow and unstow
* should receive the following and no movement should be seen in gazebo or rviz 

```
success: False
message: "Unstow action failed"
```
* uncheck the arm fault in dynamic reconfigure
*  call rosservice for stow/unstow and see execution as nominal

### DIFF
--- a/ow_lander/scripts/path_planning_commander.py
+++ b/ow_lander/scripts/path_planning_commander.py
@@ -231,9 +231,6 @@ class PathPlanningCommander(object):
 
     rospy.spin()
 
-  def stop_arm(self):
-    plan = activity_guarded_move.stop_guarded_move_plan(self.arm_move_group)
-    self.trajectory_async_executer.systemFaultListener(plan.joint_trajectory, self.handle_guarded_move_feedback)
 
   def callback(self, data):
     """
--- a/ow_lander/scripts/path_planning_commander.py
+++ b/ow_lander/scripts/path_planning_commander.py
@@ -234,6 +234,9 @@ class PathPlanningCommander(object):
     self.trajectory_async_executer.systemFaultListener(plan.joint_trajectory, self.handle_guarded_move_feedback)
 
   def callback(self, data):
+    """
+    If system fault occurs, and it is an arm failure, path planning stops
+    """
     if data.value == 4:
       self.arm_fault = True
       self.trajectory_async_executer.stop()
--- a/ow_lander/scripts/path_planning_commander.py
+++ b/ow_lander/scripts/path_planning_commander.py
@@ -168,8 +168,6 @@ class PathPlanningCommander(object):
     """
     :type feedback: FollowJointTrajectoryFeedback
     """
-    if self.arm_fault:
-      self.trajectory_async_executer.stop()
 
     if self.ground_detector.detect():
       self.trajectory_async_executer.stop()
--- a/ow_lander/scripts/path_planning_commander.py
+++ b/ow_lander/scripts/path_planning_commander.py
@@ -188,7 +188,7 @@ class PathPlanningCommander(object):
       self.ground_detector.reset()
       self.trajectory_async_executer.execute(plan.joint_trajectory,
                                             done_cb=self.handle_guarded_move_done,
-                                            active_cb=None,
+                                            active_cb=self.stop_arm_if_fault,
                                             feedback_cb=self.handle_guarded_move_feedback)
       # To preserve the previous behaviour we are adding a blocking call till the
       # execution of the trajectory is completed
@@ -231,6 +231,11 @@ class PathPlanningCommander(object):
 
     rospy.spin()
 
+  def stop_arm_if_fault(self):
+    if self.arm_fault:
+      plan = activity_guarded_move.stop_guarded_move_plan(self.arm_move_group)
+      self.trajectory_async_executer.systemFaultListener(plan.joint_trajectory, self.handle_guarded_move_feedback)
+      self.trajectory_async_executer.stop()
 
   def callback(self, data):
     """
--- a/ow_lander/scripts/path_planning_commander.py
+++ b/ow_lander/scripts/path_planning_commander.py
@@ -58,7 +58,6 @@ class PathPlanningCommander(object):
     if len(plan.joint_trajectory.points) == 0:
       return False
     self.trajectory_async_executer.execute(plan.joint_trajectory, 
-                                          active_cb=None,
                                           feedback_cb=self.stop_arm_if_fault_feedback_cb)
     self.trajectory_async_executer.wait()
     return self.return_message("Stow arm")
@@ -76,7 +75,6 @@ class PathPlanningCommander(object):
     if len(plan.joint_trajectory.points) == 0:
       return False
     self.trajectory_async_executer.execute(plan.joint_trajectory,
-                                          active_cb=None,
                                           feedback_cb=self.stop_arm_if_fault_feedback_cb)
     self.trajectory_async_executer.wait()
     return self.return_message("Unstow arm")
@@ -191,7 +189,6 @@ class PathPlanningCommander(object):
     self.ground_detector.reset()
     self.trajectory_async_executer.execute(plan.joint_trajectory,
                                           done_cb=self.handle_guarded_move_done,
-                                          active_cb=None,
                                           feedback_cb=self.handle_guarded_move_feedback)
     # To preserve the previous behaviour we are adding a blocking call till the
     # execution of the trajectory is completed
--- a/ow_lander/scripts/path_planning_commander.py
+++ b/ow_lander/scripts/path_planning_commander.py
@@ -57,10 +57,12 @@ class PathPlanningCommander(object):
     plan = self.arm_move_group.plan()
     if len(plan.joint_trajectory.points) == 0:
       return False
-    self.trajectory_async_executer.execute(plan.joint_trajectory)
+    self.trajectory_async_executer.execute(plan.joint_trajectory, 
+                                          active_cb=self.stop_arm_if_fault_active_cb,
+                                          feedback_cb=self.stop_arm_if_fault_feedback_cb)
     self.trajectory_async_executer.wait()
-    print("Stow arm activity completed")
-    return True, "Done"
+    return self.return_message("Stow arm")
+
 
   # === SERVICE ACTIVITIES - Unstow =============================
   def handle_unstow(self, req):
@@ -73,10 +75,11 @@ class PathPlanningCommander(object):
     plan = self.arm_move_group.plan()
     if len(plan.joint_trajectory.points) == 0:
       return False
-    self.trajectory_async_executer.execute(plan.joint_trajectory)
+    self.trajectory_async_executer.execute(plan.joint_trajectory,
+                                          active_cb=self.stop_arm_if_fault_active_cb,
+                                          feedback_cb=self.stop_arm_if_fault_feedback_cb)
     self.trajectory_async_executer.wait()
-    print("Unstow arm activity completed")
-    return True, "Done"
+    return self.return_message("Unstow arm")
 
   # === SERVICE ACTIVITIES - deliver sample =============================
   def handle_deliver_sample(self, req):
@@ -165,6 +168,9 @@ class PathPlanningCommander(object):
     """
     :type feedback: FollowJointTrajectoryFeedback
     """
+    if self.arm_fault:
+      self.trajectory_async_executer.stop()
+
     if self.ground_detector.detect():
       self.trajectory_async_executer.stop()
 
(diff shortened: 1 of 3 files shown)


ow_lander/scripts/path_planning_commander.py
```diff
@@ -194,8 +205,29 @@ def run(self):
     self.guarded_move_srv = rospy.Service(
         'arm/guarded_move', GuardedMove, self.handle_guarded_move)
     print("path_planning_commander has started!")
+
     rospy.spin()
 
+  def stop_arm(self):
+    plan = activity_guarded_move.stop_guarded_move_plan(self.arm_move_group)
+    self.trajectory_async_executer.systemFaultListener(plan.joint_trajectory, self.handle_guarded_move_feedback)
```
Samahu, line 213: This function is never referenced and does seems valid, i think it needs to be dropped
```suggestion
```

kmdalal:
I'm testing using autonomy instead of rqt, assuming sim behavior should be the same.  Assumption: _any_ arm fault or combination will stop the arm.  Some observations:
- GuardedMove does stop when faults are injected, but not always right away.  In particular tool selection (rotation of carousel) does not stop.
- the "guarded_move_plan failed" message was not seen when I injected the fault after GuardedMove had started.
- all other arm operations continue in the presence of faults.  Shouldn't they all be disabled, i.e. the arm stops no matter what it's doing?

Samahu:
> * the "guarded_move_plan failed" message was not seen when I injected the fault after GuardedMove had started.
This message would be printed only if you used using the service caller from the command line
> * all other arm operations continue in the presence of faults.  Shouldn't they all be disabled, with the arm stopping no matter what it's doing?
I haven't gone through the code fully yet but I believe only guarded move is subject to this faulting method

ow_lander/scripts/path_planning_commander.py
```diff
@@ -194,8 +205,29 @@ def run(self):
...
+      self.trajectory_async_executer.stop()
+    else:
+      self.arm_fault = False
+
+  def systemFaultListener(self):
+    """
+    Listens for system faults
+    """
+    rospy.Subscriber("/system_faults_status", SystemFaults, self.callback)
+
```
Samahu, line 230: To keep your code consistent with the rest of the file I recommend you move the subscription definition into the _run()_ method along with the rest of the subscribers. This would make it easier to review what services, publishers and subscribers that this class is making use of.

Samahu: To elaborate more, there is really no need to have this method that merely invokes a single line.

ow_lander/scripts/path_planning_commander.py
```diff
@@ -194,8 +205,29 @@ def run(self):
...
     rospy.spin()
 
+  def stop_arm(self):
+    plan = activity_guarded_move.stop_guarded_move_plan(self.arm_move_group)
+    self.trajectory_async_executer.systemFaultListener(plan.joint_trajectory, self.handle_guarded_move_feedback)
+
+  def callback(self, data):
+    """
+    If system fault occurs, and it is an arm failure, path planning stops
```
Samahu, line 217: Evaluate if this following state better describe the event
```suggestion
    If system fault occurs, and it is an arm failure, arm movement is halted abruptly!
```

ow_lander/scripts/path_planning_commander.py
```diff
@@ -194,8 +205,29 @@ def run(self):
...
+    """
+    if data.value == 4:
+      self.arm_fault = True
+      self.trajectory_async_executer.stop()
+    else:
+      self.arm_fault = False
```
Samahu, line 223: I think the logic around when and for how long a fault is introduced needs some discussion, I'll post some notes in my review summary.

Samahu (changes requested):
I think the logic around when and for how long a fault is introduced needs some discussion:
* You are relying on a flag to prevent the arm from executing the movement. If you want to go this route then you can extend the same logic to all the other arm activities
  - The catch is you can only apply this before the arm starts performing the activity or at a certain segment of the activity.
* You are then using `self.trajectory_async_exector.stop` to stop the trajectory execution based on the current state of `data.value==4`. I don't know much about when and for how long this flag stays set but I assume it is always 'on' or activated if any of the arm joints have their faults flag set. If this is the case, then you are constantly and repeatedly  invoking `self.trajectory_async_exector.stop`. Although this may work, I think it needs to be invoked only once and maybe at certain or random times.

ow_lander/scripts/path_planning_commander.py
lanseafood: fixed!

lanseafood:
@Samahu @mikedalal 

I've made some of the smaller changes already. To respond to some of the comments:

1) calling the stop() function actually doesn't stop the arm at all, whether it's called before or during the arm call. Once an execute() is called, the stop() function cannot stop it halfway or partially through the plan.

2) The flag was the only way I could stop the arm trajectory without actually changing the plans themselves. Because of this, the flag (which stays on as long as the rqt fault is selected) can only edit the arm trajectory prior to the actual execution() call. If the execution() is already in progress, I see no direct way to change the plan or stop the trajectory without being more involved (as I mentioned in my description). 

Some possible solutions

1) instead of a full plan, the plan is sent over as separate messages. This would allow us to interject between specific movements.

2) somehow update/change the current plan. However, I don't see a straightforward way of doing this immediately.

Any further thoughts/comments welcome!

ow_lander/scripts/path_planning_commander.py
```diff
@@ -194,9 +206,20 @@ def run(self):
...
     print("path_planning_commander has started!")
+
     rospy.spin()
 
 
+  def callback(self, data):
+    """
+    If system fault occurs, and it is an arm failure, arm movement is halted abruptly!
+    """
+    if data.value == 4:
```
kmdalal, line 217: Magic number 4.  Use an abstraction such as an enum value or macro (#define)

kmdalal:
Stopping the arm after the current plan segment completes seems reasonable for a first cut.  A more sophisticated approach would be good for a sub-group discussion, including Anjan if plan modification is considered.

@Samahu I did see the "guarded_move_plan failed" message via autonomy, but only when I injected the fault before this service started.  

If the scope of this ticket is only guarded_move, that should be in the name.  Or if it makes sense, leave this ticket open to include all the other arm activities, or make it an Epic.

kmdalal:
I don't know the path-planning or Gazebo level implementations or have time to get into them, but I'm wondering if and why a "stop arm on faults" feature needs to be operation-specific at all.  Intuitively I would think that cancelling _all_ arm plans sent to Gazebo for execution would do the trick.

Furthermore, sending back the fault status through the ROS actions's feedback or result messages (Anjan's feature in progress) strikes me as an excellent way for autonomy to detect the fault.

lanseafood:
> I don't know the path-planning or Gazebo level implementations or have time to get into them, but I'm wondering if and why a "stop arm on faults" feature needs to be operation-specific at all. Intuitively I would think that cancelling _all_ arm plans sent to Gazebo for execution would do the trick.
> 
> Furthermore, sending back the fault status through the ROS actions's feedback or result messages (Anjan's feature in progress) strikes me as an excellent way for autonomy to detect the fault.

Hey Mike, all great points here. Yes I agree that the original scope of the ticket was to stop all arm movement, which is why I initially proposed just editing the messages sent on specific topics to gazebo. Ussama at one point introduced the trajectory async class but it was only being used for the guarded move to my knowledge. 

I believe cancelling a plan before it's executed is possible, however, cancelling a plan mid-way would require us to dig further into gazebo and/or alter how we send plans in the first place (sending multiple waypoints vs sending 1 waypoint at a time).

ow_lander/scripts/path_planning_commander.py
lanseafood: this gets updated in the #82 i believe.

kmdalal (approved):
The feature as described in original description plus followon comments worked for me, and I had no major code issues.

Samahu:
@kmdalal @lanseafood I think you might be mixing between an operation of two different interfaces. Both **MoveGroupCommander** and **TrajectoryAsyncExecutor** has a **Stop** method. Although name is same but these act totally different.

* The purpose of **MoveGroupCommander::Stop** is to block and wait until the arm settles down. 
* The purpose of **TrajectoryAsyncExecutor::Stop** is to actually halt the execution of the trajectory that was sent through it.

Samahu:
> @kmdalal @lanseafood I think you might be mixing between an operation of two different interfaces. Both **MoveGroupCommander** and **TrajectoryAsyncExecutor** has a **Stop** method. Although name is same but these act totally different.
> 
> * The purpose of **MoveGroupCommander::Stop** is to block and wait until the arm settles down.
> * The purpose of **TrajectoryAsyncExecutor::Stop** is to actually halt the execution of the trajectory that was sent through it.

This being the case, any arm operation or command executed through the **MoveGroupCommander** interface has no means to actually halt the execution. The only way we can halt arm activities that use the **MoveGroupCommander** (which most of our activities do) is by intercepting every call to **go** or **execute** of that interface and basically ignoring or skipping them

lanseafood:
> * The purpose of **TrajectoryAsyncExecutor::Stop** is to actually halt the execution of the trajectory that was sent through it.

I believe I have only been calling stop() on the trajectory_asyc_exector object in path planning. And from testing I believe that calling the stop() function mid-way through the guarded move does not stop the arm fully.

Samahu:
> I don't know the path-planning or Gazebo level implementations or have time to get into them, but I'm wondering if and why a "stop arm on faults" feature needs to be operation-specific at all. Intuitively I would think that cancelling _all_ arm plans sent to Gazebo for execution would do the trick.
> 
> Furthermore, sending back the fault status through the ROS actions's feedback or result messages (Anjan's feature in progress) strikes me as an excellent way for autonomy to detect the fault.

BTW, some of the services do report success or failure, but I don't think current autonomy plans check the return values of these services.

Samahu:
> > * The purpose of **TrajectoryAsyncExecutor::Stop** is to actually halt the execution of the trajectory that was sent through it.
> 
> I believe I have only been calling stop() on the trajectory_asyc_exector object in path planning. And from testing I believe that calling the stop() function mid-way through the guarded move does not stop the arm fully.

It would only stop the arm after it has been unfolded and starting to descend towards the grounds .. since this is the only part that uses the trajectory_asyc_exector.

kmdalal:
The discussion about stop() is really beyond my understanding of the simulator, and I don't have the time to get into it.  If it makes sense, add Anjan and/or Terry as a reviewer of this feature, and in any case we can discuss it on Dev Day if merging it is not blocking you, Lanssie.

ow_lander/scripts/path_planning_commander.py
```diff
@@ -215,8 +228,40 @@ def run(self):
...
+      return False,  action_name + " failed"
+
+  def callback(self, data):
+    """
+    If system fault occurs, and it is an arm failure, an arm failure flag is set for the whole class
+    """
+    if data.value == 4:
+      self.arm_fault = True
+    else:
+      self.arm_fault = False
```
Samahu, line 264: More compact
```suggestion
    self.arm_fault = (data.value & 4 == 4)
```

ow_lander/scripts/path_planning_commander.py
```diff
@@ -215,8 +228,40 @@ def run(self):
...
+  def return_message(self, action_name, success=None):
+    if not self.arm_fault and success:
+      print(action_name + " activity completed")
+      return success, "Done" 
+    elif not self.arm_fault and success is None:
+      print(action_name + " activity completed")
+      return True, "Done" 
+    else:
+      print(action_name + " activity incomplete")
+      return False,  action_name + " failed"
```
Samahu, line 255: An activity can be completed and still fail even when no fault is enabled...
```suggestion
  def return_message(self, action_name, success=True):
    if not self.arm_fault:
      print(action_name + " activity completed")
      return success, "Done"
    else:
      print(action_name + " activity incomplete")
      return False,  action_name + " failed"
```

Samahu (commented):
The use of two callbacks to issue the `stop` seems redundant to me: 
```python
active_cb=self.stop_arm_if_fault_active_cb,
feedback_cb=self.stop_arm_if_fault_feedback_cb
```

I prefer using the feedback since it gives us more flexibility to simulate a fault at a certain time of execution rather than immediately after the trajectory is submitted.

lanseafood:
> The use of two callbacks to issue the `stop` seems redundant to me:
> 
> ```python
> active_cb=self.stop_arm_if_fault_active_cb,
> feedback_cb=self.stop_arm_if_fault_feedback_cb
> ```
> 
> I prefer using the feedback callback since it gives us more flexibility to inject a fault at a certain time of execution rather than immediately after the trajectory is submitted.

I agree it's a little clunky. active cb and feedback cb require different inputs, otherwise I'd have just let that be a single function. I think technically we could achieve the same functionality with just the feedback_cb alone.

In the future, I imagine that this feedback function is the default cb for actions, with only specific actions that may require their own feedback cbs (like guarded arm move).

Samahu:
> > The use of two callbacks to issue the `stop` seems redundant to me:
> > ```python
> > active_cb=self.stop_arm_if_fault_active_cb,
> > feedback_cb=self.stop_arm_if_fault_feedback_cb
> > ```
> > 
> > 
> > I prefer using the feedback callback since it gives us more flexibility to inject a fault at a certain time of execution rather than immediately after the trajectory is submitted.
> 
> I agree it's a little clunky. active cb and feedback cb require different inputs, otherwise I'd have just let that be a single function. I think technically we could achieve the same functionality with just the feedback_cb alone.
> 
> In the future, I imagine that this feedback function is the default cb for actions, with only specific actions that may require their own feedback cbs (like guarded arm move).

Exactly, I don't think we need to utilize the _active_cb_ for the faults.

ow_lander/scripts/path_planning_commander.py
```diff
@@ -55,10 +57,12 @@ def handle_stow(self, req):
     plan = self.arm_move_group.plan()
     if len(plan.joint_trajectory.points) == 0:
       return False
-    self.trajectory_async_executer.execute(plan.joint_trajectory)
+    self.trajectory_async_executer.execute(plan.joint_trajectory, 
+                                          active_cb=None,
```
Samahu, line 61: I think you can safely drop the active_cb
```suggestion
```

ow_lander/scripts/path_planning_commander.py
```diff
@@ -71,10 +75,11 @@ def handle_unstow(self, req):
     plan = self.arm_move_group.plan()
     if len(plan.joint_trajectory.points) == 0:
       return False
-    self.trajectory_async_executer.execute(plan.joint_trajectory)
+    self.trajectory_async_executer.execute(plan.joint_trajectory,
+                                          active_cb=None,
```
Samahu, line 79: Same
```suggestion
```

ow_lander/scripts/path_planning_commander.py
```diff
@@ -183,17 +192,21 @@ def handle_guarded_move(self, req):
       return False, "guarded_move_plan failed"
     self.ground_detector.reset()
     self.trajectory_async_executer.execute(plan.joint_trajectory,
-                                           done_cb=self.handle_guarded_move_done,
-                                           active_cb=None,
-                                           feedback_cb=self.handle_guarded_move_feedback)
+                                          done_cb=self.handle_guarded_move_done,
+                                          active_cb=None,
```
Samahu, line 196: Same
```suggestion
```

ow_lander/scripts/path_planning_commander.py
lanseafood: done locally

Samahu (approved):
Good work!

Samahu:
@lanseafood You may want to update the PR Description as it still list _guarded move_ as the target of the PR

PR #66, opened 2021-01-14
Oceanwater 534 implement current arm services as rosaction

anjan1984:
## Linked Issues:
| Jira Ticket 🎟️   | [Oceanwater-531]https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-531 |
| ----------- | ----------- |
| EPIC ⚡| [(OCEANWATER-518 | Switch to using effort_controllers/JointTrajectoryControllerl) |
| Github :octocat:  | # |

## Summary of Changes
* 
Implement available ros services as ros actions. All the services that have been converted to ros actions

1. stow 
2. unstow
3. Move guarded
3. grind
4. dig circular
5. dig linear
3. deliver sample 

## Test 
terminal 1:
roslaunch  ow europa_terminator_workspace.launch 

terminal 2:

- python unstow_action_client.py
- python stow_action_client.py
- python guarded_move_action_client.py
- python grind_action_client.py
- python dig_linear_action_client.py
- python dig_circular_action_client.py
- python deliver_action_client.py

look in client code for commented code to try default or custom values for each of the actions.

PR #84, opened 2021-02-09
Implement Stow/Unstow using trajectory async executer + Add an arm stop command

Samahu:
Add an Arm Stop command

## Linked Issues:
| EPIC ⚡| [OCEANWATER-518 / Switch to using effort_controllers/JointTrajectoryController](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-518) |
| :---------- | :---------- |
| Jira Ticket 🎟️   | [OCEANWATER-604 / Implement Stow/Unstow using TrajectoryAsyncExecuter](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-604) |
| Github :octocat:  |  related to #81  |

## Summary of Changes
* Implement **Stow** / **Unstow** arm services using TrajectoryAsyncExecuter
* Add an arm stop command

## Test
Launch the simulation
```bash
```
Execute either **Stow** or **Unstow** through RQT Service Call or using the commands 
```bash
rosservice call /arm/unstow 
rosservice call /arm/stow 
```
To stop the arm at any point (applies only to **Stow**, **Unstow** activities and the last segment of **MoveGuarded**), then open another terminal and invoke the following service:
```bash
rosservice call /arm/stop
```

### DIFF
--- a/ow_lander/CMakeLists.txt
+++ b/ow_lander/CMakeLists.txt
@@ -21,6 +21,7 @@ add_service_files(
   PublishTrajectory.srv
   Stow.srv
   Unstow.srv
+  Stop.srv
 )
 
 add_message_files(
--- a/ow_lander/scripts/path_planning_commander.py
+++ b/ow_lander/scripts/path_planning_commander.py
@@ -34,14 +34,29 @@ class PathPlanningCommander(object):
     self.trajectory_async_executer = TrajectoryAsyncExecuter()
 
   # === SERVICE ACTIVITIES - Stow =============================
+  def handle_stop(self, req):
+    """
+    :type req: class 'ow_lander.srv._Stow.StopRequest'
+    """
+    print("Stop arm activity started")
+    self.trajectory_async_executer.stop()
+    print("Stop arm activity completed")
+    return True, "Done"
+
+  # === SERVICE ACTIVITIES - Stow =============================
   def handle_stow(self, req):
     """
     :type req: class 'ow_lander.srv._Stow.StowRequest'
     """
     print("Stow arm activity started")
     goal = self.arm_move_group.get_named_target_values("arm_stowed")
-    self.arm_move_group.go(goal, wait=True)
-    self.arm_move_group.stop()
+    self.arm_move_group.set_joint_value_target(goal)
+
+    plan = self.arm_move_group.plan()
+    if len(plan.joint_trajectory.points) == 0:
+      return False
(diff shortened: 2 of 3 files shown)


Samahu:
Note that this is a considerably small change and doesn't need that many reviewers but I did include all members that were involved in the discussion today about the arm faults injection.

The TrajectoryAsyncExecuter::Stop can also be extended to the rest of the services as part of Anjan's #66

lanseafood (approved):
Looks great, all commands are working for me!

mogumbo (approved):
Approved, but the output from a stopped activity is misleading:
```
Stop arm activity started
Stop arm activity completed
Unstow arm activity completed
```
Saying something like "Unstow arm activity aborted" would make more sense. This is probably work for a separate ticket.

Samahu:
Another note to add is the call to **TrajectoryAsyncExecuter::Wait** is not actually needed but it was deliberately added to preserve the service behavior for **ow_autonomy**.

Samahu:
> Approved, but the output from a stopped activity is misleading:
> 
> ```
> Stop arm activity started
> Stop arm activity completed
> Unstow arm activity completed
> ```
> 
> Saying something like "Unstow arm activity aborted" would make more sense. This is probably work for a separate ticket.

I see your point

anjan1984 (approved):
Works fine for me

kmdalal:
> Another note to add is the call to **TrajectoryAsyncExecuter::Wait** is not actually needed but it was deliberately added to preserve the service behavior for **ow_autonomy**.

I would like to try without the wait sometime, as much has changed in ow_autonomy since this was added, I think.

I've been unable to test this because Gazebo and Rviz are crashing on me today, but the changes look fine and apparently work!

Samahu:
> > Another note to add is the call to **TrajectoryAsyncExecuter::Wait** is not actually needed but it was deliberately added to preserve the service behavior for **ow_autonomy**.
> 
> I would like to try without the wait sometime, as much has changed in ow_autonomy since this was added, I think.

Removing the wait would cause these service invocation to return immediately. Which is fine if you are using the command line to invoke these services. However, this wouldn't play nicely for **ow_autonomy** since current PLEXIL plans expect the service invocation to only return when the activity is complete. **ow_autonomy**  would need to subscribe to the **follow_joint_trajectory** topics/callbacks to know when an activity is actually done. In any case, this will be covered by the #66 PR.

PR #97, opened 2021-03-04
OCEANWATER-612: Integrate Functionality of GSAP into OW

jfugate1:
## Linked Issues:
| Jira Ticket 🎟️   | [Oceanwater-612](url) |
| ----------- | ----------- |
| Github :octocat:  | # |

## Summary of Changes
* Implemented GSAP functionality such that Monte Carlo prognostics model is used to compute battery state information rather than reading directly from a file

## Test
* subscribe to remaining_useful_life rostopic within power system node


jfugate1:
I'll have to make a minor adjustment to the format of the output as it is currently outputting a float when it appears that RUL is simply an integer in units of seconds

jfugate1:
Also note that this only includes RUL for the time being with the code for SOC and Temperature commented out. Updates are required on the GSAP end in order to utilize SOC and Temp values properly.

Samahu:
How would I tell in a black box testing ... that the code is using the battery model implemented in GSAP to report the RUL and not the precomputed values that were once stored in the csv files. Is there a way to distinguish that without checking the code?

ow_power_system/src/power_system_node.cpp
```diff
@@ -90,7 +39,16 @@ int main(int argc, char* argv[]) {
...
+  // Create a configuration from a file
+  string config_path = ros::package::getPath("ow_power_system")+"/config/example.cfg";
+  ConfigMap config(config_path);
+  // Contruct a new prognoser using the prognoser factory. The prognoser
+  // will automatically construct an appropriate model, observer and predictor
+  // based on the values specified in the config.
+  auto prognoser = PrognoserFactory::instance().Create("ModelBasedPrognoser", config);
 
   // Retrieve our publication rate expressed in Hz
   double power_update_rate = csv_path.find("onewatt") != string::npos ? 0.1 : 1;  // if onewatt default to 0.1 Hz otherwise 1 Hz
```
Samahu, line 54: The **power_update_rate** was based on the name of the three precomputed csv files. If these files are no longer going to be an option (which doesn't seem the case since the code here has certain expectation of the loaded csv file) then this line should be changed to something like this ("onewatt is not an option and the condition will always evaluate to 1): 
```suggestion
  double power_update_rate = 1;
```

ow_power_system/src/power_system_node.cpp
```diff
@@ -129,23 +79,69 @@ int main(int argc, char* argv[]) {
...
+          continue;
+      }
+      auto eod_event = prediction.getEvents().front();
+      // The time of event is a `UData` structure, which represents a data
+      // point while maintaining uncertainty. For the MonteCarlo predictor
+      // used by this example, the uncertainty is captured by storing the
+      // result of each particle used in the prediction.
+      UData eod_time = eod_event.getTOE();
+      if (eod_time.uncertainty() != UType::Samples) {
+        std::cerr << "Unexpected uncertainty type for EoD prediction" << std::endl;
```
Samahu, line 101: Probably better to use `ROS_ERROR_NAMED` to ros-ify the code

ow_power_system/src/power_system_node.cpp
```diff
@@ -129,23 +79,69 @@ int main(int argc, char* argv[]) {
...
+        return 1;
```
Samahu, line 102: I understand this is part of their sample code but does there logic carries out to our use case? Why is error catastrophic and requires shutting down the power node? Can you elaborate?

Samahu (changes requested):
This definitely needs changes and fixes some of them I can't comment on since they don't show in the PR
* the function **load_csv** is no longer referenced and it doesn't seem we are going to use it in the future so it should be removed from `power_system_util.cpp`
  - also remove the declaration from `power_system_util.h` 
* The three csv files `onewatt.csv`, `eightwatt.csv` and `sixteenwatt.csv` no longer work with the current code and better be removed.
* The `README.md` should be updated accordingly

ow_power_system/src/power_system_node.cpp
jfugate1: Very good points @Samahu. I think I rushed to get this PR in before the end of the day to show the functionality is there without tending to some of the other "cleanup" related issues created by the change. I should have taken the time to address these items beforehand. I was planning a second cleanup commit, but didn't anticipate your prompt review. Thank you

jfugate1:
> How would I tell in a black box testing ... that the code is using the battery model implemented in GSAP to report the RUL and not the precomputed values that were once stored in the csv files. Is there a way to distinguish that without checking the code?

@Samahu It's not easy to tell based on the fact that RUL is the only rostopic being generated at this time, and that was a relatively accurate value in the preexisting code base. I could add something to the info stream that says the values are being computed. Once the temperature/SOC are added, it will be clear. Also, opening the data_const_load.csv file reveals that RUL values are not included in a precomputed column as they were for onewatt.csv

ow_power_system/src/power_system_node.cpp
jfugate1: True. This will be changed.

ow_power_system/src/power_system_node.cpp
jfugate1: Yes. This is being changed

ow_power_system/src/power_system_node.cpp
jfugate1: This was an artifact of the example code that I had not yet cleaned up. The catastrophic nature of the error will be removed such that it is simply a ROS error printout

jfugate1:
> This definitely needs changes and fixes some of them I can't comment on them directly since they are not part of PR
> 
> * the function **load_csv** is no longer referenced and it doesn't seem we are going to use it in the future so it should be removed from `power_system_util.cpp`
>   
>   * also remove the declaration from `power_system_util.h`
> * The three csv files `onewatt.csv`, `eightwatt.csv` and `sixteenwatt.csv` no longer work with the current code and better be removed.
> * The `README.md` should be updated accordingly

Great points. Thank you.

ow_power_system/src/power_system_node.cpp
jfugate1: Resolved.

ow_power_system/src/power_system_node.cpp
jfugate1: Resolved.

ow_power_system/src/power_system_node.cpp
jfugate1: Resolved.

Samahu:
> > How would I tell in a black box testing ... that the code is using the battery model implemented in GSAP to report the RUL and not the precomputed values that were once stored in the csv files. Is there a way to distinguish that without checking the code?
> 
> @Samahu It's not easy to tell based on the fact that RUL is the only rostopic being generated at this time, and that was a relatively accurate value in the preexisting code base. I could add something to the info stream that says the values are being computed. Once the temperature/SOC are added, it will be clear. Also, opening the data_const_load.csv file reveals that RUL values are not included in a precomputed column as they were for onewatt.csv

This explanation is sufficient, no need to take further action on this.

kmdalal:
@jfugate1, what is type and unit for RUL?  I'm currently treating it as real number and getting real values (non-zero decimal part).  As for unit, I can't tell by the value.  Reference Mission 1 starts with an RUL of about 2900.

kmdalal:
I've done black-box testing in which I run autonomy plans that print the RUL every 120 seconds.  While I'm now seeing monotonically decreasing values as expected, the rate of decreasing RUL does not seem well-correlated with what the lander is doing.  In particular, for a plan that does nothing, the RUL decreases by widely varying values -- I would expect a fairly constant decline.  And there's little difference between the burn rate of the do-nothing plan vs Ref Mission 1 which is digging and sampling.   I wish I could show you graphs, but my rqt_plot is broken.

jfugate1:
> @jfugate1, what is the type and unit for RUL? I'm currently treating it as real number and getting real values (non-zero decimal part). As for unit, I can't tell by the value. Reference Mission 1 starts with an RUL of about 2900.

@mikedalal The unit of RUL is seconds. It is currently set as a Float64, but should likely be converted to an integer, as the GSAP calculation results in no variance in the decimal output. I will make this change.

ow_power_system/src/power_system_node.cpp
```diff
@@ -17,71 +17,20 @@
...
+  //Construct our State of Charge (SOC) publisher (%)
   ros::Publisher SOC_pub = nh.advertise<std_msgs::Float64>("state_of_charge",1000);
   //Construct our Remaining Useful Life (RUL) publisher (Seconds)
   ros::Publisher RUL_pub = nh.advertise<std_msgs::Float64>("remaining_useful_life",1000);
   
   //Load power values csv
   string csv_path;
   auto csv_path_param_exist = nh.param("power_draw_csv_path", csv_path,
-    ros::package::getPath("ow_power_system") + "/data/onewatt.csv");
+    ros::package::getPath("ow_power_system") + "/data/data_const_load.csv");
```
kmdalal, line 33: So the CSV file is hardcoded, and the one specified in launch file argument is ignored?

jfugate1:
> I've done black-box testing in which I run autonomy plans that print the RUL every 120 seconds. While I'm now seeing monotonically decreasing values as expected, the rate of decreasing RUL does not seem well-correlated with what the lander is doing. In particular, for a plan that does nothing, the RUL decreases by widely varying values -- I would expect a fairly constant decline. And there's little difference between the burn rate of the do-nothing plan vs Ref Mission 1 which is digging and sampling. I wish I could show you graphs, but my rqt_plot is broken.

It's good that you are seeing monotonically decreasing RUL values; however, since the power input is still being read in from a precomputed csv, there should not be any correlation with the actual move or variance between missions. The shift to using the real power input values will be implemented in a later PR

ow_power_system/src/power_system_node.cpp
jfugate1: I believe that the way it is written, the launch CSV file is hardcoded; however, if a csv_path_parameter exists in the launch file, that path overwrites the hardcoded path. The default (if not present) is to use the hardcoded path

jfugate1: I believe that the way it is written, the launch CSV file is hardcoded; however, if a csv_path_parameter exists in the launch file, that path overwrites the hardcoded path. The default (if not present) is to use the hardcoded path

kmdalal:
> > I've done black-box testing in which I run autonomy plans that print the RUL every 120 seconds. While I'm now seeing monotonically decreasing values as expected, the rate of decreasing RUL does not seem well-correlated with what the lander is doing. In particular, for a plan that does nothing, the RUL decreases by widely varying values -- I would expect a fairly constant decline. And there's little difference between the burn rate of the do-nothing plan vs Ref Mission 1 which is digging and sampling. I wish I could show you graphs, but my rqt_plot is broken.
> 
> It's good that you are seeing monotonically decreasing RUL values; however, since the power input is still being read in from a precomputed csv, there should not be any correlation with the actual move or variance between missions. The shift to using the real power input values will be implemented in a later PR

Sorry, I hadn't read the README which explains this.  I'll change the RUL type to Integer eventually -- it's not hurting anything as Real for now.

jfugate1:
> > > I've done black-box testing in which I run autonomy plans that print the RUL every 120 seconds. While I'm now seeing monotonically decreasing values as expected, the rate of decreasing RUL does not seem well-correlated with what the lander is doing. In particular, for a plan that does nothing, the RUL decreases by widely varying values -- I would expect a fairly constant decline. And there's little difference between the burn rate of the do-nothing plan vs Ref Mission 1 which is digging and sampling. I wish I could show you graphs, but my rqt_plot is broken.
> > 
> > 
> > It's good that you are seeing monotonically decreasing RUL values; however, since the power input is still being read in from a precomputed csv, there should not be any correlation with the actual move or variance between missions. The shift to using the real power input values will be implemented in a later PR
> 
> Sorry, I hadn't read the README which explains this. I'll change the RUL type to Integer eventually -- it's not hurting anything as Real for now.

I should also mention that the slight variation you are seeing is likely due to the uncertainty in the Monte Carlo prediction (also included this info in the README)

ow_power_system/src/power_system_node.cpp
kmdalal: I see.  This code would be clearer (especially to someone who doesn't know the 'param' interface) if you:

- replace 'auto' with 'bool'
- use a named constant for the default CSV

ow_power_system/src/power_system_node.cpp
```diff
@@ -90,10 +39,19 @@ int main(int argc, char* argv[]) {
```
kmdalal, line 39: onewatt.csv is incorrect.  Use the named constant for the default file I suggested above.

ow_power_system/src/power_system_node.cpp
jfugate1: That's fair. I will make this change

ow_power_system/src/power_system_node.cpp
jfugate1: Not entirely sure what this comment is in reference to, but I believe I removed all references to onewatt.csv. I will make the change to a 'bool' variable and named constant in a new commit

kmdalal: onewatt.csv still shows up on line 37 of my version of this file, and I have pulled your latest branch.   Maybe your local copy is not committed/pushed?

kmdalal (approved):
Minor code comments only, looks fine and works as advertised.

ow_power_system/src/power_system_node.cpp
```diff
@@ -129,23 +79,66 @@ int main(int argc, char* argv[]) {
...
+      }
+      
+      // For this example, we will print the median EoD.
+      auto samples = eod_time.getVec();
+      std::sort(samples.begin(), samples.end());
+      double eod_median = samples.at(samples.size() / 2);
+      auto now =  MessageClock::now();
+      auto now_s = duration_cast<std::chrono::seconds>(now.time_since_epoch());
+      double rul_median = eod_median - now_s.count();
+      rul_msg.data = rul_median;
```
lanseafood, line 111: Last time I spoke with Chetan he mentioned that RUL was a calculation based partially on the SOC value. Does this new function take that into account? My PR also remaps the SOC topic and updates that with fault values when triggered, so that new fault value should be the new value used to RUL calculations (assuming that's still the case).

ow_power_system/src/power_system_node.cpp
```diff
@@ -129,23 +79,66 @@ int main(int argc, char* argv[]) {
...
+      //std::vector<double> medianSystemState(systemStates.size());
+      //for (int i = 0; i < medianSystemState.size(); i++) 
+        //medianSystemState[i] = systemStates[i][MEDIAN]; // Where MEDIAN corresponds to the sampleID of the median EOL prediction
+        
+      // Next, You will pass it into the model outputEqn to get the outputs:
+      //auto output = model.outputEqn(medianSystemState);
+      //double temperature = output[1];
+      
+      //publish current SOC & RUL
+      //SOC_pub.publish(soc_msg);
```
lanseafood, line 135: is SOC no longer being published here at all? If so where is that being moved to?

ow_power_system/src/power_system_node.cpp
jfugate1: Ahh nope, you're right! I missed that one bc it's a longer line, so the portion containing that reference had been cut off in my editing window. Good catch!

ow_power_system/src/power_system_node.cpp
jfugate1: From what I understand, RUL is based on the End of Discharge (EOD) calculation (simply EOD - current time), which is extracted from a separate state matrix than that which contains SOC. That being said, both are based on the power and voltage inputs, so they are certainly intertwined in ways that are hidden within the black box of GSAP. I had not discussed this with Chetan (mainly bc he was out most of the week), but I will touch base with him to clarify.

ow_power_system/src/power_system_node.cpp
jfugate1: Sorry, you missed this in the tagup meeting yesterday: the SOC value is inaccessible based on the current framework of GSAP. I spoke with the main author of the software, Chris Teubert, who informed me that a GSAP PR had been created to change this that will hopefully be merged soon. Same goes for temperature. Mike suggested, however, that in the meantime, it might be useful to create a PR for the updated RUL calculation (which actually uses the prognostics library rather than just reading from a csv), and then create a separate PR once those two values are available in GSAP. You'll notice that the code for SOC and temperature is there, but has been commented out. For now, you might use the existing code base, which includes the place-holding (albeit very inaccurate) SOC values, but keep in mind that my next PR will resolve this and the range of the SOC values should be between 0 and 1 (as opposed to the current values that are around 4)

ow_power_system/src/power_system_node.cpp
jfugate1: Resolved

jfugate1:
> I'll have to make a minor adjustment to the format of the output as it is currently outputting a float when it appears that RUL is simply an integer in units of seconds

Done.

ow_power_system/src/power_system_node.cpp
jfugate1: Resolved

jfugate1:
Logic has been updated such that commented out SOC publishing works with updatedSavePts feature branch in GSAP. Comments will be removed pending merge of branch into master.

Samahu:
> > @jfugate1, what is the type and unit for RUL? I'm currently treating it as real number and getting real values (non-zero decimal part). As for unit, I can't tell by the value. Reference Mission 1 starts with an RUL of about 2900.
> 
> @mikedalal The unit of RUL is seconds. It is currently set as a Float64, but should likely be converted to an integer, as the GSAP calculation results in no variance in the decimal output. I will make this change.

I am a bit skeptic about this... If this is the case why did they choose to use `double` to represent their generated samples?
Could the fact that we are observing complete numbers with no fractions be a result of a given setting or maybe they have used `double` to hold integer values for now with the intention for future functionality.

Samahu:
also note that we are currently using the `data_const_load.csv` and the `data_variable_load.csv` for now, both of these files have a timestamp column that is assumed to have a unit of seconds when the file gets loaded. In a real, implementation we will set the timestamp at runtime from ros clock, also the rate at which we query **prognoser** (which is currently set to 1 Hz) may play part in why are seeing the EoD sampled values rounded to whole numbers with no fractions. I'd say staying with `Float64` would have been the safe option but we can discuss this more in detail tomorrow.

jfugate1:
@Samahu I changed the GSAP uncertainty error to a warning and added a loop counter such that if the warning occurs on the first iteration, the power system node will publish 0 for RUL. Otherwise, it will publish the most recent value. This way nothing else breaks by not producing a value when the warning occurs.

ow_power_system/src/power_system_node.cpp
```diff
@@ -113,39 +75,94 @@ int main(int argc, char* argv[]) {
...
-  int rul_i = 2;
-  
   // ROS Loop. Note that once this loop starts,
   // this function (and node) is terminated with an interrupt.
   
   ros::Rate rate(power_update_rate);
   //individual soc_msg to be published by SOC_pub
   std_msgs::Float64 soc_msg;
-  std_msgs::Float64 rul_msg;
+  std_msgs::Int16 rul_msg;
```
Samahu, line 84:
```suggestion
  std_msgs::Int16 rul_msg;
  rul_msg.data = 0;  // immediately set to a good default
```

ow_power_system/src/power_system_node.cpp
```diff
@@ -113,39 +75,94 @@ int main(int argc, char* argv[]) {
...
+  std_msgs::Float64 tempbat_msg;
   ROS_INFO ("Power system node running");
 
-  int power_lifetime = power_csv.size();
+  int count = 0;
+
```
Samahu, line 89: No need for the counter if you already have the rul_msg.data in a good default
```suggestion
```

ow_power_system/src/power_system_node.cpp
```diff
@@ -113,39 +75,94 @@ int main(int argc, char* argv[]) {
...
+          //TempBat_pub.publish(tempbat_msg);
+        }
+        ros::spinOnce();
+        rate.sleep();
+        continue;
+      }
+      
+      // For this example, we will print the median EoD.
+      auto samples = eod_time.getVec();
+      std::sort(samples.begin(), samples.end());
```
Samahu, line 128: Paraphrasing here, this can be simplified as follows (this code assumes you have set `rul_msg.data = 0` on the start:
```suggestion
      UData eod_time = eod_event.getTOE();
      if (eod_time.uncertainty() != UType::Samples) {
        // Log warning and don't update the last value
        ROS_WARN_NAMED("power_system_node", "Unexpected uncertainty type for EoD prediction");
      } else { // valid prediction
        // For this example, we will print the median EoD.
        auto samples = eod_time.getVec();
        std::sort(samples.begin(), samples.end());
        ...
        rul_msg.data = rul_median;
      }
```

ow_power_system/src/power_system_node.cpp
```diff
@@ -113,39 +75,94 @@ int main(int argc, char* argv[]) {
...
+      // Next, You will pass it into the model outputEqn to get the outputs:
+      //auto output = model.outputEqn(medianSystemState);
+      //double temperature = output[1];
+      //tempbat_msg.data = temperature;
+      
+      //publish current SOC & RUL
+      //SOC_pub.publish(soc_msg);
+      RUL_pub.publish(rul_msg);
+      //TempBat_pub.publish(tempbat_msg);
+      count+=1;
```
Samahu, line 161: Not needed if you have followed the previous steps
```suggestion
```

Samahu (changes requested):
@jfugate1 you code should work but unnecessarily you used more code and duplicated some function calls `publish`, `ros::spinOnce`, `sleep`,...
Check the provided code tips to avoid this

jfugate1:
> @jfugate1 you code should work but unnecessarily you used more code and duplicated some function calls `publish`, `ros::spinOnce`, `sleep`,...
> Check the provided code tips to avoid this

Thanks @Samahu. I committed the changes as you suggested. Definitely much cleaner this way.

jfugate1:
Fixed the paraphrased suggested correction.

Samahu (approved):
Great Work!

PR #98, opened 2021-03-05
OCEANWATER-581 Power Fault Impact Ros Topics

lanseafood:
## Linked Issues:
| Jira Ticket 🎟️   | [Oceanwater-581](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-581)  [Oceanwater-622](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-622)|
| ----------- | ----------- |
| Github :octocat:  | #111 (cleanup) |

Please Note: 
This PR will act as the base PR for a series of power faults prs. 

## Summary of Changes
*  faults node published a trigger topic when users select /rqt
** low volt = 1, cap loss = 2, thermal = 4 (in bits)
** power node listens to this topic /faults/power_faults
* power node sets node booleans to determine which power fault has been triggered
** new power node function sets the new value of the mechanical power depending on which fault is chosen
* faults node listens for selection of any fault and checks SOC values and TEMP values to publish power faults message to topic /power_faults_status

## Test

[Oceanwater-581](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-581)  
 TEST RQT FAULT TRIGGER:
* see Ussama's power PR #113 

[Oceanwater-622](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-622)
TEST FAULTS OUTPUTTING TRIGGERS FOR  LOW VOLTAGE / CAP LOSS / THERMAL 
* start any launch file
* `rostopic echo /faults/jpl/power_faults_status`
* see value as hardware fault (1) for any power fault

[future PR, do not test yet]
TEST REAL POWER OUTPUT 
* roslaunch any launch file
* select any:
** `rostopic echo /power_system_node/state_of_charge`
*** for low volt, value < 10
*** for cap loss, value difference should be > 5%
** `rostopic echo /power_system_node/battery_temperature`
*** temperature must be > 50C


ow_faults/src/FaultInjector.cpp
```diff
@@ -75,16 +76,54 @@ void FaultInjector::setFaultsMessage(ow_faults::PTFaults& msg, ComponentFaults v
...
-    thermal_val =  50.0 + static_cast <float> (rand()) /( static_cast <float> (RAND_MAX/(90.0-50.0)));
-    powerTemperatureOverloadValue = thermal_val;
-  } else if (!getTempBool) {
-    powerTemperatureOverloadValue = NAN;
+float FaultInjector::getRandomValueFromRange( float min_val, float max_val){
+  float randomValue = min_val + static_cast <float> (rand()) /( static_cast <float> (RAND_MAX/(max_val-min_val)));
+  return randomValue;
+}
+
+void FaultInjector::setPowerFaultValues(string powerType, float min_val, float max_val){
```
kmdalal, line 84: Minor: const string& is better, and idiomatic for pass-by-value strings.

ow_faults/include/ow_faults/FaultInjector.h
```diff
@@ -67,13 +67,18 @@ class FaultInjector
   
 private:
   float powerTemperatureOverloadValue;
+  float powerStateOfChargeValue;
+  float originalSOC;
+  // float powerStateOfChargeValue;
   
   // renders new temperature when thermal power fault is re-triggered
-  void setPowerTemperatureFaultValue(bool getTempBool);
+  float getRandomValueFromRange(float min_val, float max_val);
```
kmdalal, line 75: Minor: you could rename this function for the probability distribution it uses, if that's knowable and this is useful.

kmdalal (approved):
Minor code comments, features worked as advertised.

ow_faults/include/ow_faults/FaultInjector.h
lanseafood: gotcha! will do

ow_faults/src/FaultInjector.cpp
lanseafood: good point, will do!

Samahu (changes requested):
I haven't checked code yet .. but when merely performing the test procedure I have noticed that the original topic had a different publish rate than the remapped one. I don't think that remapped topics should have a different publish rate than the original topic!

lanseafood:
> I haven't checked code yet .. but when merely performing the test procedure I have noticed that the original topic had a different publish rate than the remapped one. I don't think that remapped topics should have a different publish rate than the original ones!

Yes I believe that I have them publishing at the rate of all other faults. I can adjust the remapped topic to update to be the same as the original though.

Samahu:
> > I haven't checked code yet .. but when merely performing the test procedure I have noticed that the original topic had a different publish rate than the remapped one. I don't think that remapped topics should have a different publish rate than the original ones!
> 
> Yes I believe that I have them publishing at the rate of all other faults. I can adjust the remapped topic to update to be the same as the original though.

I think the intent is to manipulate the values of topics as part of faults. We can later on manipulate the publish rate of the topics too but ultimately we should maintain the same publish rate of each topic that we manipulate.

ow_faults/src/FaultInjector.cpp
```diff
@@ -75,16 +76,54 @@ void FaultInjector::setFaultsMessage(ow_faults::PTFaults& msg, ComponentFaults v
...
-void FaultInjector::setPowerTemperatureFaultValue(bool getTempBool){
-  float thermal_val;
-  if (isnan(powerTemperatureOverloadValue)) {
-    thermal_val =  50.0 + static_cast <float> (rand()) /( static_cast <float> (RAND_MAX/(90.0-50.0)));
-    powerTemperatureOverloadValue = thermal_val;
-  } else if (!getTempBool) {
-    powerTemperatureOverloadValue = NAN;
+float FaultInjector::getRandomValueFromRange( float min_val, float max_val){
+  float randomValue = min_val + static_cast <float> (rand()) /( static_cast <float> (RAND_MAX/(max_val-min_val)));
+  return randomValue;
```
Samahu, line 81: This formula is not correct. The term `max_val-min_val` should be multiplied by the term `rand/rand_max` not divided! + you may simplify things a bit as follow
```suggestion
  return min_val + (max_val - min_val) * (rand() / float(RAND_MAX));
```

Samahu (changes requested):
Correcting the formula used `getRandomFloatFromRange()` will definitely have an impact on the dependent method `setPowerFaultValues()`, make sure the induced failures still provided the expected outcome.

ow_faults/src/FaultInjector.cpp
lanseafood: updated and can confirm still works as expected.

ow_faults/src/FaultInjector.cpp
```diff
@@ -19,8 +19,10 @@ FaultInjector::FaultInjector(ros::NodeHandle node_handle)
   m_joint_state_pub = node_handle.advertise<sensor_msgs::JointState>("/joint_states", 10); 
 
   //power fault publishers and subs
-  m_fault_power_state_of_charge_pub = node_handle.advertise<std_msgs::Float64>("temporary/power_fault/state_of_charge", 10);
-  m_fault_power_temp_pub = node_handle.advertise<std_msgs::Float64>("temporary/power_fault/temp_increase", 10);
+  m_power_soc_sub = node_handle.subscribe("/_original/power_system_node/state_of_charge", 10, &FaultInjector::powerSOCListener, this);
+  m_fault_power_state_of_charge_pub = node_handle.advertise<std_msgs::Float64>("/power_system_node/state_of_charge", 10);
+  m_fault_power_temp_pub = node_handle.advertise<std_msgs::Float64>("/temporary/power_fault/temp_increase", 10);
+  timerPublishTemperature = node_handle.createTimer(ros::Duration(60), std::bind(&FaultInjector::powerFaultCb, this));
```
Samahu, line 25: I don't think you need a timer for our case here
```suggestion
```

Samahu: We can instead just use the original topic callback as the trigger to publish a new message

ow_faults/src/FaultInjector.cpp
```diff
@@ -32,6 +34,7 @@ FaultInjector::FaultInjector(ros::NodeHandle node_handle)
   m_antennae_fault_status_pub = node_handle.advertise<ow_faults::PTFaults>("/pt_faults_status", 10);
 
   srand (static_cast <unsigned> (time(0)));
+  // ros::spin();
```
Samahu, line 37:
```suggestion
```

ow_faults/src/FaultInjector.cpp
```diff
@@ -75,16 +78,78 @@ void FaultInjector::setFaultsMessage(ow_faults::PTFaults& msg, ComponentFaults v
...
+      powerTemperatureOverloadValue += change_in_val;
+    }
+  }
+}
+
+void FaultInjector::powerSOCListener(const std_msgs::Float64& msg)
+{
+  originalSOC = msg.data;
+
+}
```
Samahu, line 151: Use the original topic callback as a trigger instead of a timer, this way you don't need to worry about what is the publish rate of the original topic currently is 
```suggestion
void FaultInjector::powerSOCListener(const std_msgs::Float64& msg)
{
  originalSOC = msg.data;
  powerFaultCb();
}
```

Samahu (changes requested):
I don't understand why you went this route for maintaining the original publish rate. Couldn't you just use the original topic callback as a trigger for crafting and publishing the manipulate power value?

With the current implementation:
A. The original and manipulated topic even might be having the same publish rate but they do publish at quite distinct times
B. When no faults are enabled there seem to be a disparity between what the original SOC value is and the un-manipulated SOC value. That's likely due to the asynchronous nature of the current implementation
C. Further more the original topic could potentially change its publish rate at runtime, meanwhile your timer is hard-coded.

Please consider using the original topic callback as the trigger rather than a timer based trigger.

ow_faults/src/FaultInjector.cpp
lanseafood: I believe we had discussed in the dev day that they wanted the publish rate to be 60 seconds. I don't know what the original topic had as the rate and I believe that will be updated since Jason's PR removed publishing SOC altogether. We will need to wait until the new PR to see what the rate it. If their rate is also 1 minute, that would make your approach reasonable. If not, then we would still need to trigger a 60s publish rate ourselves.

Samahu: If the original topic currently doesn't  have the correct publish rate then it should be fixed by the originating node, not corrected by the faults injection node. When no faults injected your node should just relay the information as is with no delay.

lanseafood: fair enough, will update it to be no timer. I'm also going to preset the soc to be a value because otherwise jason's PR would break this work.

lanseafood:
reversed changes for using fake soc

Samahu (commented):
This looks good in terms of getting the same update rate as the original topics.
However, I feel there are still two things that are worth noting:
1. The drop in power is a factor of the frequency of update rather than a factor of time
    * This needs to be addressed but can be handled in a separate ticket 
2. Once the fault is lifted what should be the expected behavior, do we return to the original value or continue from current point?
   * This point is worth some discussion, current behavior just emits the original value

Samahu:
Another point that I have noticed whilst testing is that when a power fault is in effect and the power value drops to near 0.5 it stops at that level, I haven't looked fully in the power drop models but this doesn't seem to be realistic.

Samahu (approved):
factoring in the time for the simulated power drop to be addressed in a separate ticket

Samahu (changes requested):
Based on the description of [OCEANWATER-581](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-581) you are supposed to handle the **remaining_useful_life** too.

ow_faults/src/FaultInjector.cpp
```diff
@@ -26,8 +29,10 @@ FaultInjector::FaultInjector(ros::NodeHandle node_handle)
   m_dist_pitch_ft_sensor_pub = node_handle.advertise<geometry_msgs::WrenchStamped>(ft_sensor_dist_pitch_str, 10);
 
   //power fault publishers and subs
-  m_fault_power_state_of_charge_pub = node_handle.advertise<std_msgs::Float64>("temporary/power_fault/state_of_charge", 10);
-  m_fault_power_temp_pub = node_handle.advertise<std_msgs::Float64>("temporary/power_fault/temp_increase", 10);
+    // will return to checking state of charge once that returns
+  m_power_soc_sub = node_handle.subscribe("/power_system_node/state_of_charge", 1000, &FaultInjector::powerSOCListener, this); 
+  m_power_temp_sub = node_handle.subscribe("/power_system_node/battery_temperature", 1000, &FaultInjector::powerTempListener, this);
```
kmdalal, line 34: Curious if there's a specific reason why the queue size is so large (1000).  I've typically used and seen values < 10.

ow_faults/src/FaultInjector.cpp
```diff
@@ -103,14 +109,59 @@ void FaultInjector::antennaeTiltFaultCb(const std_msgs::Float64& msg){
...
 
-void FaultInjector::setPowerTemperatureFaultValue(bool getTempBool){
-  float thermal_val;
-  if (isnan(powerTemperatureOverloadValue)) {
-    thermal_val =  50.0 + static_cast <float> (rand()) /( static_cast <float> (RAND_MAX/(90.0-50.0)));
-    powerTemperatureOverloadValue = thermal_val;
-  } else if (!getTempBool) {
-    powerTemperatureOverloadValue = NAN;
+float FaultInjector::getRandomFloatFromRange( float min_val, float max_val){
+  return min_val + (max_val - min_val) * (rand() / float(RAND_MAX));
```
kmdalal, line 113: Minor comment: static_cast<float>(RAND_MAX) is better, I think.

ow_faults/src/FaultInjector.cpp
```diff
@@ -103,14 +109,59 @@ void FaultInjector::antennaeTiltFaultCb(const std_msgs::Float64& msg){
...
+  setBitsetFaultsMessage(system_faults_msg, systemFaultsBitmask);
+  m_fault_status_pub.publish(system_faults_msg);
+  m_power_fault_status_pub.publish(power_faults_msg);
+
+}
+
+void FaultInjector::powerTempListener(const std_msgs::Float64& msg)
+{
+  bool fault = false;
+  if(m_faults.thermal_power_failure && msg.data > 50){
```
kmdalal, line 143: Magic number 50 - don't know what it means, a named constant would clarify.

ow_faults/src/FaultInjector.cpp
```diff
@@ -103,14 +109,59 @@ void FaultInjector::antennaeTiltFaultCb(const std_msgs::Float64& msg){
...
+}
+
+void FaultInjector::powerSOCListener(const std_msgs::Float64& msg)
+{
+  bool fault = false;
+  float newSOC = msg.data;
+  if (isnan(originalSOC)){
+    originalSOC = newSOC;
+  }
+  if ((m_faults.low_state_of_charge_power_failure && newSOC <= 10)  ||
```
kmdalal, line 157: Magic number 10

ow_power_system/include/power_system_node.h
```diff
@@ -5,20 +5,32 @@
...
+#include <ow_faults/SystemFaults.h>
 
 class PowerSystemNode
 {
 public:
   void Run();
 
+  int LOW_VOLTAGE = 1;
+  int CAP_LOSS = 2;
+  int THERMAL_FAULT = 4;
```
kmdalal, line 17: I think these would better be an enum rather than public class data, since they are effectively constants.  And member data are required to have an "m_" prefix.  If they are used only in the .cpp file, they can be defined there.

ow_power_system/src/power_system_node.cpp
```diff
@@ -17,8 +16,8 @@ using namespace std::chrono;
 void PowerSystemNode::powerCallback(const std_msgs::Float64::ConstPtr& msg)
 {
   // Set mechanical power value to rostopic subscription
-  double mechanical_power = msg->data;  // [W]
-
+  double mechanical_power = checkForFaults(msg->data);  // [W]
+  
   // Temperature estimate based on pseudorandom noise and fixed range
   double temperature_estimate =
       m_min_temperature + static_cast<double>(rand()) / RAND_MAX * (m_max_temperature - m_min_temperature);
```
kmdalal, line 23: Minor comment: contrast with FaultInjector::getRandomFloatFromRange, assuming they do the same thing except use double vs float.   Factoring possible but not important.

ow_power_system/src/power_system_node.cpp
```diff
@@ -102,6 +101,32 @@ void PowerSystemNode::powerCallback(const std_msgs::Float64::ConstPtr& msg)
...
+  if (m_lowVoltageFault) {
+    std::cout << "low volt" << std::endl;
+    // return 30.0;
+  }
+  if (m_capLossFault) {
+    std::cout << "cap loss" << std::endl;
+    // return 20.0;
+  }
+  if (m_thermalFault) {
+    std::cout << "thermal" << std::endl;
```
kmdalal, line 117: Assuming the couts are for debugging.   Use ROS_INFO, etc otherwise.

ow_power_system/src/power_system_node.cpp
```diff
@@ -128,6 +153,7 @@ void PowerSystemNode::Run()
 
   // Finally subscribe to mechanical power topic (Watts)
   m_mechanical_power_sub = m_nh.subscribe("/mechanical_power/average", 1, &PowerSystemNode::powerCallback, this);
+  m_power_fault_sub = m_nh.subscribe("/faults/power_faults", 1, &PowerSystemNode::powerFaultsCallback, this);
```
kmdalal, line 156: I'm noting the queue size of 1 here, vs 1000 in other subscriptions.  I don't know what the best choices are, but curious if there's a motivation for these values.

ow_faults/src/FaultInjector.cpp
lanseafood: that was the prior publish rate from the power node, but I can double check on if that's changed.

ow_faults/src/FaultInjector.cpp
lanseafood: good point, will update

kmdalal:
I'm confused that the test mentions OW-622.   Isn't this just a review and test of branch OW-581?

ow_power_system/src/power_system_node.cpp
```diff
@@ -102,6 +101,32 @@ void PowerSystemNode::powerCallback(const std_msgs::Float64::ConstPtr& msg)
   m_battery_temperature_pub.publish(battery_temperature_msg);
 }
 
+double PowerSystemNode::checkForFaults(double originalValue)
```
kmdalal, line 104: It seems this function could be better named given how it is used, since it modifies the original mechanical power value.  Something like adjustMechPowerForFaults() or adjustForFaults() , etc.

kmdalal:
The OW-581 test passes for me.  Is this the full scope of this PR?    The test for OW-622 interests me more, but it seems that is a future PR...

ow_faults/src/FaultInjector.cpp
kmdalal: My understanding of the queue size is that is how many messages are allowed to queue up.   Not sure how this relates numerically to the publish rate of the publisher, but would like to understand it better...

lanseafood:
> The OW-581 test passes for me. Is this the full scope of this PR? The test for OW-622 interests me more, but it seems that is a future PR...

test 622 is the second test. I put both 622 and 581 in this single PR.

ow_power_system/src/power_system_node.cpp
lanseafood: Ussama/Jason wrote that separately from the Fault work, but yea we could refactor this in the future.

ow_power_system/src/power_system_node.cpp
lanseafood: I plan to fix this in a future PR when we know what values are needed

ow_power_system/src/power_system_node.cpp
lanseafood: Here I'm keeping the same value as what the mech power sub is using in order to make sure it's using the most updated values from faults.

ow_power_system/include/power_system_node.h
lanseafood: updating to static constexpr ints instead.

ow_power_system/src/power_system_node.cpp
lanseafood: updated in #111

Samahu:
My take on this is you are adding an additional hop that doesn't seem to be needed

in your **FaultInjector::jointStateCb** you have the following snippet:
```C++
    // power faults
  if (m_faults.low_state_of_charge_power_failure) {
    powerFaultsBitmask |= islowVoltageError;
  } 
  if (m_faults.instantaneous_capacity_loss_power_failure){
    powerFaultsBitmask |= isCapLossError;
  } 
  if (m_faults.thermal_power_failure) {
    powerFaultsBitmask |= isThermalError;
  } 
```
Merely checks for the dynamic_reconfigure flags and then forms a trigger fault message that is published on `/faults/trigger/power..`. The **PowerSystemNode** subscribes to this trigger topic and repeats the three checks again as follows:
 
```C++
double PowerSystemNode::checkForFaults(double originalValue)
This conversation was marked as resolved by lanseafood
{
  //will likely need to modify after Chetan's feedback
  // assuming that we can only have one power fault triggered at a time.
  if (m_lowVoltageFault) {
    std::cout << "low volt" << std::endl;
    // return 30.0;
  }
  if (m_capLossFault) {
    std::cout << "cap loss" << std::endl;
    // return 20.0;
  }
  if (m_thermalFault) {
    std::cout << "thermal" << std::endl;
}
```

I don't see the point of this hop and this repetition .. The trigger message is really private to the PowerSystemNode, no other nodes would listen to it, plus it can be confused with the actual **ow_faults::PowerFaults** message.

My recommendation is to check the dynamic reconfigure power fault flags directly in **PowerSystemNode** and avoid this unnecessary indirection. Think of delay and code maintainability with the latter being of a higher concern to me.

@anjan1984 @kmdalal for perspective on the point that I was trying to make

Samahu:
> My take on this is you are adding an additional hop that doesn't seem to be needed
> 
> in your **FaultInjector::jointStateCb** you have the following snippet:
> 
> ```c++
>     // power faults
>   if (m_faults.low_state_of_charge_power_failure) {
>     powerFaultsBitmask |= islowVoltageError;
>   } 
>   if (m_faults.instantaneous_capacity_loss_power_failure){
>     powerFaultsBitmask |= isCapLossError;
>   } 
>   if (m_faults.thermal_power_failure) {
>     powerFaultsBitmask |= isThermalError;
>   } 
> ```
> 
> Merely checks for the dynamic_reconfigure flags and then forms a trigger fault message that is published on `/faults/trigger/power..`. The **PowerSystemNode** subscribes to this trigger topic and repeats the three checks again as follows:
> 
> ```c++
> double PowerSystemNode::checkForFaults(double originalValue)
> {
>   //will likely need to modify after Chetan's feedback
>   // assuming that we can only have one power fault triggered at a time.
>   if (m_lowVoltageFault) {
>     std::cout << "low volt" << std::endl;
>     // return 30.0;
>   }
>   if (m_capLossFault) {
>     std::cout << "cap loss" << std::endl;
>     // return 20.0;
>   }
>   if (m_thermalFault) {
>     std::cout << "thermal" << std::endl;
> }
> ```
> 
> I don't see the point of this hop and this repetition .. The trigger message is really private to the PowerSystemNode, no other nodes would listen to it, plus it can be confused with the actual **ow_faults::PowerFaults** message.
> 
> My recommendation is to check the dynamic reconfigure power fault flags directly in **PowerSystemNode** and avoid this unnecessary indirection. Think of delay and code maintainability with the latter being of a higher concern to me.
> 
> @anjan1984 @kmdalal for perspective on the point that I was trying to make

@lanseafood In this PR Draft #112, you can see how we can [check for faults directly within the PowerSystemNode](https://github.com/nasa/ow_simulator/blob/a4cac6e9ecdc0e1ec0942bd374280281fbacea00/ow_power_system/src/power_system_node.cpp#L59-L76) without having to create a link (through publish/subscribe) between the Faults Node and the PowerSystemNode.  I think this would more maintainable than the current approach of triggering the power faults.  Note that the PR is a still a draft and not done yet .. I pushed to GitHub so we can discuss it before we commit to a given route.

lanseafood:
> @lanseafood In this PR Draft #112, you can see how we can [check for faults directly within the PowerSystemNode](https://github.com/nasa/ow_simulator/blob/a4cac6e9ecdc0e1ec0942bd374280281fbacea00/ow_power_system/src/power_system_node.cpp#L59-L76) without having to create a link (through publish/subscribe) between the Faults Node and the PowerSystemNode. I think this would more maintainable than the current approach of triggering the power faults. Note that the PR is a still a draft and not done yet .. I pushed to GitHub so we can discuss it before we commit to a given route.

@Samahu updated to reflect discussion today

Samahu (changes requested):
@lanseafood I think the way **publishPowerSystemFault** method is tapped into the different listeners, is going to be very confusing. Say you have detected a state of thermal fault which is checked for on the `/power_system_node/battery_temperature`. Assuming that the battery state of charge was in good standing (no faults for low state of charge or instantaneous capacity loss). Then anyone monitoring the `**/power_faults_status/` topic would get a varying status for the fault since the three listeners publish to the same topic. I think this is problematic and needs a reconsideration.

lanseafood:
> @lanseafood I think the way **publishPowerSystemFault** method is tapped into the different listeners, is going to be very confusing. Say you have detected a state of thermal fault which is checked for on the `/power_system_node/battery_temperature`. Assuming that the battery state of charge was in good standing (no faults for low state of charge or instantaneous capacity loss). Then anyone monitoring the `**/power_faults_status/` topic would get a varying status for the fault since the three listeners publish to the same topic. I think this is problematic and needs a reconsideration.

@Samahu the /fauls/jpl/power_faults_status is a general topic, much like the other components. Since all 3 power faults were determined to be 'hardware', they all publish the same value of 1(hardware) on the msg. This goes back to keeping the way we publish component jpl fault messages consistent.

Samahu:
> > @lanseafood I think the way **publishPowerSystemFault** method is tapped into the different listeners, is going to be very confusing. Say you have detected a state of thermal fault which is checked for on the `/power_system_node/battery_temperature`. Assuming that the battery state of charge was in good standing (no faults for low state of charge or instantaneous capacity loss). Then anyone monitoring the `**/power_faults_status/` topic would get a varying status for the fault since the three listeners publish to the same topic. I think this is problematic and needs a reconsideration.
> 
> @Samahu the /fauls/jpl/power_faults_status is a general topic, much like the other components. Since all 3 power faults were determined to be 'hardware', they all publish the same value of 1(hardware) on the msg. This goes back to keeping the way we publish component jpl fault messages consistent.

I understand that part, but possibly I haven't well explained the point that I was trying to raise. The way you are checking and announcing a fault indicates will be reflected as a state:
```C++
if( msg.data > THERMAL_MAX){
    fault = true;
  }
```
meaning while temperatures exceed the set limit the fault flag will always be **on** and this is what is sent through `**/power_faults_status/` topic. Meanwhile, say the other faults are not taking place at the current period, thus their fault status is **off** and this status sent through the same topic `**/power_faults_status/`. Monitoring this topic over a period of 3 seconds would thus yield the following status sequence: `true (thermal), false, false; true (thermal), false, false; true (thermal), false, false`. (Note don't put assumption on order ..it can be `false, true, false` ..). Thus if I am monitoring this topic to know when there is a power fault I would need to develop a logic of my own to know that this power fault is not being set and cleared within a second but it is actually _an ongoing power fault_.

Samahu:
> > > @lanseafood I think the way **publishPowerSystemFault** method is tapped into the different listeners, is going to be very confusing. Say you have detected a state of thermal fault which is checked for on the `/power_system_node/battery_temperature`. Assuming that the battery state of charge was in good standing (no faults for low state of charge or instantaneous capacity loss). Then anyone monitoring the `**/power_faults_status/` topic would get a varying status for the fault since the three listeners publish to the same topic. I think this is problematic and needs a reconsideration.
> > 
> > 
> > @Samahu the /fauls/jpl/power_faults_status is a general topic, much like the other components. Since all 3 power faults were determined to be 'hardware', they all publish the same value of 1(hardware) on the msg. This goes back to keeping the way we publish component jpl fault messages consistent.
> 
> I understand that part, but possibly I haven't well explained the point that I was trying to raise. The way you are checking and announcing a fault indicates it will be reflected as a state:
> 
> ```c++
> if( msg.data > THERMAL_MAX){
>     fault = true;
>   }
> ```
> 
> meaning while temperatures exceed the set limit the fault flag will always be **on** and this is what is sent through `**/power_faults_status/` topic. Meanwhile, say the other faults are not taking place at the current period, thus their fault status is **off** and this status sent through the same topic `**/power_faults_status/`. Monitoring this topic over a period of 3 seconds would thus yield the following status sequence: `true (thermal), false, false; true (thermal), false, false; true (thermal), false, false`. (Note don't put assumption on order ..it can be `false, true, false` ..). Thus if I am monitoring this topic to know when there is a power fault I would need to develop a logic of my own to know that this power fault is not being set and cleared within a second but rather there is actually _an ongoing power fault_.

If you don't think this is an issue then you may disregard my change request.

ow_faults/src/FaultInjector.cpp
```diff
@@ -288,6 +308,12 @@ void FaultInjector::checkAntFaults(){
                 m_faults.ant_tilt_encoder_failure || m_faults.ant_tilt_effort_failure);
 }
 
+void FaultInjector::checkPowerFaults(){
+  m_powerFault =  ( m_faults.low_state_of_charge_power_failure || 
+                    m_faults.instantaneous_capacity_loss_power_failure ||
+                    m_faults.thermal_power_failure);
+}
+
```
Samahu, line 316: I think what I was hoping for is something similar to what you had here in this no longer referenced method.
Then you would simply publish the **m_powerFault**. The challenge is however, you need to listen to at least two different topics to conclude that there is an **m_powerFault**.

Samahu: What you could probably do is set the variables `low_state_of_charge_power_failure` and `thermal_power_failure` each in its own listener and then produce and publish the value of m_powerFault based on that.

ow_faults/include/ow_faults/FaultInjector.h
```diff
@@ -58,97 +53,121 @@ class FaultInjector
...
+  void powerSOCListener(const std_msgs::Float64& msg);
+  void powerTemperatureListener(const std_msgs::Float64& msg);
 
-  //component faults
-  bool m_armFault;
-  bool m_antFault;
+  // Antennae functions
+  void antennaePanFaultCb(const std_msgs::Float64& msg);
+  void antennaeTiltFaultCb(const std_msgs::Float64& msg);
+  void publishAntennaeFaults(const std_msgs::Float64& msg, bool encoder, bool torque, float& m_faultValue, ros::Publisher& m_publisher);
```
kmdalal, line 107: Shorten line.

ow_faults/src/FaultInjector.cpp
```diff
@@ -26,8 +30,8 @@ FaultInjector::FaultInjector(ros::NodeHandle node_handle)
   m_dist_pitch_ft_sensor_pub = node_handle.advertise<geometry_msgs::WrenchStamped>(ft_sensor_dist_pitch_str, 10);
 
   //power fault publishers and subs
-  m_fault_power_state_of_charge_pub = node_handle.advertise<std_msgs::Float64>("temporary/power_fault/state_of_charge", 10);
-  m_fault_power_temp_pub = node_handle.advertise<std_msgs::Float64>("temporary/power_fault/temp_increase", 10);
+  m_power_soc_sub = node_handle.subscribe("/power_system_node/state_of_charge", 1000, &FaultInjector::powerSOCListener, this); 
+  m_power_temperature_sub = node_handle.subscribe("/power_system_node/battery_temperature", 1000, &FaultInjector::powerTemperatureListener, this);
```
kmdalal, line 34: Shorten lines.

ow_faults/src/FaultInjector.cpp
```diff
@@ -12,9 +12,13 @@ constexpr std::bitset<10> FaultInjector::isPanTiltExecutionError;
 constexpr std::bitset<10> FaultInjector::isArmExecutionError;
 constexpr std::bitset<10> FaultInjector::isPowerSystemFault;
 
+constexpr std::bitset<3> FaultInjector::islowVoltageError;
+constexpr std::bitset<3> FaultInjector::isCapLossError;
+constexpr std::bitset<3> FaultInjector::isThermalError;
 
 FaultInjector::FaultInjector(ros::NodeHandle node_handle)
```
kmdalal, line 19: Minor comment: const ros::NodeHandle& is preferable

ow_faults/src/FaultInjector.cpp
```diff
@@ -12,9 +12,13 @@ constexpr std::bitset<10> FaultInjector::isPanTiltExecutionError;
...
 {
+  //  arm pub and subs
   auto joint_states_str = "joint_states";
```
kmdalal, line 22: Minor comment: making this a static (file) variable, and then using a class initializer list instead of all the following assignments, is cleaner and a bit more efficient.

ow_faults/src/FaultInjector.cpp
kmdalal: Since this function (and the m_powerFault member) are not referenced, they should be removed, but I agree with @Samahu  that they appeared to be useful and have the right logic.

ow_faults/src/FaultInjector.cpp
```diff
@@ -93,24 +98,68 @@ void FaultInjector::antennaePanFaultCb(const std_msgs::Float64& msg){
...
+  m_system_fault_jpl_msg_pub.publish(system_faults_msg);
+  m_power_fault_jpl_msg_pub.publish(power_faults_msg);
+
+}
+
+void FaultInjector::powerTemperatureListener(const std_msgs::Float64& msg)
+{
+  bool fault = false;
+  if( msg.data > THERMAL_MAX){
+    fault = true;
```
kmdalal, line 143: Suggestion: 
bool fault = msg.data > THERMAL_MAX

or eliminate 'fault' variable and have just:
publishPowerSystemFault(msg.data > THERMAL_MAX);

kmdalal:
I resumed this review yesterday only to look at @Samahu's issue with /power_faults_status and sorry that I only made small code comments!  Autonomy subscribes to this topic (in the OW-658 branch) and I don't see any unusual behavior, in fact it works exactly as I expect.  To be honest I have not yet fully understood the fault framework implementation and will have some questions outside this review.

I suggest to let this issue rest for this review in the interest of completing our release on time. But keep the notes made, e.g. in our Confluence.  We should come up with a failing example and use that in a future Jira to fix it.

Samahu:
> I resumed this review yesterday only to look at @Samahu's issue with /power_faults_status and I only made small, unrelated code comments! Autonomy subscribes to this topic (in the OW-658 branch) and I don't see any unusual behavior, in fact it works exactly as I expect. To be honest I have not yet fully understood the fault framework implementation and will have some questions outside this review.
> 
> I suggest to let this issue rest for this review in the interest of completing our release on time. But keep the notes, e.g. in our Confluence. We should come up with a failing example if possible, and use that in a future Jira to fix it.

How did you test that it works exactly as you expected. Are you able to inject power fault using this PR alone? also this is dependent on what you do as a reaction to fault once it has been detected and or cleared. If you are only halting the execution one a fault has been raised but not trying to resume when the fault has been cleared then yea it wouldn't make a difference on what is been published to topic.

Again, I am not saying this is necessarily an issue .. all I am saying that it would be confusing to someone tapping into the topic and doesn't know the details of how this topic is handled internally.

kmdalal:
Good points, @Samahu.  And actually I tested my OW-658 branch against melodic-devel, not this branch.  I will retest against this branch, and see if I get "toggling" faults as we would expect.  I may write another short plan to test your concern as well.  Hope to have an answer this afternoon.

kmdalal:
In my latest test of this branch, two serious problems:

- the value of /faults/jpl/power_faults_status keeps toggling between 0 and 1 at each time step, regardless of how power faults are injected or now.
- the power_system_node suddenly crashed after a while with no explanation.  The referenced log file didn't exist (a common and annoying ROS problem).

I also noticed that nothing is published on /power_faults_status anymore.  If it has been replaced by /faults/jpl/power_faults_status it should be deleted.  I'd prefer a new name without JPL in it, e.g. just /faults/power_faults_status.

Samahu:
> * the power_system_node suddenly crashed after a while with no explanation.  The referenced log file didn't exist (a common and annoying ROS problem).

This could be related to the ticket [OCEANWATER-663 / PowerSystemNode crashes when consumed power is increased above a certain threshold](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-663) which I am still working on.

> I also noticed that nothing is published on /power_faults_status anymore. If it has been replaced by /faults/jpl/power_faults_status it should be deleted. On that note, I'd prefer a new name without JPL in it, e.g. just /faults/power_faults_status.

I noticed earlier the PR description still uses `/power_faults_status` but the code only has `/faults/jpl/power_faults_status`

kmdalal:
> This could be related to the ticket OCEANWATER-663 / PowerSystemNode crashes when consumed power is increased above a certain threshold which I am still working on.

Indeed, the crash happened after the sim ran a long time, and I saw the battery temp go up to 31 degrees for the first time.

kmdalal:
I tested with autonomy branch OW-658, using the new power fault topic, and as expected, the toggling 0/1 values break that plan.  They happen even with no fault injected.

ow_faults/src/FaultInjector.cpp
lanseafood: removed !

ow_faults/src/FaultInjector.cpp
lanseafood: oh nice shortcut here, will do

kmdalal:
Now, the /faults/power_faults_status value is always 0, regardless of whether any power faults are injected.

Samahu:
> Now, the /faults/power_faults_status value is always 0, regardless of whether any power faults are injected.

I think this PR has a dependency on #112 and also the power profiles for each fault.

Samahu:
@lanseafood when you get the chance can you sync this branch with melodic-devel and resolve build conflicts so we can take another look at it. Thanks!

ow_faults/src/FaultInjector.cpp
```diff
@@ -93,24 +104,62 @@ void FaultInjector::antennaePanFaultCb(const std_msgs::Float64& msg){
...
+}
+
+void FaultInjector::powerTemperatureListener(const std_msgs::Float64& msg)
+{
+  m_temperatureFault = ( msg.data > THERMAL_MAX);
+  publishPowerSystemFault();
+
+}
+
+void FaultInjector::powerSOCListener(const std_msgs::Float64& msg)
```
Samahu, line 152: Please leave a TODO comment to get back to this since we couldn't test this

ow_faults/src/FaultInjector.cpp
Samahu: I don't think you can use const here.

ow_faults/src/FaultInjector.cpp
Samahu: The variable is not used beyond this function .. so I wouldn't recommend converting it to a static (file) variable

Samahu: also I think that the auto type is deduced to const char * which means this is a const value rather than a variable.

ow_faults/src/FaultInjector.cpp
```diff
@@ -26,8 +30,14 @@ FaultInjector::FaultInjector(ros::NodeHandle node_handle)
...
-  m_fault_power_state_of_charge_pub = node_handle.advertise<std_msgs::Float64>("temporary/power_fault/state_of_charge", 10);
-  m_fault_power_temp_pub = node_handle.advertise<std_msgs::Float64>("temporary/power_fault/temp_increase", 10);
+  m_power_soc_sub = node_handle.subscribe("/power_system_node/state_of_charge", 
+                                          1000, 
+                                          &FaultInjector::powerSOCListener, 
+                                          this); 
+  m_power_temperature_sub = node_handle.subscribe("/power_system_node/battery_temperature", 
+                                                  1000, 
+                                                  &FaultInjector::powerTemperatureListener, 
+                                                  this);
```
Samahu, line 40: Maybe it doesn't matter too much but in general I recommend that you move the subscriptions to be after you advertise the topics.

ow_faults/include/ow_faults/FaultInjector.h
```diff
@@ -58,97 +53,122 @@ class FaultInjector
...
+  bool m_armFault;
+  bool m_antFault;
+  bool m_socFault = false;
+  bool m_temperatureFault = false;
+
+  //arm joint faults
+  ow_faults::FaultsConfig m_faults;
+
+  //power vars
+  float m_originalSOC;
```
Samahu, line 162: better name this:
```suggestion
  float m_last_SOC;
```

ow_faults/include/ow_faults/FaultInjector.h
```diff
@@ -58,97 +53,122 @@ class FaultInjector
...
+
+  ////////// vars
+  //system
+  std::bitset<10> systemFaultsBitset;
+
+  //general component faults
+  bool m_armFault;
+  bool m_antFault;
+  bool m_socFault = false;
+  bool m_temperatureFault = false;
```
Samahu, line 156: please follow [OceanWATERS member variable naming guidelines](https://babelfish.arc.nasa.gov/confluence/display/OCEANWATERS/Coding+Style+and+Practices)

ow_faults/include/ow_faults/FaultInjector.h
```diff
@@ -58,97 +53,122 @@ class FaultInjector
...
+  ow_faults::FaultsConfig m_faults;
+
+  //power vars
+  float m_originalSOC;
+
+  // antenna vars
+  std_msgs::Float64 m_realPanMsg;
+  std_msgs::Float64 m_realTiltMsg;
+  float m_faultPanValue;
+  float m_faultTiltValue;
```
Samahu, line 168: Please follow [OceanWATERS's member variable naming guidlines](https://babelfish.arc.nasa.gov/confluence/display/OCEANWATERS/Coding+Style+and+Practices)

ow_faults/msg/SystemFaults.msg
```diff
@@ -9,6 +9,7 @@
 # PtGoalError=64, 
 # PtExecutionError=128, 
 # LanderExecutionError = 256
+# PowerExecutionError = 512
 Header header
 uint64 value
```
Samahu, line 14: Can you add a simple description to the constants specified in the comments. I do understand them personally but I don't think people new to project might as well.

ow_faults/src/FaultInjector.cpp
```diff
@@ -93,24 +104,62 @@ void FaultInjector::antennaePanFaultCb(const std_msgs::Float64& msg){
...
-  } else if (!getTempBool) {
-    powerTemperatureOverloadValue = NAN;
+float FaultInjector::getRandomFloatFromRange( float min_val, float max_val){
+  return min_val + (max_val - min_val) * (rand() / static_cast<float>(RAND_MAX));
+}
+
+void FaultInjector::publishPowerSystemFault(){
+  //power
+  ow_faults::PowerFaults power_faults_msg;
+  ComponentFaults hardwareFault =  ComponentFaults::Hardware;
```
Samahu, line 124: This is something we need to rethink .. as we discussed the m_temperatureFault doesn't qualify to be a **Hardware** fault but I don't see any other better option in **ComponentFault** enum.

ow_faults/include/ow_faults/FaultInjector.h
```diff
@@ -58,97 +53,122 @@ class FaultInjector
...
+
+  // jpl message publishers
+  ros::Publisher m_system_fault_jpl_msg_pub;
+  ros::Publisher m_arm_fault_jpl_msg_pub;
+  ros::Publisher m_power_fault_jpl_msg_pub;
+  ros::Publisher m_antennae_fault_jpl_msg_pub;
+
+  ////////// vars
+  //system
+  std::bitset<10> systemFaultsBitset;
```
Samahu, line 150: please follow [OceanWATERS member variable naming guidelines](https://babelfish.arc.nasa.gov/confluence/display/OCEANWATERS/Coding+Style+and+Practices)

ow_faults/src/FaultInjector.cpp
Samahu: also the variable doesn't seem of any use .. just pass it the enum value to the call.

ow_faults/src/FaultInjector.cpp
```diff
@@ -93,24 +104,62 @@ void FaultInjector::antennaePanFaultCb(const std_msgs::Float64& msg){
...
+  ow_faults::SystemFaults system_faults_msg;
+
+  systemFaultsBitmask |= systemFaultsBitset;
+  //update if fault
+  if (m_temperatureFault || m_socFault) {
+    //system
+    systemFaultsBitmask |= isPowerSystemFault;
+    systemFaultsBitset = systemFaultsBitmask;
+    //power
+    setComponentFaultsMessage(power_faults_msg, hardwareFault);
```
Samahu, line 136:
```suggestion
    setComponentFaultsMessage(power_faults_msg, ComponentFaults::Hardware);
```

Samahu, line 136: To me the **systemFaultsBitmask** seems completely necessary and the code can be reduced to 
```suggestion
  ow_faults::SystemFaults system_faults_msg;

  systemFaultsBitmask |= systemFaultsBitset;
  //update if fault
  if (m_temperatureFault || m_socFault) {
    //system
    systemFaultsBitset |= isPowerSystemFault;
    //power
    setComponentFaultsMessage(power_faults_msg, hardwareFault);
    }
    setBitsetFaultsMessage(system_faults_msg, systemFaultsBitset);
```

Can you explain why the above reduced code wouldn't work?

ow_faults/src/FaultInjector.cpp
```diff
@@ -93,24 +104,62 @@ void FaultInjector::antennaePanFaultCb(const std_msgs::Float64& msg){
...
+{
+  m_temperatureFault = ( msg.data > THERMAL_MAX);
+  publishPowerSystemFault();
+
+}
+
+void FaultInjector::powerSOCListener(const std_msgs::Float64& msg)
+{
+  float newSOC = msg.data;
+  if (isnan(m_originalSOC)){
```
Samahu, line 155: to make use of the isnan function you need to initialize m_originalSOC - which should be m_last_SOC -  to **std::numeric_limits<float>::quiet_NaN()**

Samahu (changes requested):
The functionality works for me, however, I do have some code change requests.

Also, there is an issue with the consistency of publish rate for fault topics but I am willing to bypass for now and address that in a separate ticket later on.

ow_faults/src/FaultInjector.cpp
kmdalal: Good point that 'const char*' is a better type.  The reason I suggested file variable is so that a class initializer list could be used in the constructor.  That's evaluated before the function body.

ow_faults/src/FaultInjector.cpp
kmdalal: Right, it's non-const, but it seems that reference (&) is required to do the right thing, otherwise this function acts on a copy of the node handle.   This is no longer a minor comment :)

ow_faults/src/FaultInjector.cpp
lanseafood: will need to see what Chetan says about the classification of temperature faults then

ow_faults/src/FaultInjector.cpp
lanseafood: i believe this would overwrite any other faults that might be occurring in the system (arm, antenna) as they also get published to the system faults msg.

Samahu: You are performing an OR assignment operation not AND.. so it shouldn't clear flags that are set. You can try it of course if that would make you more comfortable.

ow_faults/src/FaultInjector.cpp
```diff
@@ -93,24 +106,61 @@ void FaultInjector::antennaePanFaultCb(const std_msgs::Float64& msg){
...
+  //publish
+  setBitsetFaultsMessage(system_faults_msg, systemFaultsBitmask);
+  m_system_fault_jpl_msg_pub.publish(system_faults_msg);
+  m_power_fault_jpl_msg_pub.publish(power_faults_msg);
+
+}
+
+void FaultInjector::powerTemperatureListener(const std_msgs::Float64& msg)
+{
+  m_temperature_fault = ( msg.data > THERMAL_MAX);
```
Samahu, line 148:
```suggestion
  m_temperature_fault = (msg.data > THERMAL_MAX);
```
Or remove the parentheses .. they are not needed

ow_faults/include/ow_faults/FaultInjector.h
```diff
@@ -58,97 +53,120 @@ class FaultInjector
...
-	static constexpr std::bitset<10> isPowerSystemFault{	0b10'0000'0000 };
-  
-private:
-  float powerTemperatureOverloadValue;
+	static constexpr std::bitset<10> isPowerSystemFault{    	0b10'0000'0000 };
   
+  //power
+  static constexpr std::bitset<3> islowVoltageError{ 0b001 };
+	static constexpr std::bitset<3> isCapLossError{  	 0b010 };
+	static constexpr std::bitset<3> isThermalError{	   0b100 };
```
Samahu, line 71: The reason these lines appear unaligned on GitHub because you are mixing tabs and spaces. Convert all tabs to spaces.

ow_faults/include/ow_faults/FaultInjector.h
```diff
@@ -39,16 +39,11 @@ class FaultInjector
 
   void faultsConfigCb(ow_faults::FaultsConfig& faults, uint32_t level);
 
-  static constexpr float FAULT_ZERO_TELEMETRY = 0.0;
-
   enum Nominal { None=0 };
```
Samahu, line 42: Not used .. consider dropping it
```suggestion
```

ow_faults/msg/SystemFaults.msg
```diff
@@ -1,5 +1,5 @@
 # System Fault Status Mask
-# Value is the fault classification. See bitmask in FaultInjector.h for implementation of the following:
+# Value is the fault classification. Internally within faults, we use a bitmask to set the different faults such that different faults can occur at the same time (ie. arm and antenna). To translate the bitmask in the message, we translate the bitmask value to an integer to send it to the corresponding topic. See bitmask in FaultInjector.h for implementation of the following:
```
Samahu, line 2: stick to a max of 120 characters per line.

ow_faults/msg/SystemFaults.msg
```diff
@@ -9,6 +9,7 @@
 # PtGoalError=64, 
 # PtExecutionError=128, 
 # LanderExecutionError = 256
```
Samahu, line 11: I would expect to see a comma at the end of this line. Either maintain the commas or remove them

ow_faults/src/FaultInjector.cpp
```diff
@@ -93,24 +106,61 @@ void FaultInjector::antennaePanFaultCb(const std_msgs::Float64& msg){
...
+  //update if fault
+  if (m_temperature_fault || m_soc_fault) {
+    //system
+    systemFaultsBitmask |= isPowerSystemFault;
+    m_system_faults_bitset = systemFaultsBitmask;
+    //power
+    setComponentFaultsMessage(power_faults_msg, ComponentFaults::Hardware);
   }
+  //publish
+  setBitsetFaultsMessage(system_faults_msg, systemFaultsBitmask);
```
Samahu, line 140: I still think it can be safely reduced to:
```suggestion
  ow_faults::PowerFaults power_faults_msg;
  ow_faults::SystemFaults system_faults_msg;

  //update if fault
  if (m_temperature_fault || m_soc_fault) {
    //system
    m_system_faults_bitset |= isPowerSystemFault;
    //power
    setComponentFaultsMessage(power_faults_msg, ComponentFaults::Hardware);
  }
  //publish
  setBitsetFaultsMessage(system_faults_msg, m_system_faults_bitset);
```

Samahu (approved):
Still had few follow up comments but generally looking good

ow_faults/msg/SystemFaults.msg
Samahu: Note this is the line before last

ow_faults/src/FaultInjector.cpp
lanseafood: why does it have   systemFaultsBitmask |= systemFaultsBitset;
 but never use the bitmask var in the loop? Seems like we would only need the systemFaultsBitset |= isPowerSystemFault line?

Samahu: Yes, the line `systemFaultsBitmask |= systemFaultsBitset;` is not required .. I had fixed it in my repeated comment on the same issue.

ow_faults/src/FaultInjector.cpp
lanseafood: ah yes the version without the first declaration makes more sense

lanseafood:
@kmdalal  ready for review

kmdalal:
I'm satisfied with the code (don't care if the minor comments are not addressed).  I'm fine with integrating the power profiles after this ticket, e.g. as a hotfix to the release.  But I would like to solve the problem of the toggling power fault detection in ow_autonomy, which persists:

```
[ WARN]: Fault in SYSTEM: POWER_EXECUTION_ERROR
[ WARN]: Resolved fault in SYSTEM: POWER_EXECUTION_ERROR
[ WARN]: Fault in SYSTEM: POWER_EXECUTION_ERROR
[ WARN]: Resolved fault in SYSTEM: POWER_EXECUTION_ERROR
```

Nothing apparent from code inspection.  I'll continue debugging this afternoon.

kmdalal:
The problem is that the sequence [0, 512, 512]  (in some order) gets published indefinitely on /faults/system_faults_status once the battery temperature and/or state of charge cross their thresholds.   Autonomy interprets this as a power execution error appearing and disappearing.  

I'll take a closer look at the publishing side, but at the moment I'm not sure what the best solution is...

ow_faults/include/ow_faults/FaultInjector.h
```diff
@@ -58,97 +51,120 @@ class FaultInjector
...
+  ros::Publisher m_power_fault_jpl_msg_pub;
+  ros::Publisher m_antennae_fault_jpl_msg_pub;
+
+  ////////// vars
+  //system
+  std::bitset<10> m_system_faults_bitset;
+
+  //general component faults
+  bool m_arm_fault;
+  bool m_ant_fault;
```
kmdalal, line 152: Why not initialize these two to false, like the next two?

kmdalal:
The root problem is that power fault status and antenna/arm fault status are published independently to /faults/system_fault_status.   The code should be refactored so that this topic is published to in just one place, where all 3 fault components are known and the bit vector is set appropriately.   This doesn't look hard to do.  @lanseafood I can take a stab at it if you don't have time today.    Hoping we can get this branch merged today or tomorrow.

kmdalal:
I am testing a simple fix -- no refactoring required.  If it works I'll push a child branch for you to try.

kmdalal (approved):

PR #99, opened 2021-03-11
OCEANWATER 607  Pan/Tilt Faults Halt Antennae Movement 

lanseafood:
## Linked Issues:
| Jira Ticket 🎟️   | [Oceanwater-607](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-607) |
| ----------- | ----------- |
| Github :octocat:  | ow_autonomy [PR](https://github.com/nasa/ow_autonomy/pull/29)  |

IMPORTANT! To run this, you will need to also pull the ow_autonomy [pr](https://github.com/nasa/ow_autonomy/pull/29) of the same name. 

## Summary of Changes
* antennae faults now cause antennae movement to be halted depending on pan or tilt
* remapping pan and tilt topics from rviz and autonomy nodes
* set original topics to /_original
* 'NEW' topics:
** /ant_pan_position_controller/command
** /ant_tilt_position_controller/command

## Test
TEST REMAPPING
* roslaunch any launch file
** to see autonomy remapping, you will need to also roslaunch an autonomy file
* go to rqt -> plugins -> node graph   (select view all)
or 
* roslaunch any launch file
** to see autonomy remapping, you will need to also roslaunch an autonomy file
* rostopic info either to see faults is now the sole publisher
* ** /ant_pan_position_controller/command
** /ant_tilt_position_controller/command
* rostopic info either to see faults now is a subscriber to this
** /_original/ant_pan_position_controller/command
** /_original/ant_tilt_position_controller/command

 TEST RVIZ
* roslaunch any launch file
* in a new terminal `rostopic echo /ant_pan_position_controller/command` or `rostopic echo /ant_tilt_position_controller/command`
* go to rqt and from Control and Viz tab update value to any value for 'command' for either pan or tilt
* go to rqt dynamic reconfigure and select either pan or tilt fault (respective to the topic you're observing)
** click and unclick the fault in rqt to see the output from the 'echo' update.

 TEST AUTONOMY
* roslaunch any launch file
* in a new terminal `rostopic echo /ant_pan_position_controller/command` or `rostopic echo /ant_tilt_position_controller/command`
* roslaunch any autonomy , I recommend demo since it includes antenna movement.
* go to rqt dynamic reconfigure and select either pan or tilt FAULT (respective to the topic you're observing)
** click and unclick the fault in rqt to see the output from the 'echo' update.

### DIFF
--- a/ow_faults/include/ow_faults/FaultInjector.h
+++ b/ow_faults/include/ow_faults/FaultInjector.h
@@ -79,7 +79,7 @@ private:
   // Antennae functions
   void antennaePanFaultCb(const std_msgs::Float64& msg);
   void antennaeTiltFaultCb(const std_msgs::Float64& msg);
-  void publishAntennaeFaults(const std_msgs::Float64& msg, bool encoder, bool torque, float& m_faultValue, ros::Publisher m_publisher);
+  void publishAntennaeFaults(const std_msgs::Float64& msg, bool encoder, bool torque, float& m_faultValue, ros::Publisher& m_publisher);
 
   // Output /faults/joint_states, a modified version of /joint_states, injecting
   // simple message faults that don't need to be simulated at their source.
--- a/ow_faults/include/ow_faults/FaultInjector.h
+++ b/ow_faults/include/ow_faults/FaultInjector.h
@@ -68,8 +68,18 @@ public:
 private:
   float powerTemperatureOverloadValue;
   
+
+
+
+  float faultPanValue;
+  float faultTiltValue;
+
   // renders new temperature when thermal power fault is re-triggered
   void setPowerTemperatureFaultValue(bool getTempBool);
+
+  // Antennae functions
+  void antennaePanFaultCb(const std_msgs::Float64& msg);
+  void antennaeTiltFaultCb(const std_msgs::Float64& msg);
 
   // Output /faults/joint_states, a modified version of /joint_states, injecting
   // simple message faults that don't need to be simulated at their source.
@@ -109,6 +119,12 @@ private:
   ros::Publisher m_arm_fault_status_pub;
   ros::Publisher m_power_fault_status_pub;
   ros::Publisher m_antennae_fault_status_pub;
+
+  //antenna pub and subs
+  ros::Subscriber m_fault_ant_pan_sub;
+  ros::Subscriber m_fault_ant_tilt_sub;
+  ros::Publisher m_fault_ant_pan_pub;
+  ros::Publisher m_fault_ant_tilt_pub;
 
   // Map ow_lander::joint_t enum values to indices in JointState messages
   std::vector<unsigned int> m_joint_state_indices;
--- a/ow_faults/include/ow_faults/FaultInjector.h
+++ b/ow_faults/include/ow_faults/FaultInjector.h
@@ -70,9 +70,8 @@ private:
   
 
 
-
-  float faultPanValue;
-  float faultTiltValue;
+  float m_faultPanValue;
+  float m_faultTiltValue;
 
   // renders new temperature when thermal power fault is re-triggered
   void setPowerTemperatureFaultValue(bool getTempBool);
@@ -80,6 +79,7 @@ private:
   // Antennae functions
   void antennaePanFaultCb(const std_msgs::Float64& msg);
   void antennaeTiltFaultCb(const std_msgs::Float64& msg);
+  void publishAntennaeFaults(const std_msgs::Float64& msg, bool encoder, bool torque, float& m_faultValue, ros::Publisher m_publisher);
 
   // Output /faults/joint_states, a modified version of /joint_states, injecting
   // simple message faults that don't need to be simulated at their source.
--- a/ow_faults/src/FaultInjector.cpp
+++ b/ow_faults/src/FaultInjector.cpp
@@ -87,16 +87,11 @@ void FaultInjector::setFaultsMessage(ow_faults::PTFaults& msg, ComponentFaults v
 void FaultInjector::publishAntennaeFaults(const std_msgs::Float64& msg, bool encoder, bool torque, float& m_faultValue, ros::Publisher m_publisher){
   std_msgs::Float64 out_msg;
 
-  if (encoder || torque) {
-    if (isnan(m_faultValue)){
-      m_faultValue = msg.data;
-    }
-    out_msg.data = m_faultValue;
-    m_publisher.publish(out_msg);
-  } else {
-    m_publisher.publish(msg);
+  if (!(encoder || torque)) {
     m_faultValue = msg.data;
   }
+  out_msg.data = m_faultValue;
+  m_publisher.publish(out_msg);
 }
 
 void FaultInjector::antennaePanFaultCb(const std_msgs::Float64& msg){
--- a/ow_faults/src/FaultInjector.cpp
+++ b/ow_faults/src/FaultInjector.cpp
@@ -22,6 +22,12 @@ FaultInjector::FaultInjector(ros::NodeHandle node_handle)
   m_fault_power_state_of_charge_pub = node_handle.advertise<std_msgs::Float64>("temporary/power_fault/state_of_charge", 10);
   m_fault_power_temp_pub = node_handle.advertise<std_msgs::Float64>("temporary/power_fault/temp_increase", 10);
 
+  //antenna fault publishers and subs
+  m_fault_ant_pan_sub = node_handle.subscribe("/_original/ant_pan_position_controller/command", 3, &FaultInjector::antennaePanFaultCb, this);
+  m_fault_ant_tilt_sub = node_handle.subscribe("/_original/ant_tilt_position_controller/command", 3, &FaultInjector::antennaeTiltFaultCb, this);
+  m_fault_ant_pan_pub = node_handle.advertise<std_msgs::Float64>("/ant_pan_position_controller/command", 10);
+  m_fault_ant_tilt_pub = node_handle.advertise<std_msgs::Float64>("/ant_tilt_position_controller/command", 10);
+
   // topic for system fault messages, see Faults.msg
   m_fault_status_pub = node_handle.advertise<ow_faults::SystemFaults>("/system_faults_status", 10); 
   // topic for arm fault status, see ArmFaults.msg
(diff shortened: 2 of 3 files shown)


ow_faults/src/FaultInjector.cpp
```diff
@@ -75,6 +81,35 @@ void FaultInjector::setFaultsMessage(ow_faults::PTFaults& msg, ComponentFaults v
...
+    pan_msg.data = faultPanValue;
+    m_fault_ant_pan_pub.publish(pan_msg);
+  } else {
+    //turning off fault, should revert to most recent msg val
+    m_fault_ant_pan_pub.publish(msg);
+    faultPanValue = msg.data;
+  }
+}
+
+void FaultInjector::antennaeTiltFaultCb(const std_msgs::Float64& msg){
```
lanseafood, line 99: I realize that this is functionally the same as the antennaePanFaultCb(). The difference is the we are checking either the pan or tilt fault selection/variable. Is there a way in a callback to find out which topic it was subscribed to? If so I can condense these functions.

lanseafood: I tried to look into boost::bind for the subscriber, but running into trouble. I'm seeing a lot of references initialize subscribers outside of the class in a main, but for our purposes, the subscriber is initialized in the class initialization. Would love any pointers.

ow_faults/include/ow_faults/FaultInjector.h
```diff
@@ -68,9 +68,19 @@ class FaultInjector
 private:
   float powerTemperatureOverloadValue;
   
+
+
+
+  float faultPanValue;
+  float faultTiltValue;
```
kmdalal, line 75: These variables need to be prefixed with "m_" per our coding standard for member variables.  I was confused seeing them in the .cpp file, not knowing where they were defined.

ow_faults/src/FaultInjector.cpp
kmdalal: The only modern technique I know to refactor these functions into one is a template function parameterized over the classes involved, and reference arguments (&) for the float-valued member variables that are being set.   The resulting signature and calls are a bit longer, but I think this should be done.  (The old-school C technique is a macro, which I don't recommend).

ow_faults/src/FaultInjector.cpp
```diff
@@ -22,6 +22,12 @@ FaultInjector::FaultInjector(ros::NodeHandle node_handle)
   m_fault_power_state_of_charge_pub = node_handle.advertise<std_msgs::Float64>("temporary/power_fault/state_of_charge", 10);
   m_fault_power_temp_pub = node_handle.advertise<std_msgs::Float64>("temporary/power_fault/temp_increase", 10);
 
+  //antenna fault publishers and subs
+  m_fault_ant_pan_sub = node_handle.subscribe("/_original/ant_pan_position_controller/command", 3, &FaultInjector::antennaePanFaultCb, this);
+  m_fault_ant_tilt_sub = node_handle.subscribe("/_original/ant_tilt_position_controller/command", 3, &FaultInjector::antennaeTiltFaultCb, this);
```
kmdalal, line 27: Please shorten these two lines -- they are really too long, and will get longer if you refactor the callback functions.  Shortening the long (>80-90 column) lines before and after would be nice, but don't worry if it's a hassle.  I realize they contain long names.

ow_faults/include/ow_faults/FaultInjector.h
```diff
@@ -68,9 +68,19 @@ class FaultInjector
 private:
   float powerTemperatureOverloadValue;
```
kmdalal, line 70: same as next comment

kmdalal:
I'm done with code comments, as well as testing.  I did some but not all of your described tests, but added my own autonomy-based tests.   I discovered that if I inject a pan fault, the antenna still tilts, and injecting a tilt fault does not stop pans.   I'm assuming this is the intention, and I think it's fine, though I don't know how antennas really work, or what effect each kind of fault should have.   We can wait for user feedback before changing anything.    I'm ready to approve this after the requested code changes.

ow_faults/src/FaultInjector.cpp
Samahu: You can factor out the shared code into a private method and then invoke the shared private method from each call back. for example:
```suggestion
void FaultInjector::private_shared_method(const std_msgs::Float64& msg, bool flag1, bool flag2, float& m_value, Publisher m_publisher){
std_msgs::Float64 out_msg;
  if (flag1 || flag2) {
    if (isnan(m_value)){
      m_value = msg.data;
    }
    out_msg.data = m_value;
    m_publisher.publish(out_msg);
  } else {
    m_publisher.publish(msg);
    m_value = msg.data;
  }
}
```
Then each callback would invoke the `private_shared_method`

kmdalal: I like this, and using the base class Publisher as parameter is much better than template.

kmdalal:
I forgot to mention that TestAntennaCamera.plx, which I modified on the ow_autonomy version of this branch, is a better test for this feature than Demo.plx.

ow_faults/src/FaultInjector.cpp
lanseafood: updated! TY for the suggestions!

ow_faults/src/FaultInjector.cpp
lanseafood: I dont think the names could be shortened without losing info unless we assume people would know pan and tilt only applies to the antenna?

ow_faults/include/ow_faults/FaultInjector.h
lanseafood: this is the power work, so i'll include it in the other PR

lanseafood:
@kmdalal  updated per your suggestions!

ow_faults/src/FaultInjector.cpp
kmdalal: Agreed, though I was just requesting shorting the _lines_ if feasible, e.g. line breaks between tokens or separators.

Samahu (changes requested):
@lanseafood the last refactoring you made created a coupling between the pan and tilt faults. i.e. if we inject a fault to **(pan or tilt)**, then fault is actually imposed on both joints **(pan and tilt)**.

Make sure to remove any coupling between the two faults.

lanseafood:
> @lanseafood the last refactoring you made created a coupling between the pan and tilt faults. i.e. if we inject a fault to **(pan or tilt)**, then fault is actually imposed on both joints **(pan and tilt)**.
> 
> Make sure to remove any coupling between the two faults.

Hey Ussama, that was actually the point. After speaking with Anna, she confirmed a pan fault will also impact the tilt, resulting in a full stop of the antennae movements.

Samahu:
> > @lanseafood the last refactoring you made created a coupling between the pan and tilt faults. i.e. if we inject a fault to **(pan or tilt)**, then fault is actually imposed on both joints **(pan and tilt)**.
> > Make sure to remove any coupling between the two faults.
> 
> Hey Ussama, that was actually the point. After speaking with Anna, she confirmed a pan fault will also impact the tilt, resulting in a full stop of the antennae movements.

:open_mouth:  :thinking: if this is the case why not merge the two faults into one **pan_tilt_fault**??

lanseafood:
> > > @lanseafood the last refactoring you made created a coupling between the pan and tilt faults. i.e. if we inject a fault to **(pan or tilt)**, then fault is actually imposed on both joints **(pan and tilt)**.
> > > Make sure to remove any coupling between the two faults.
> > 
> > 
> > Hey Ussama, that was actually the point. After speaking with Anna, she confirmed a pan fault will also impact the tilt, resulting in a full stop of the antennae movements.
> 
>   if this is the case why not merge the two faults into one **pan_tilt_fault**??

Because tilt and pan are still separate topics so we would still need two subscribers to make sure we're grabbing the original message content. Based on that we still need to store the respective messages to the respective part (pan/tilt). You can see the function both subscribers call is pretty agnostic of where the fault came from.

ow_faults/src/FaultInjector.cpp
lanseafood: I see, yea in that case I can break apart the lines to make it easier to read then

Samahu (changes requested):
Based on your comments the coupling between pan and tilt faults was intentional i.e. failure in one of the encoder causes failure to the other. This is I can understand if that's actually the case for the physical lander. Nonetheless, the refactored code still have resulted in an avoidable coupling between the two pan and tilt commanding topics. A message sent to one of the topics results in two commands being sent. This shouldn't be the case and is completely avoidable.

Samahu:
@lanseafood even if you address the last comment I will hold off this PR until we discuss the two items:
1. We are trying to mimic the behavior suggested by JPL with regard to the antenna pan and tilt but as far as I recall they haven't really modeled the antenna
2. In your PR you faults for force torque sensor for each of the pan and tilt joints.
  * We don't currently have an F/T sensor for these joints, do we need to? what purpose do they serve?

lanseafood:
> Based on your comments the coupling between pan and tilt faults was intentional i.e. failure in one of the encoder causes failure to the other. This is I can understand if that's actually the case for the physical lander. Nonetheless, the refactored code still have resulted in an avoidable coupling between the two pan and tilt commanding topics. A message sent to one of the topics results in two commands being sent. This shouldn't be the case and is completely avoidable.

Actually I disagree here because if the antennae is not supposed to move at all once there's an error in 1 part (like pan), then the antennae should be halted for tilt by sending the message to stop current movement immediately rather than wait for the next loop of tilt to happen.

lanseafood:
> @lanseafood even if you address the last comment I will hold off this PR until we discuss the two following items:
> 
> 1. We are trying to mimic the behavior suggested by JPL with regard to the antenna pan and tilt but as far as I recall they haven't really modeled the antenna
> 2. In your PR you faults for force torque sensor for each of the pan and tilt joints.
>    
>    * We don't currently have an F/T sensor for these joints, do we need to? what purpose do they serve?

1) Even if they haven't modeled the antenna, I would still consider them experts and consider that as the ground truth for this development. 
2) antenna force and torque were initially set there by Terry. I just left it as it was initially designed there.

Samahu:
> > Based on your comments the coupling between pan and tilt faults was intentional i.e. failure in one of the encoder causes failure to the other. This is I can understand if that's actually the case for the physical lander. Nonetheless, the refactored code still have resulted in an avoidable coupling between the two pan and tilt commanding topics. A message sent to one of the topics results in two commands being sent. This shouldn't be the case and is completely avoidable.
> 
> Actually I disagree here because if the antennae is not supposed to move at all once there's an error in 1 part (like pan), then the antennae should be halted for tilt by sending the message to stop current movement immediately rather than wait for the next loop of tilt to happen.

I see your point but the code still has a problem in which the faults node double the counts each command is sent.

ow_faults/src/FaultInjector.cpp
```diff
@@ -75,6 +84,35 @@ void FaultInjector::setFaultsMessage(ow_faults::PTFaults& msg, ComponentFaults v
...
+  if (encoder || torque) {
+    if (isnan(m_faultValue)){
+      m_faultValue = msg.data;
+    }
+    out_msg.data = m_faultValue;
+    m_publisher.publish(out_msg);
+  } else {
+    m_publisher.publish(msg);
+    m_faultValue = msg.data;
+  }
```
Samahu, line 99: Simplify
```suggestion
  if (!(encoder || torque)) {
    m_faultValue = msg.data;
  }
  out_msg.data = m_faultValue;
  m_publisher.publish(out_msg);
```

The `isnan` is never is actually executed unless your code can actually starts with faults enabled on init

Samahu (changes requested):
Should you include `m_faults.ant_pan_torque_sensor` and `m_faults.ant_tilt_torque_sensor_failure` in the evaluation of introducing an antenna halt movement? According to @mogumbo the intention of this fault is manipulate the effort value for the respective joints. which is something already being done as part of `FaultInjector::jointStateCb`

ow_faults/src/FaultInjector.cpp
```diff
@@ -22,13 +22,22 @@ FaultInjector::FaultInjector(ros::NodeHandle node_handle)
...
+                                              &FaultInjector::antennaePanFaultCb, 
+                                              this);
+  m_fault_ant_tilt_sub = node_handle.subscribe("/_original/ant_tilt_position_controller/command", 
+                                              3, 
+                                              &FaultInjector::antennaeTiltFaultCb, 
+                                              this);
+  m_fault_ant_pan_pub = node_handle.advertise<std_msgs::Float64>("/ant_pan_position_controller/command", 10);
+  m_fault_ant_tilt_pub = node_handle.advertise<std_msgs::Float64>("/ant_tilt_position_controller/command", 10);
+
+  // topic for system fault messages, see Faults.msg, Amrms.msg, Power.msg, PTFaults.msg
```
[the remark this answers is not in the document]
lanseafood: typo

Samahu (approved):

PR #102, opened 2021-03-17
OCEANWATER-613_Add_Temperature_as_a_ROS_Topic

jfugate1:
## Linked Issues:
| Jira Ticket 🎟️   | [Oceanwater-613](https://babelfish.arc.nasa.gov/jira/secure/RapidBoard.jspa?rapidView=168&projectKey=OCEANWATER&view=detail&selectedIssue=OCEANWATER-613) |
| DoD: One of the output parameters of the GSAP prognostics factory is temperature. It needs to be converted to a ROS topic that can be subscribed to by users for fault injection. |
| EPIC ⚡| [Oceanwater-84](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-84) |
| Github :octocat:  | # |

## Summary of Changes
* Identified and helped fix a GSAP issue that previously made temperature inaccessible
* Identified and helped fix a GSAP bug that caused an under-prediction of temperature values 
* Added Temperature calculation and extraction from state matrix
* Created rostopic to publish value to power system node
* Updated README to reflect recent changes to power system node functionality

## Test
* Pull updates to OW-1.0 tagged branch of GSAP
* Recompile GSAP (1. navigate to build folder 2. cmake .. 3. make)
* rebuild the ow_power_system
* Launch OW
* Subscribe to power_system_node/battery_temperature rostopic
* Verify the temperature hovers around a constant value of approximately 20 deg. C.

### DIFF
--- a/ow_power_system/src/power_system_node.cpp
+++ b/ow_power_system/src/power_system_node.cpp
@@ -80,11 +80,12 @@ int main(int argc, char* argv[]) {
   
   ros::Rate rate(power_update_rate);
   //individual soc_msg to be published by SOC_pub
-  std_msgs::Float64 soc_msg;
-  soc_msg.data = 0.0;
   std_msgs::Int16 rul_msg;
   rul_msg.data = 0;  // immediately set to a good default
+  std_msgs::Float64 soc_msg;
+  soc_msg.data = 0.0;
   std_msgs::Float64 tempbat_msg;
+  tempbat_msg.data = 0.0;
   ROS_INFO ("Power system node running");
 
   while (ros::ok()) {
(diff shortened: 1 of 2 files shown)


ow_power_system/src/power_system_node.cpp
```diff
@@ -124,26 +125,23 @@ int main(int argc, char* argv[]) {
...
-      //auto output = model.outputEqn(medianSystemState);
-      //double temperature = output[1];
-      //tempbat_msg.data = temperature;
+        // Temperature Code
+        auto stateSamples = eod_event.getSystemState()[0];
+        std::vector<double> state;
+        for (auto sample : stateSamples) {
+          state.push_back(sample[0]);
+        }
+        auto& model = ((ModelBasedPrognoser *) prognoser.get())->getModel();
```
Samahu, line 135: Better to use dynamic_cast here
```suggestion
        const auto& model = dynamic_cast<ModelBasedPrognoser*>(prognoser.get())->getModel();
```

ow_power_system/src/power_system_node.cpp
```diff
@@ -124,26 +125,23 @@ int main(int argc, char* argv[]) {
...
+        auto z = model.outputEqn(now_s.count(), (PrognosticsModel::state_type) state);
+        double temperature = z[1];
```
Samahu, line 137: I understand it returns a vector of values .. Are all of these values represent temperature and why do you specifically use index 1?

Samahu (commented):
A comment and a question

ow_power_system/src/power_system_node.cpp
lanseafood: agree with Ussama here, the 1 should be parameterized if possible

lanseafood:
Just commenting regarding temperature value, do we expect it to always hover around 20 or is that still a temp value until we get real readings of the termp.

ow_power_system/src/power_system_node.cpp
jfugate1: Yes, I will change this now.

jfugate1:
> Just commenting regarding temperature value, do we expect it to always hover around 20 or is that still a temp value until we get real readings of the termp.

@lanseafood that is just a temporary thing until Chetan gets a more versatile temperature model integrated into GSAP and/or I integrate realistic input parameters.

ow_power_system/src/power_system_node.cpp
jfugate1: Resolved

Samahu (approved):

PR #107, opened 2021-03-30
Oceanwater 630 rename ft faults

lanseafood:
## Linked Issues:
| Jira Ticket 🎟️   | [Oceanwater-624](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-630) |
| ----------- | ----------- |
| EPIC ⚡| [Oceanwater-551](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-551) |
| Github :octocat:  | https://github.com/nasa/ow_autonomy/pull/30 |

*** requires ow_autonomy to be updated to the linked PR

## Summary of Changes
*renaming force torque to effort

## Test
* roslaunch any
* ropstopic echo system_faults_status
* select any faults from rqt
* see output still in rostopic.

PR #109, opened 2021-04-01
Oceanwater 618 model potential failures for the ft sensor

Samahu:
## Linked Issues:
| EPIC ⚡| [OCEANWATER-614 / Implement a force torque sensor for the lander arm ](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-614) |
| Jira Ticket 🎟️   | [OCEANWATER-618 / Model potential faults or failures for the ft sensor ](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-618) |
| Github :octocat:  | N/A |

## Summary of Changes
* Remap `/ft_sensor_dist_pitch` topic to `/_original/ft_sensor_dist_pitch`
* Re-route the topic through the FaultsInjector node
* Model three types of faults for the ft_sensor
  - zero signal
  - signal bias
  - increase signal noise
* Grouped ft_sensor faults under one config with master enable flag
image
* Introduced a random number generator to faults module
* Use normal distribution to simulate increase noise for the sensor

| Fault | Signal |
| Zero Signal |image|
| Signal Bias |image|
| Increased Signal Noise |image|

## Test
1. Launch the simulation
```bash
roslaunch ow (atacma_y1a.launch | europa_terminator_workspace.launch)
```
2. listen to `/ft_sensor_dist_pitch` topic in a another terminal, verify the signal is coming as expected
```bash
rostopic echo /ft_sensor_dist_pitch
```
3. Use the dyanmic_reconfigure ui to apply any of the three faults (zero signal, signal bias, increase signal noise)
image
>Note: you can apply them individually or in combination

4. Verify that the master enable/disable flag enable or suppresses the three faults togther

Suggestions or code improvements are welcome!

### DIFF
--- a/ow_faults/include/ow_faults/FaultInjector.h
+++ b/ow_faults/include/ow_faults/FaultInjector.h
@@ -123,7 +123,7 @@ private:
   // Map ow_lander::joint_t enum values to indices in JointState messages
   std::vector<unsigned int> m_joint_state_indices;
 
-  std::mt19937 m_randomGenerator; // Utilize a Mersenne Twister pesduo random generation
+  std::mt19937 m_random_generator; // Utilize a Mersenne Twister pesduo random generation
 };
 
 #endif
--- a/ow_faults/include/ow_faults/FaultInjector.h
+++ b/ow_faults/include/ow_faults/FaultInjector.h
@@ -18,6 +18,7 @@
 #include "ow_faults/PTFaults.h"
 #include <ow_lander/lander_joints.h>
 #include <sensor_msgs/JointState.h>
+#include <geometry_msgs/WrenchStamped.h>
 #include <unordered_map>
 
 
@@ -75,6 +76,10 @@ private:
   // simple message faults that don't need to be simulated at their source.
   void jointStateCb(const sensor_msgs::JointStateConstPtr& msg);
 
+  // Output /faults/joint_states, a modified version of /joint_states, injecting
+  // simple message faults that don't need to be simulated at their source.
+  void distPitchFtSensorCb(const geometry_msgs::WrenchStamped& msg);
+
   //Setting the correct values for faults messages via function overloading
   //System Faults
   void setFaultsMessage(ow_faults::SystemFaults& msg, std::bitset<10> systemFaultsBitmask);
@@ -99,6 +104,10 @@ private:
   // arm faults
   ros::Subscriber m_joint_state_sub;
   ros::Publisher m_joint_state_pub;
+
+  // ft sensor
+  ros::Subscriber m_dist_pitch_ft_sensor_sub;
+  ros::Publisher m_dist_pitch_ft_sensor_pub;
 
   // temporary placeholder publishers until power feautre is finished
   ros::Publisher m_fault_power_state_of_charge_pub;
--- a/ow_faults/include/ow_faults/FaultInjector.h
+++ b/ow_faults/include/ow_faults/FaultInjector.h
@@ -20,6 +20,7 @@
 #include <sensor_msgs/JointState.h>
 #include <geometry_msgs/WrenchStamped.h>
 #include <unordered_map>
+#include <random>
 
 
 // This class injects simple message faults that don't need to be simulated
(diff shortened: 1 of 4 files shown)


ow_faults/include/ow_faults/FaultInjector.h
```diff
@@ -112,6 +122,8 @@ class FaultInjector
 
   // Map ow_lander::joint_t enum values to indices in JointState messages
   std::vector<unsigned int> m_joint_state_indices;
+
+  std::mt19937 m_randomGenerator; // Utilize a Mersenne Twister pesduo random generation
```
mogumbo, line 126: should be m_random_generator

ow_faults/src/FaultInjector.cpp
```diff
@@ -16,7 +16,10 @@ FaultInjector::FaultInjector(ros::NodeHandle node_handle)
 {
   m_joint_state_sub = node_handle.subscribe("/_original/joint_states", 10, &FaultInjector::jointStateCb, this);
   // 'new' joint states topic where fault data is now publishes on
-  m_joint_state_pub = node_handle.advertise<sensor_msgs::JointState>("/joint_states", 10); 
+  m_joint_state_pub = node_handle.advertise<sensor_msgs::JointState>("/joint_states", 10);
+
+  m_dist_pitch_ft_sensor_sub = node_handle.subscribe("/_original/ft_sensor_dist_pitch", 10, &FaultInjector::distPitchFtSensorCb, this);
```
mogumbo, line 21: Since it is used twice, consider putting "ft_sensor_dist_pitch" in a const string to avoid future mix ups. I could have done that with "joint_states" as well.

mogumbo (changes requested):

mogumbo:
@Samahu, did you come to any conclusions on the naming of redundant force torque sensors for the joints? And since we seem to have redundant sensors here, perhaps they should be grouped differently where all sensors for the same joint are grouped together.

Samahu:
> @Samahu, did you come to any conclusions on the naming of redundant force torque sensors for the joints? And since we seem to have redundant sensors here, perhaps they should be grouped differently where all sensors for the same joint are grouped together.

@lanseafood has renamed them from torque_sensor failure to effort failures as part of #107

ow_faults/src/FaultInjector.cpp
Samahu: addressed

ow_faults/include/ow_faults/FaultInjector.h
Samahu: Why is that? I thought we are using camelCase for member variables

Samahu: It seems to be the case .. I will update it

mogumbo: snake_case - https://babelfish.arc.nasa.gov/confluence/display/OCEANWATERS/Coding+Style+and+Practices
This is why I wanted one IRG coding standard all along. There are too many too keep track of.

Samahu: addressed

mogumbo (approved):

ow_faults/include/ow_faults/FaultInjector.h
kmdalal: I have overlooked this from the beginning and thought the convention was m_CamelCase.  This is already pervasive in our code.  I would like to propose that we retain just the "m_" and not dictate the style that follows.  This would be sufficient to identify member variables.

lanseafood:
roslaunch ow (atacma_y1a.launch | europa_terminator_workspace.launch)
bash: syntax error near unexpected token `('

I copy/pasted from the test section, is there something I'm missing here?

lanseafood (approved):
it looks good to me code-wise and works!

kmdalal:
I've tested using autonomy, and verified that GuardedMove fails in the presence of F/T faults (zero-signal).  Are there other arm operations that should fail when injecting F/T faults?

kmdalal (approved):
Nice feature!

Samahu:
> I've tested using autonomy, and verified that GuardedMove fails to find ground in the presence of F/T faults (zero-signal). Are there other arm operations that should fail when injecting F/T faults?

I don't think other arm activities are prone to errors of the F/T sensor. GuardedMove fails because in this branch it uses readings of the F/T sensor to detect the ground. I expect the two other failure types (signal bias and noise) can also cause the GuardedMove to break too at a given threshold (which maybe not covered by set limits).

PR #113, opened 2021-04-20
Oceanwater 653 add the ability to apply additional load on the battery at runtime

Samahu:
## Linked Issues:
| EPIC ⚡| [OCEANWATER-84 / Power System Modelling](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-84) |
| Jira Ticket 🎟️  | [OCEANWATER-653 / Add the ability to apply additional load on the battery at runtime ](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-653) |
| Github :octocat:  | N/A |

## Summary of Changes
* Load a csv file representing a given faulty pattern
* Apply the fault at run time while the fault is activated
  - This is currently implemented as an addition to currently generated estimates

## Test
1) Launch the simulation
```bash
```

2) In another terminal with oceanwaters workspace sourced, monitor the different power system output
```bash
```

3) Enable any of three different power faults from RQT Dynamic Reconfigure
* low_state_of_charge_power_failure
* instantaneous_capacity_loss_power_failure
* thermal_power_failure

> **Note**: Currently there is no difference in the behavior of the three different faults since the three faults are using the same profile. However, enabling multiple faults at the same time would accumulate the total fault value applied to power, voltage and temperature. 

4) Observe that the values of state_of_charge starts to drop significantly

> **Note:** If the total power applied to power exceed a certain limit (this happens when combining multiple power faults at the same time), then some bad things happen and the power system node crashes at some point.


lanseafood:
@Samahu I tried to look through this branch but getting errors when launching.

lanseafood:
@Samahu 

I ran the test and saw the error you refer too. The value does seem to drop (just not from the correct values since we don't have that yet I think). I get what you're going for here.

Samahu:
> @Samahu
> 
> I ran the test and saw the error you refer too. The value does seem to drop (just not from the correct values since we don't have that yet I think). I get what you're going for here.

Yes, this will depend on the values that are in the csv files .. we can amend them later to get exactly what we want.

lanseafood (approved):

PR #115, opened 2021-05-06
Remove erroneous publish of system faults message

kmdalal:
## Summary of Changes
*  joint_states_cb was publishing to /faults/system_faults_status without heeding the power fault status.  Some refactoring could remedy this, but I tried simply removing this publish, since the (correct) system fault status is being published by the power monitors.  This has probably affected the publish rate of this telemetry, assuming this is harmless for now.    Perhaps a better solution would be a code refactor.

NOTE: this change just removes line 257 in FaultInjector.cpp.  However, the diff shows many lines changed because my C++ editing mode in Emacs automatically removes extraneous spaces at the end of lines, and there were many of these.

## Test
* Same tests as for the parent branch's PR: https://github.com/nasa/ow_simulator/pull/98.  In particular, let the unrecoverable power fault happen, and verify that a continuous stream of 512 (plus any other arm/antenna faults injected) is published on /faults/system_faults_status.   Before, 512 and 0 kept alternating.

### DIFF
--- a/ow_faults/src/FaultInjector.cpp
+++ b/ow_faults/src/FaultInjector.cpp
@@ -29,36 +29,36 @@ FaultInjector::FaultInjector(ros::NodeHandle& node_handle)
     10, &FaultInjector::distPitchFtSensorCb, this);
   m_dist_pitch_ft_sensor_pub = node_handle.advertise<geometry_msgs::WrenchStamped>(ft_sensor_dist_pitch_str, 10);
 
-  
+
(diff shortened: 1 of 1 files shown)


kmdalal:
@lanseafood and @Samahu, if either of you approves this I think it is merge-able, since the parent branch was already fully approved except for this problem.

lanseafood (approved):
i think this makes sense! Thanks for handling it!

PR #116, opened 2021-05-06
Decouple dependency of PowerNode on joint_states topic +

Samahu:
## Linked Issues:
| EPIC ⚡| [OCEANWATER-84 / Power System Modelling](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-84) |
| Jira Ticket 🎟️  | [OCEANWATER-652 / Remove the strong dependency of PowerSystemNode on joint_states topic ](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-652) |
| Jira Ticket 🎟️  | [OCEANWATER-673 / When power faults are injected the publish rate of power value is increased ](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-673) |
| Github :octocat:  | N/A |

## Summary of Changes
* Decouple the processing of power values from the joint_states callback. This was critical for the following factors:
  - Prognoser takes too much time to process data and thus is not able to keep up
  - This entails that not all power values were actually being processed
  - Also, this change is required for the future when integrated power consumption from other devices (such as the compute board, the antenna, cameras and lander lights, ..)
* Rather than invoking the prognoser on the mean mechanical power value use the accumulated raw power values consumed over a processing cycle
* Set a fixed processing rate for the power (hard coded to 0.5 for now)  

## Test
1) Launch the simulation
```bash
```

2) In another terminal with oceanwaters workspace sourced, monitor the different power system output
```bash
```

3) Also monitor the publish rate of the previous topics
```bash
rostopic hz /power_system_node/state_of_charge
rostopic hz /power_system_node/remaining_useful_life
rostopic hz /power_system_node/battery_temperature
```
the update rate should on the order of  ~0.5 hz

4) Enable one or two at most of the different power faults from RQT Dynamic Reconfigure
* low_state_of_charge_power_failure
* instantaneous_capacity_loss_power_failure
* thermal_power_failure

5) Check that the update rate of power node topics stays about the same and doesn't exceed 0.5

### DIFF
--- a/ow_power_system/src/power_system_node.cpp
+++ b/ow_power_system/src/power_system_node.cpp
@@ -25,35 +25,35 @@ m_power_values(m_moving_average_window, 0)
 
 bool PowerSystemNode::Initialize()
 {
+  if (!loadSystemConfig())
+    return false;
+
+  if (!loadFaultPowerProfiles())
+    return false;
+
+  if (!initPrognoser())
+    return false;
+
+  if (!initTopics())
+    return false;
+
+  return true;
+}
+
+bool PowerSystemNode::loadSystemConfig()
+{
   auto system_config_path = ros::package::getPath("ow_power_system") + "/config/system.cfg";
   auto system_config = ConfigMap(system_config_path);
-
   m_initial_power = system_config.getDouble("initial_power");
   m_initial_voltage = system_config.getDouble("initial_voltage");
   m_initial_temperature = system_config.getDouble("initial_temperature");
-
   m_base_voltage = system_config.getDouble("base_voltage");
   m_voltage_range = system_config.getDouble("voltage_range");
-
   m_min_temperature = system_config.getDouble("min_temperature");
   m_max_temperature = system_config.getDouble("max_temperature");
-
   m_battery_lifetime = system_config.getDouble("battery_lifetime");
-
   m_efficiency = system_config.getDouble("efficiency");
-
   m_temperature_dist = uniform_real_distribution<double>(m_min_temperature, m_max_temperature);
-
-  string path;
-  path = ros::package::getPath("ow_power_system") + "/data/low_state_of_charge_power_failure.csv";
-  m_low_state_of_charge_power_failure_sequence = loadPowerProfile(path);
-
-  path = ros::package::getPath("ow_power_system") + "/data/instantaneous_capacity_loss_power_failure.csv";
-  m_instantaneous_capacity_loss_power_failure_sequence = loadPowerProfile(path);
-
-  path = ros::package::getPath("ow_power_system") + "/data/thermal_power_failure.csv";
-  m_thermal_power_failure_sequence = loadPowerProfile(path);
-
   return true;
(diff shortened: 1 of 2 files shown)


Samahu:
@reviewers: I recommend producing and observing the issue on **melodic-devel** first to understand what the problem was. Then come back to this branch

kmdalal:
In melodic-devel, the initial rate was about .45 for all three topics.   I injected each power fault separately and this did not affect the rate.  I tried each combination of two faults, and only one combination had an effect on the rate: low SOC and instantaneous capacity.  The rate increased to around .51.  I then added the thermal fault, and the rate started to increase quickly getting as high as 7Hz.  I turned off all faults and the rate is very slowly decreasing, down to about 3Hz after several minutes.

I presume I have reproduced the problem.  Now I will try the fix....

ow_power_system/src/power_system_node.cpp
```diff
@@ -106,6 +106,59 @@ vector<map<MessageId, Datum<double>>> PowerSystemNode::loadPowerProfile(const st
...
+  auto prognoser_config_path = ros::package::getPath("ow_power_system") + "/config/prognoser.cfg";
+  ConfigMap prognoser_config(prognoser_config_path);
+
+  // Contruct a new prognoser using the prognoser factory. The prognoser
+  // will automatically construct an appropriate model, observer and predictor
+  // based on the values specified in the config.
+  m_prognoser = PrognoserFactory::instance().Create("ModelBasedPrognoser", prognoser_config);
+
+  if (m_prognoser == nullptr)
+    return false;
```
kmdalal, line 136: Issue a ROS error here, so the user knows what went wrong?

kmdalal (approved):
The test passed, even with all 3 power faults injected, and I also regression-tested some other fault-related behavior.  Code changes look fine.   Good fix!

ow_power_system/src/power_system_node.cpp
Samahu: There is already an error message for that on. Check line 358: `ROS_ERROR("Power system node failed to initialize");`. Of course it is not detailed but it should let the user know there was a problem during initialization. Let me know if this is not sufficient and a more detailed message is needed.

kmdalal: It's fine as it is. then.  I'm guessing it's a rare problem.

Samahu: It didn't happen to me .. I just added this as an extra check

Samahu:
> For me the 3 power topics publish at rates between 0.25 and 0.38 Hz. The rates sometimes change as I select and deselect different power faults.

This means that the Prognoser step method is taking more than 2 seconds on your machine to finish. If we want to ensure that we maintain the same rate across all machines all we have to do is choose lowest publish rate across the board. And if Chetan thinks we should only invoke the Prognoser every minute then the code still holds. All we have to do is set **m_power_node_processing_rate** variable to 1/60

PR #119, opened 2021-05-14
 switching back controllers in grind in the event of fault

anjan1984:
## Linked Issues:
| ----------- | ----------- |
| Github :octocat:  | # |

## Summary of Changes
* Change back controller in the event of fault. 

## Test
* Test 6.4 . Re-do test 6.4 to see if it works.

### DIFF
--- a/ow_lander/scripts/grind_action_server.py
+++ b/ow_lander/scripts/grind_action_server.py
@@ -120,6 +120,10 @@ class GrindActionServer(object):
         else:
             rospy.loginfo('%s: Failed' % self._action_name)
             self._server.set_aborted(self._result)
+            switch_success = self.switch_controllers('arm_controller','grinder_controller')
+            if not switch_success:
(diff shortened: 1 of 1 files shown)


kmdalal:
Test 6.4 still fails, but in different ways:

1. On Unstow, arm moves a little bit.
2. GuardedMove is performed, after arm bounces and makes strange movements.
3. On Grind, arm extends and stops.
4. On DigLinear, arm raises a bit.
5. On DigCircular, arm raises a bit more.  
    [WARN] Inbound TCP/IP connection failed: connection from sender terminated before handshake header recieved...
6. On Deliver, same TCP/IP warning

However, the controller errors/warnings did not appear.

kmdalal (approved):
If this fix was intended just to fix the controller errors in test 6.4, it seems to work.  The other pre-existing errors persist, with somewhat different arm behavior.

PR #122, opened 2021-05-18
fix index out of range error

anjan1984:
## Linked Issues:
| ----------- | ----------- |
| Github :octocat:  | # |

## Summary of Changes
* 
Fix index out of range errors. When Planner does not find any valid plans action should be terminated without errors.  
## Test
* 
Test 6.4 Index out of range errors should not happen.

### DIFF
--- a/ow_lander/scripts/action_dig_circular.py
+++ b/ow_lander/scripts/action_dig_circular.py
@@ -179,7 +179,8 @@ def dig_circular(move_arm, move_limbs, robot, moveit_fk, args):
   if not parallel:
      
     plan_a= move_to_pre_trench_configuration_dig_circ(move_arm, robot, x_start, y_start) 
-    
+    if len(plan_a.joint_trajectory.points) == 0:  # If no plan found, abort
+      return False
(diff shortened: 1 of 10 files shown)


kmdalal (approved):
Tried many of the actions, and did not get the previous error, which would always occur after:
```
```
Now, there is no error after this warning, and test 6.5.2 still passes.

Out of scope for this PR, but can we avoid _planning_ altogether in presence of faults?  Also, the clients take a long time to return even when there is no arm movement.

anjan1984:
> Tried many of the actions, and did not get the previous error, which would always occur after:
> 
> ```
> [ WARN]: Fail: ABORTED: No motion plan found. No execution attempted.
> ```
> 
> Now, there is no error after this warning, and test 6.5.2 still passes.
> 
> Out of scope for this PR, but can we avoid _planning_ altogether in presence of faults? Also, the clients take a long time to return even when there is no arm movement.

We can make a separate JIRA ticket for this and I can look into it

kmdalal:
@lanseafood, I'm not sure if you were planning to review this, but @anjan1984 I think it is safe to merge -- my test was pretty thorough.

PR #125, opened 2021-05-20
Changed python to python2 in several scripts.

mogumbo:
## Linked Issues:
| Jira Ticket 🎟️   | [Oceanwater-678](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-678) |

### DIFF
--- a/ow_lander/scripts/LanderInterface.py
+++ b/ow_lander/scripts/LanderInterface.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 
 # The Notices and Disclaimers for Ocean Worlds Autonomy Testbed for Exploration
 # Research and Simulation can be found in README.md in the root directory of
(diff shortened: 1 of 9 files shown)


kmdalal:
I'm not familiar with the Python conventions.  Is this critical to get into release 8?  Is it safe?  Is something breaking right now?  Is there a test for this change, other than trying all the scripts?   Does it need a test?

mogumbo:
> I'm not familiar with the Python conventions. Is this critical to get into release 8? Is it safe? Is something breaking right now? Is there a test for this change, other than trying all the scripts? Does it need a test?

I ran ReferenceMission1.plx with no problem, so it's probably safe. One user had a problem using "python" in our scripts because his system defaulted to python3. We fixed the problem by making our scripts explicitly "python2." This shouldn't do any damage but might save people trouble in the future.

kmdalal (approved):

Samahu (approved):
consistent with the rest of our python scripts

PR #142, opened 2021-06-22
Oceanwater 720 Formatting with f-string

AliMuhammadOfficial:
## added f-string formatting instead of concatenation and removed unused imports  
* 

## Test
*


Samahu:
@AliMuhammadOfficial We are still using python 2 for the master branch and f-string formatting is not a supported feature in python 2. Thus, we won't be approving this PR.

AliMuhammadOfficial:
Ok thanks!

PR #138, opened 2021-06-15
HOTFIX: Disable velocity limits on all joints

Samahu:
## Linked Issues:
| EPIC ⚡| [OCEANWATER-577 / Support for ROS Noetic](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-577) |
| :----------- | :----------- |
| Jira Ticket 🎟️   | [OCEANWATER-705 / Arm moves very slow under ROS Melodic and Noetic](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-705)) |
| Github :octocat:  | #136  |

## Summary of Changes
**Problem Statement:** After some recent updates made to ROS Moveit (Melodic/Noetic), the lander arm started to move very slow. This issue was probably caused by a feature in MoveIt that OceanWATERS was making use of but it wasn't functional. The feature in question basically gives users the ability to limit the speed at which joints can move. 
  
**Solution:** Disable velocity limits on all joints

## Test
* Launch the simulation
```bash
roslaunch ow atacama_y1a.launch
```
* Run any arm operation
```bash
rosservice call /arm/stow
# or
rosservice call /arm/unstow
# or
python stow_action_client.py
# ...
```
* Arm should execute the operation normally as before

PR #147, opened 2021-06-23
fixing velocity limits in both joint_limit.yamls and lander.xarco file

anjan1984:
## Linked Issues:
| EPIC ⚡| [OCEANWATER-XXX](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-XXX) |
) |
| Jira Ticket 🎟️   | [OCEANWATER-XXX](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-XXX)) |
| Github :octocat:  | # |

## Summary of Changes
*  Updated velocity limits for all joints in both joint_limits.yaml and lander.xacro file 
* After carefully checking all the possible changes, it seems that re-timing of the trajectory is not necessary  as in branch 711. 

## Test
* Please check all the arm movements in both actions and services. 
* Test 2 and Test 9 of our Release 8.

### DIFF
--- a/ow_lander/config/joint_limits.yaml
+++ b/ow_lander/config/joint_limits.yaml
@@ -4,46 +4,46 @@
 joint_limits:
   j_ant_pan:
     has_velocity_limits: true
-    max_velocity: 0.4
+    max_velocity: 3.0
     has_acceleration_limits: false
     max_acceleration: 0
   j_ant_tilt:
     has_velocity_limits: true
-    max_velocity: 0.4
+    max_velocity: 3.0
     has_acceleration_limits: false
     max_acceleration: 0
   j_dist_pitch:
     has_velocity_limits: true
-    max_velocity: 0.2
+    max_velocity: 3.0
     has_acceleration_limits: false
     max_acceleration: 0
   j_grinder:
     has_velocity_limits: true
-    max_velocity: 0.1
+    max_velocity: 3.0
     has_acceleration_limits: false
     max_acceleration: 0
   j_hand_yaw:
     has_velocity_limits: true
-    max_velocity: 0.2
+    max_velocity: 3.0
     has_acceleration_limits: false
     max_acceleration: 0
   j_prox_pitch:
     has_velocity_limits: true
-    max_velocity: 0.2
+    max_velocity: 3.0
     has_acceleration_limits: false
     max_acceleration: 0
   j_scoop_yaw:
     has_velocity_limits: true
-    max_velocity: 0.2
+    max_velocity: 3.0
     has_acceleration_limits: false
     max_acceleration: 0
   j_shou_pitch:
     has_velocity_limits: true
-    max_velocity: 0.2
+    max_velocity: 3.0
     has_acceleration_limits: false
     max_acceleration: 0
   j_shou_yaw:
     has_velocity_limits: true
-    max_velocity: 0.15
+    max_velocity: 3.0
     has_acceleration_limits: false
     max_acceleration: 0
--- a/ow_lander/scripts/action_grind.py
+++ b/ow_lander/scripts/action_grind.py
@@ -90,7 +90,6 @@ def plan_cartesian_path(move_group, wpose, length, alpha, parallel, z_start, cs)
                                waypoints,   # waypoints to follow
                                0.01,        # end effector follow step (meters)
                                0.0)         # jump threshold
-
   return plan, fraction
 
 
--- a/ow_lander/urdf/lander.xacro
+++ b/ow_lander/urdf/lander.xacro
@@ -165,7 +165,7 @@
     <axis
       xyz="0 0 -1" />
     <dynamics damping="50.0" friction="2.6"/>
-    <limit effort="70.9" velocity="0.15" lower="-1.8" upper="1.8"/>
+    <limit effort="70.9" velocity="1.0" lower="-1.8" upper="1.8"/>
   </joint>
   <transmission name="shou_yaw_transmission">
     <type>transmission_interface/SimpleTransmission</type>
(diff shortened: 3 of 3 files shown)


kmdalal:
The Grind in test 2.7 was uneven -- the arm jigged and went up/down.   Revealed in the resulting trench, shown here.
grind1

kmdalal:
The DigLinear in test 2.8 had good movement, but was not aligned with the trench.  See photo here.
dig-linear

kmdalal:
The actions (Release test 9) all looked good!

kmdalal (approved):

kmdalal:
Please make sure there's at least one regression test (using MoveIt 1.0.7).   I used 1.0.8.

AstroStucky:
I am testing on MoveIt 1.0.7...

### Release 8 - Test 9 -  Arm Actions Test Results
All action arm movements seem to move 2-3x faster than I'm used to seeing them move before the OCEANWATER-711 bug. They also all appeared to execute their individual joint movements out of synchronicity with other joint movements, causing the scoop trajectory to be kind of random.

**grind_action_client.py** dragged the grinder along the ground, rather than moving it through it. Seen here 
Screenshot from 2021-06-23 17-10-36

**dig_circular_action_client.py** stopped at this position
Screenshot from 2021-06-23 17-30-15

**dig_linear_action_client.py** did something similar to dig_circular

### Arm Services Test Results
All services were called from the command line and used default arguments.
The arm services all suffered from the same fast motion and unpredictable timing of joint movements as the arm actions did. 

**/arm/guarded_move** tilted the scoop a bit before slamming it rather quickly into the ground

**/arm/grind** seemed to perform the correct grind trajectory, but did so several centimeters above the ground

**/arm/dig_circular** never made contact with the ground either, and performed the dig circular trajectory without having the scoop in the downward pointing position. See how in the screen capture of immediately after dig circular the scoop is off on tool selector joint and not even facing the right direction.
Screenshot from 2021-06-23 17-45-48

**/arm/dig_linear** was even stranger. The scoop was not pointing downwards either, and the scoop's yaw was not aligned in the direction of digging.
Screenshot from 2021-06-23 17-51-35

**/arm/deliver_sample** managed to do something like the arm delivery trajectory (but far too fast and with nonsynchronous joint movements), and then somehow wound up in this position
Screenshot from 2021-06-23 17-53-15
Calling it a second time made it end in a position closer to where I'd expect, so this behavior seems to be random.

It looks as though this PR does not work if the user has MoveIt 1.0.7, but another test with 1.0.7 to verify this would be nice to see.

mogumbo:
Can anyone explain exactly what changed that made this PR necessary? Did MoveIt gain a new behavior, or did it gain a bug?

Why do joint_limits.yaml and lander.xacro have different values for velocity limits? This was also the case before this PR. Are velocity limits in one of the files ignored? Why do velocity limits need to increase so much?

Samahu:
> Can anyone explain exactly what changed that made this PR necessary? Did MoveIt gain a new behavior, or did it gain a bug?
> 
> Why do joint_limits.yaml and lander.xacro have different values for velocity limits? This was also the case before this PR. Are velocity limits in one of the files ignored? Why do velocity limits need to increase so much?

The following comment from `planning_context.launhc` describes the purpose of `joint_limits.yaml`:  
```xml
  <!-- Load updated joint limits (override information from URDF) -->
  <group ns="$(arg robot_description)_planning">
    <rosparam command="load" file="$(find ow_lander)/config/joint_limits.yaml"/>
  </group>
```

This basically sets the speeds that the planner would use when generating the trajectory. What is the net effect if the velocities specified here are higher than the ones in urdf is an open question to me. I think `joint_limits.yaml` has no effect on the simulated arm, it only affects the planner and the velocities of the generated trajectories.

Samahu:
I think I partially know what was the issue with my original hotfix #138.
When **has_velocity_limits** flag is set to true, it forces the planner to generate a trajectory that is linear in nature. When I unchecked that flag it allowed the planner to generate trajectories that has acceleration in them. We have been testing and developing with linear movement for the arm the entire time. We even optimized our PID values around this stetting.

@mogumbo To directly answer your question:
> Why do joint_limits.yaml and lander.xacro have different values for velocity limits? This was also the case before this PR. Are velocity limits in one of the files ignored? Why do velocity limits need to increase so much?

The values in the xacro files are completely overridden by joints_limits.yaml. The velocity values you specify in the xacro don't have much of an impact if you enable _has_velocity_limits_ and set a reasonable max_velocity.

This fact made me think that the issue maybe not in joint_limits itself. Rather, there could be revisions that affected how the PID  values are used in the sim and broke things.

anjan1984:
I did try to tune the PID values with not much success.... One thing that I see different in MoveIt 1.0.8  is https://github.com/ros-planning/moveit/pull/2610... looking both in the planning and the control side of this ...

Samahu:
> I did try to tune the PID values with not much success.... One thing that I see different in MoveIt 1.0.8 is [ros-planning/moveit#2610](https://github.com/ros-planning/moveit/pull/2610)... looking both in the planning and the control side of this ...

I have noticed the moveit  2610 issue but we are not using the `pilz_industrial_motion_planner` so I don't know how it affects our simulation.

Samahu:
> > I did try to tune the PID values with not much success.... One thing that I see different in MoveIt 1.0.8 is [ros-planning/moveit#2610](https://github.com/ros-planning/moveit/pull/2610)... looking both in the planning and the control side of this ...
> 
> I have noticed the moveit 2610 issue but we are not using the `pilz_industrial_motion_planner` so I don't know how could this affected our simulation.

Samahu (changes requested):
Please close this PR without merge in favor of PR #151 which should fix the root cause of the problem.

kmdalal:
Replaced by new approach that was merged.

PR #154, opened 2021-07-01
remove rviz visualization of planned path

anjan1984:
## Linked Issues:
| EPIC ⚡| [OCEANWATER-XXX](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-XXX) |
) |
| Jira Ticket 🎟️   | [OCEANWATER-XXX](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-XXX)) |
| Github :octocat:  | # |

## Summary of Changes
* Removes Rviz Visualization of the plans. (Turns out it was not the fake controller but Rviz was showing the planned path.)

## Test
* Run any Ros action to see no displayed rviz trjactory before movement of Gazebo

### DIFF
--- a/ow_lander/config/moveit.rviz
+++ b/ow_lander/config/moveit.rviz
@@ -183,7 +183,7 @@ Visualization Manager:
         Robot Alpha: 0.5
         Robot Color: 150; 50; 150
         Show Robot Collision: false
-        Show Robot Visual: true
+        Show Robot Visual: false
         Show Trail: false
         State Display Time: 0.05 s
         Trail Step Size: 1


kmdalal:
This works as advertised, but does not address the serious problem of the _delay_ between when the action is activated (goal sent) and the actual arm movement starting.  This can take up to 30 seconds.  Is the planning algorithm really taking that long?  If so, then it's better to visualize the planning like before, because now both Rviz and Gazebo are still and one wonders what's going on.

anjan1984:
Yes .. it will take that time... Might improve a little bit, now  that RRT outputs are not displayed ... Best not to merge this then....

Samahu:
Based on the feedback from @mikedalal this doesn't seem to help improve the execution speed of arm operations. If that's the case then I would suggest closing this PR.

kmdalal:
> Based on the feedback from @mikedalal this PR doesn't seem to help improve the execution speed of arm operations. If that's the case then I would suggest closing it.

This has been agreed on by the team, so closing for now.

PR #159, opened 2021-07-14
Successive Guarded move call failed

anjan1984:
## Linked Issues:
| EPIC ⚡ | [OCEANWATER-XXX](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-XXX) |
| Jira Ticket 🎟️ | [OCEANWATER-668_successive_guarded_move_action](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-XXX) |
| Github :octocat:  | # |

## Summary of Changes
* Successive Guarded move call failed. This PR fixes the issue using estimated plan fraction instead of using distance from ground. Makes it general for all types of guarded move. In services ground detection is used only on the guarded move trajectory and not the pre-guarded move trajectory. Same concept has been used using estimated plan fraction. 

## Test

In separate terminal: 
cd ow_simulator/ow_lander/scripts/

python guarded_move_action_client.py 

repeat this for at least two times to test successive guarded move


kmdalal (approved):
Did many consecutive guarded moves, same result each time, all worked!  I'll assume that this one invocation is a sufficient test.

kmdalal:
I just noticed many LCP errors from consecutive identical guarded moves.  This does not seem right, since there are no arm collisions...

anjan1984:
> I just noticed many LCP errors from consecutive identical guarded moves. This does not seem right, since there are no arm collisions...

there is collision between arm and ground. Are you getting LCP errors in actions only or both actions and services?

kmdalal:
> there is collision between arm and ground. Are you getting LCP errors in actions only or both actions and services?

Both, but only a few, and I realize these are normal now.

kmdalal (approved):
Useful fix!

Samahu:
These values don't work for me and causes the arm to snap after failing to find the ground on first contact:
```python
    goal.start.x = 1.7
    goal.start.y = 0.0
    goal.start.z = 0.5
    goal.normal.x = 0.5
    goal.normal.y = -0.5
    goal.normal.z = 1.0
    goal.search_distance = 0.8
```

image

Note that this works smoothly on the ros services side, you can try it here:
`rosservice call /arm/guarded_move -- False 1.7 0.0 0.5 0.5 -0.5 1 0.8`

Samahu:
> These values don't work for me and causes the arm to snap after failing to find the ground on first contact:
> 
> ```python
>     goal.start.x = 1.7
>     goal.start.y = 0.0
>     goal.start.z = 0.5
>     goal.normal.x = 0.5
>     goal.normal.y = -0.5
>     goal.normal.z = 1.0
>     goal.search_distance = 0.8
> ```
> 
> image
> 
> Note that this works smoothly on the ros services side, you can try it here:
> `rosservice call /arm/guarded_move -- False 1.7 0.0 0.5 0.5 -0.5 1 0.8`

I have even seen the ros action server for guarded move crash in one of the attempts.

anjan1984:
Updated the branch. removed hard coded ratio between guardedmove and pre-guardedmove trajectory with expected ratio and  a tolerance. tested with different terrains. please review

Samahu (approved):
Works as described!

PR #170, opened 2021-08-30
OCEANWATER-785: prototype antenna faults by locking joints

Samahu:
## Linked Issues:
| EPIC ⚡ | [OCEANWATER-551 / OW Faults Injection](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-551) |
| Jira Ticket 🎟️ | [OCEANWATER-785 / Prototype antenna faults by locking joints](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-785) |
| Github :octocat:  | N/A |

## Summary of Changes
* Implemented a **JointsFaults** module that utilizes joint limits to lock the joints at their current position
  - Changed antenna joint types from **continuous** to **revolute** joint types (required to impose joint limits)
* Currently only applied to _j_ant_pan_ and _j_ant_tilt_ joints but the concept can be expanded to other joints

> Note: this is a quick prototype to get some feedback ad may not be eventually merged into **melodic-devel**

## Test
* Launch the simulation
```bash
```
* Verify that you can freely control the antenna joints
```bash
rostopic pub --once /ant_pan_position_controller/command std_msgs/Float64 "data: 2.0"
rostopic pub --once /ant_pan_position_controller/command std_msgs/Float64 "data: 0.0"
rostopic pub --once /ant_pan_position_controller/command std_msgs/Float64 "data: 1.0"
```
* Enable a fault on the antenna pan joint
```bash
rosrun dynamic_reconfigure dynparam set /faults ant_pan_effort_failure True
```
* Try to control the antenna pan again
```bash
rostopic pub --once /ant_pan_position_controller/command std_msgs/Float64 "data: 0.0"
```
* The antenna should stand still and not move until you clear the fault:
```bash
rosrun dynamic_reconfigure dynparam set /faults ant_pan_effort_failure False
```
* Repeat same steps for the antenna tilt joint

Notes:
1. You can inject the fault while the antenna is still moving and it should stop immediately
2. When you clear the fault the antenna continues to execute the last command that was submitted
   - This is an observed behavior and was not intentional but it doesn't seem to be harmless (otherwise please comment)

### DIFF
--- a/ow_faults/include/ow_faults/JointsFaults.h
+++ b/ow_faults/include/ow_faults/JointsFaults.h
@@ -2,39 +2,39 @@
 // Research and Simulation can be found in README.md in the root directory of
 // this repository.
 
-#ifndef LinkForcePlugin_h
-#define LinkForcePlugin_h
+#ifndef JOINTS_FAULTS_H
+#define JOINTS_FAULTS_H
 
 #include <gazebo/common/Plugin.hh>
 
-namespace gazebo
-{
-class JointsFaults : public ModelPlugin
+
+class JointsFaults : public gazebo::ModelPlugin
 {
 public:
   JointsFaults();
   ~JointsFaults();
 
-  virtual void Load(physics::ModelPtr model, sdf::ElementPtr /* _sdf */);
+  virtual void Load(gazebo::physics::ModelPtr model, sdf::ElementPtr /* _sdf */);
 
 private:
-  void OnUpdate();
-
-  physics::ModelPtr m_model;
-  double m_AntennaTiltLowerLimit;
-  double m_AntennaTiltUpperLimit;
-  double m_AntennaPanLowerLimit;
-  double m_AntennaPanUpperLimit;
-  bool m_AntennaTiltFaultActivated;
-  bool m_AntennaPanFaultActivated;
+  void onUpdate();
 
   void injectFault(const std::string& joint_fault, bool& fault_activated, const std::string& joint_name,
                    double lower_limit, double upper_limit);
 
+  gazebo::physics::ModelPtr m_model;
+  double m_antennaTiltLowerLimit;
+  double m_antennaTiltUpperLimit;
+  double m_antennaPanLowerLimit;
+  double m_antennaPanUpperLimit;
+  bool m_antennaTiltFaultActivated;
+  bool m_antennaPanFaultActivated;
+
+
+
   // Connection to the update event
-  event::ConnectionPtr m_updateConnection;
+  gazebo::event::ConnectionPtr m_updateConnection;
 };
 
-}  // namespace gazebo
 
-#endif
+#endif  // JOINTS_FAULTS_H
--- a/ow_faults/include/ow_faults/JointsFaults.h
+++ b/ow_faults/include/ow_faults/JointsFaults.h
@@ -1,0 +1,40 @@
+// The Notices and Disclaimers for Ocean Worlds Autonomy Testbed for Exploration
+// Research and Simulation can be found in README.md in the root directory of
+// this repository.
+
+#ifndef LinkForcePlugin_h
+#define LinkForcePlugin_h
+
+#include <gazebo/common/Plugin.hh>
+
+namespace gazebo
+{
+class JointsFaults : public ModelPlugin
+{
+public:
+  JointsFaults();
+  ~JointsFaults();
+
+  virtual void Load(physics::ModelPtr model, sdf::ElementPtr /* _sdf */);
+
+private:
+  void OnUpdate();
+
+  physics::ModelPtr m_model;
+  double m_AntennaTiltLowerLimit;
+  double m_AntennaTiltUpperLimit;
+  double m_AntennaPanLowerLimit;
+  double m_AntennaPanUpperLimit;
+  bool m_AntennaTiltFaultActivated;
+  bool m_AntennaPanFaultActivated;
+
+  void injectFault(const std::string& joint_fault, bool& fault_activated, const std::string& joint_name,
+                   double lower_limit, double upper_limit);
+
+  // Connection to the update event
+  event::ConnectionPtr m_updateConnection;
+};
+
+}  // namespace gazebo
+
+#endif
--- a/ow_faults/src/JointsFaults.cpp
+++ b/ow_faults/src/JointsFaults.cpp
@@ -39,7 +39,7 @@ void JointsFaults::Load(physics::ModelPtr model, sdf::ElementPtr /* sdf */)
   // If result goes out of scope updates will stop, so it is assigned to a member variable.
   m_updateConnection = event::Events::ConnectBeforePhysicsUpdate(std::bind(&JointsFaults::onUpdate, this));
 
-    gzlog << "JointsFaultsP lugin - successfully loaded!" << endl;
+  gzlog << "JointsFaultsPlugin - successfully loaded!" << endl;
 }
 
 void JointsFaults::onUpdate()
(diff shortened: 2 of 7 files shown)


kmdalal:
I'm pretty buried and won't be a good reviewer for the code change.   I think @mogumbo and @lanseafood would be the best reviewers for the code, and suggest that @scheidere just try the test.

ow_faults/include/ow_faults/JointsFaults.h
```diff
@@ -0,0 +1,40 @@
+// The Notices and Disclaimers for Ocean Worlds Autonomy Testbed for Exploration
+// Research and Simulation can be found in README.md in the root directory of
+// this repository.
+
+#ifndef LinkForcePlugin_h
```
mogumbo, line 5: Wrong name

ow_faults/include/ow_faults/JointsFaults.h
```diff
@@ -0,0 +1,40 @@
...
+  JointsFaults();
+  ~JointsFaults();
+
+  virtual void Load(physics::ModelPtr model, sdf::ElementPtr /* _sdf */);
+
+private:
+  void OnUpdate();
+
+  physics::ModelPtr m_model;
+  double m_AntennaTiltLowerLimit;
```
mogumbo, line 24: "Antenna" could be "antenna" to be consistent with the rest of our code. Though, looking at the rules, I don't think they specify that this is required. I'm not sure if it matters. Any opinions?

ow_lander/urdf/lander.xacro
```diff
@@ -844,7 +844,7 @@
     <axis
       xyz="0 0 -1" />
     <dynamics damping="50.0" />
-    <limit effort="80.0" velocity="0.2" />
+    <limit effort="80.0" velocity="0.2" lower="-3.14" upper="3.14" />
```
mogumbo, line 847: There's a tiny gap in the rotation range here. I suspect a real pan joint would be engineered to rotate a little more than 2*pi to avoid getting stuck in some state where it has to wrap around over and over again. Maybe -3.5 to 3.5 or something like that.

mogumbo (changes requested):
There's a lot of commented code. Are you essentially replacing messaging failures with simulated hardware failures? That seems okay to me. Is there any reason to keep the messaging failures? If not, then can you just remove all the commented code?

ow_faults/src/JointsFaults.cpp
```diff
@@ -0,0 +1,78 @@
...
+  bool fault_enabled;
+  ros::param::param("/faults/" + joint_fault, fault_enabled, false);
+  if (!fault_activated && fault_enabled)
+  {
+    ROS_INFO_STREAM(joint_fault << " activated!");
+    fault_activated = true;
+    // lock the joint to current position
+    auto j = m_model->GetJoint(joint_name);
+    auto p = j->Position(0);
+    j->SetLowerLimit(0, p);
```
mogumbo, line 66: Wish I had thought of this when I was trying to lock joints way back when. Good idea.

ow_faults/include/ow_faults/JointsFaults.h
Samahu: Good catch, will address

ow_faults/include/ow_faults/JointsFaults.h
Samahu: you are right, will change

ow_lander/urdf/lander.xacro
Samahu: Makes sense

ow_faults/include/ow_faults/JointsFaults.h
Samahu: addressed

ow_lander/urdf/lander.xacro
Samahu: increase from 3.14 to 3.50 +/-

ow_faults/include/ow_faults/JointsFaults.h
Samahu: addressed

Samahu:
> There's a lot of commented code. Are you essentially replacing messaging failures with simulated hardware failures? That seems okay to me. Is there any reason to keep the messaging failures? If not, then can you just remove all the commented code?

I have removed commented code.

One thing I want to note to @lanseafood that in this PR I am only performing injection but no detection is done.

mogumbo (approved):

scheidere:
I ran the test for both pan and tilt a variety of times, and I am not observing consistent joint locking behavior. I believe I saw the antenna remain still when the injected pan fault was active (as seen in RQT) and I attempt control again once. However, I did not see this in subsequent test repeats, and I did not see this happen for tilt tests at all for some reason. I could be doing something wrong, but all I did was confirm I could control either pan or tilt (depending on what I was testing, making sure the fault I set to True matched), inject the associated fault, and attempt control again. 

I am seeing movement still while a fault is active.

Also, should the antenna stop what it is doing immediately if a relevant fault is set to True *during* movement? It seems this test sets the fault to True between control attempts and when I tested changing it during, I did not see the antenna stop for some reason.

Happy to test again if you have an idea of what I might be doing wrong, or if another test would be helpful.

Samahu:
> I ran the test for both pan and tilt a variety of times, and I am not observing consistent joint locking behavior. I believe I saw the antenna remain still when the injected pan fault was active (as seen in RQT) and I attempt control again once. However, I did not see this in subsequent test repeats, and I did not see this happen for tilt tests at all for some reason. I could be doing something wrong, but all I did was confirm I could control either pan or tilt (depending on what I was testing, making sure the fault I set to True matched), inject the associated fault, and attempt control again.
> 
> I am seeing movement still while a fault is active.

Maybe the instruction I provided have been confusing a bit, when you enable the fault using: `rosrun dynamic_reconfigure dynparam set /faults ant_pan_effort_failure True`, then no matter how many subsequent calls to pan the antenna using  `rostopic pub --once /ant_pan_position_controller/command std_msgs/Float64 "data: x"` (try x with different values), the antenna shouldn't pan an inch. It should still be able to tilt, they are distinct from each other. Once you clear the fault using `rosrun dynamic_reconfigure dynparam set /faults ant_pan_effort_failure False`, then panning should resume. Is this not consistent with the behavior you are seeing?

> Also, should the antenna stop what it is doing immediately if a relevant fault is set to True _during_ movement? It seems this test sets the fault to True between control attempts and when I tested changing it during, I did not see the antenna stop for some reason.

I think when I tried it on my machine it would halt movement immediately. Try executing a long pan movement, for example: from 0 to 3.14, and while the antenna still moving inject the fault. I am pretty sure this was working for me. Once you clear the fault, it would resume last issued command.

> Happy to test again if you have an idea of what I might be doing wrong, or if another test would be helpful.

If your expectations were different from what I just described then I would appreciate that you give it another try and let me know that it doesn't meet the description.

scheidere:
> Maybe the instruction I provided have been confusing a bit, when you enable the fault using: `rosrun dynamic_reconfigure dynparam set /faults ant_pan_effort_failure True`, then no matter how many subsequent calls to pan the antenna using `rostopic pub --once /ant_pan_position_controller/command std_msgs/Float64 "data: x"` (try x with different values), the antenna shouldn't pan an inch. It should still be able to tilt, they are distinct from each other. Once you clear the fault using `rosrun dynamic_reconfigure dynparam set /faults ant_pan_effort_failure False`, then panning should resume. Is this not consistent with the behavior you are seeing?
> 
That all makes sense! I only expect panning to be locked when I inject a pan fault using the command you gave, and I am not observing that. I just tried again. However, I do see an error when I run `roslaunch ow atacama_y1a.launch` that I think is related. I have sent an image of it to you on Teams.

> If your expectations were different from what I just described then I would appreciate that you give it another try and let me know that it doesn't meet the description.

I ran another test with 0 to 3.14 panning just to be sure, and I still do not see panning locked with the pan fault is active. I assume I would still see the same issue for tilt as well. Given it works on your machine, I am thinking that error I mentioned could be the cause.

PR #174, opened 2021-09-15
OCEANWATER-807: Realtime factor cut in half following merge of fault handling changes

Samahu:
## Linked Issues:
| EPIC ⚡ | [OCEANWATER-726 / Fault Handling Plans and Patterns](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-726) |
| Jira Ticket 🎟️ | [OCEANWATER-807 / realtime factor cut in half following merge of fault handling changes](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-807) |
| Github :octocat:  | N/A |

## Summary of Changes
* Problem Root Cause: The API `ros::param::param` is rather slow and invoking it with the `BeforePhysicsUpdate` event causes a significant slowdown to the physics update loop.
* Solution: Switched from `ros::param::param` to `ros::param::getCached`. The `getCached` version provides a huge speed as it queries that param value locally which gets updated by ros using a internal subscription.

## Test
* Launch the simulation
```bash
```
* Verify
  - Realtime Factor stays close to ~1.0 when the lander is in standstill position
  - Joints Fault Injection works the same as before.

### DIFF
--- a/ow_faults/src/JointsFaults.cpp
+++ b/ow_faults/src/JointsFaults.cpp
@@ -9,8 +9,9 @@
 #include <gazebo/rendering/Scene.hh>
 #include <ros/ros.h>
 
-using namespace gazebo;
 using namespace std;
+using namespace gazebo;
+using namespace gazebo::event;
 
 constexpr double JointsFaults::MAX_FRICTION;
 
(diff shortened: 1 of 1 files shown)


AstroStucky (approved):
I can confirm the 50% performance drop is gone! My simulation idles at about 0.8, and that is nominal for this PC. Great work finding the bottleneck!

I did a search in all 4 OceanWATERS repositories, and the only other place I see getParam being used is in an constructor, which shouldn't affect performance. This is still a really good thing to know about the ROS API going forward.

PR #180, opened 2021-10-13
Oceanwater 747 Fill Scoop with Sample when it Digs

AstroStucky:
## Linked Issues:
| Jira Ticket 🎟️ | [OCEANWATER-747](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-747) |
| Github :octocat:  | n/a |

https://user-images.githubusercontent.com/17039973/137366612-4da6d98b-72dd-448c-842a-c83bcc874a3c.mp4

## Summary of Changes

* Both the visual and collision Gazebo plug-ins defined in ow_dynamic_terrain now publish a differential image describing how terrain has been modified on /ow_dynamic_terrain/modification_differential/visual and /ow_dynamic_terrain/modification_differential/collision, respectively.
* API method calls that reinitialize terrain objects and physics and are required following a visual or collision terrain modification will now only ever occur if there was an actual change in height computed by the `merge_method`. Similarly, the modification_differential topic will only be published if a change of height has occurred. This has the benefit of significantly speeding up the simulation when arm digging and grinding actions are operating.
* A new package, ow_regolith, has been added which defines `regolith_node`, a ROS node, that tracks changes to the visual terrain via the modification_differential topic. When a threshold volume of terrain has been removed, a regolith model will be spawned into the scoop, and held there by a fake force until the dig operation completes. The package also handles clean-up of these regolith models by simply deleting all of them at the end of an deliver arm action. 

## Caveats 

* The node will not work properly if the user calls the ROS Service versions of arm operations, instead of their action counterparts. Keep this in mind when testing! Do not test with ROS services.
* While sample is in the scoop you may see the following errors flood the console `ODE Message 3: LCP internal error, s <= 0 (s=0.0000e+00)`. I will attempt to address this in [OCEANWATER-819](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-819)

## Test 1 - using default digging parameters
1. Launch the simulation to the `atacama_y1a` world settings the sim_regolith flag to true
```roslaunch ow atacama_y1a.launch sim_regolith:=true```
2. In another terminal call the grind action
```./src/ow_simulator/ow_lander/scripts/grind_action_client.py```
No material will spawn as a result of the grind operation, but ensure the operation performs nominally.
3. Call either the dig linear or dig circular actions
```./src/ow_simulator/ow_lander/scripts/dig_linear_action_client.py```
While digging occurs terrain will be modified, and every so often a black sphere should spawn in the scoop and get pushed to the back. Some may fall out, but most will be held in the scoop. When the dig action completes, the fake force holding the spheres should be release and you should see them settle to the bottom.
4. Call the deliver arm action.
```./src/ow_simulator/ow_lander/scripts/deliver_action_client.py```
The spheres should remain in the scoop up until the actual delivery. Do not worry if the scoops fail to fall into the sample dock, that's a problem outside of the scope of this PR. When the deliver action is complete, you should see the spheres be deleted. You can confirm this by looking in the models dropdown box in Gazebo and confirming all models with the name `regolith_#` are gone.

## Test 2 - using non-default digging parameters
1. Use non-default values by modifying `grind_action_client.py` and either `dig_circular_action_client.py` or `dig_linear_action_client.py` files located in `ow_lander/scripts`.  To change action parameters modify where they are defined under the `# Default trenching values` line. Careful not to pick values that will place the arm outside of the workspace. What follows are examples of how I modified my `grind_action_client.py` and `dig_circular_action_client.py` to perform my test.
```Python
    # in grind_action_client.py
    goal.x_start = 1.65
    goal.y_start = 0.5
    goal.depth = 0.05
    goal.length = 0.6 
    goal.parallel = False
    goal.ground_position = constants.DEFAULT_GROUND_HEIGHT
```
```Python
    # in dig_circular_action_client.py
    goal.x_start = 1.65
    goal.y_start = 0.5
    goal.depth = 0.01
    goal.parallel = False
    goal.ground_position = constants.DEFAULT_GROUND_HEIGHT
```
2. Repeat Test 1 using your modified scripts and calling either dig linear or dig circular (depending which you modified).
3. When finished, be sure to revert the action script changes you made with the following command
```Python 
git checkout ow_lander/scripts/dig_linear_action_client.py ow_lander/scripts/dig_circular_action_client.py ow_lander/scripts/grind_action_client.py
```


ow_regolith/README.md
```diff
@@ -0,0 +1,129 @@
...
+- While sample is in the scoop you may see the following error flood the console
+```ODE Message 3: LCP internal error, s <= 0 (s=0.0000e+00)```
+We're still looking into how to avoid or suppress this message. For now it just
+has to be ignored.
+
+## Usage
+
+### Launch File
+
+The `regolith_node` may be added to a launch file like so
```
kmdalal, line 77: suggest replacing "like so" with "as follows".

ow_regolith/README.md
```diff
@@ -0,0 +1,129 @@
...
+terrain before a regolith model is spawned. It may have to be fine-tuned if a 
+regolith model of larger size is used in place of the default to avoid 
+consecutively spawned models colliding with each other.
+
+`regolith_model_uri` tells the node which model out of the Gazebo model database
+should be spawned each time the `spawn_volume_threshold` is reached.
+
+### ROS Service
+
+ROS Service support for this node is still to come...
```
kmdalal, line 98: Suggest simply saying that ROS Services are not supported at this time.  I'm not sure if and when we will prioritize further ROS Service work.

ow_regolith/include/RegolithSpawner.h
```diff
@@ -0,0 +1,82 @@
...
+
+#include "ow_dynamic_terrain/modified_terrain_diff.h"
+#include "ow_lander/DigLinearActionResult.h"
+#include "ow_lander/DigCircularActionResult.h"
+#include "ow_lander/DeliverActionResult.h"
+
+class RegolithSpawner
+{
+public:
+  RegolithSpawner() = delete;
```
kmdalal, line 19: Good.  I would also be explicit regarding the copy constructor, assignment operator, and destructor.

ow_regolith/models/ball_icefrag_2cm/model.config
```diff
@@ -0,0 +1,19 @@
...
+  <version>1.0</version>
+  <sdf version="1.6">model.sdf</sdf>
+
+  <author>
+    <name>Thomas Stucky</name>
+    <email>thomas.stucky@nasa.gov</email>
+  </author>
+
+  <description>
+    Sphere of 2cm diameter with material properties that immitates ground ice.
```
kmdalal, line 14: typo: imitate.

ow_regolith/include/RegolithSpawner.h
AstroStucky: Agreed. It's good to be thorough. Could you explain the reasoning for deleting the class' destructor? Is this simply done if you never intend to manually allocate/deallocate a class using new/delete keywords?

ow_regolith/src/RegolithSpawner.cpp
```diff
@@ -0,0 +1,295 @@
...
+const static string SRV_SPAWN_MODEL    = "/gazebo/spawn_sdf_model";
+const static string SRV_DELETE_MODEL   = "/gazebo/delete_model";
+const static string SRV_APPLY_WRENCH   = "/gazebo/apply_body_wrench";
+const static string SRV_CLEAR_WRENCH   = "/gazebo/clear_body_wrenches";
+
+// topic paths used in class
+const static string TOPIC_MODIFY_TERRAIN_VISUAL = "/ow_dynamic_terrain/modification_differential/visual";
+const static string TOPIC_DIG_LINEAR_RESULT     = "/DigLinear/result";
+const static string TOPIC_DIG_CIRCULAR_RESULT   = "/DigCircular/result";
+const static string TOPIC_DELIVER_RESULT        = "/Deliver/result";
```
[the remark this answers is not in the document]
Samahu: >Side Note: ROS melodic is C++14, ROS noetic is C++17

AstroStucky:
Made changes suggested by @kmdalal. I held off on merging in melodic-devel for the time being, since we're seeing a slowdown issue in that branch.

kmdalal (approved):
I'm satisfied with my testing, and am okay with the loose ends being addressed after Release 9.  Since this feature is optional at launch, there is no risk.  Great addition to the simulation!

ow_dynamic_terrain/msg/modified_terrain_diff.msg
```diff
@@ -0,0 +1,6 @@
+sensor_msgs/Image diff          # an image containing changes in height
+geometry_msgs/Point32 position  # position in the real world where the terrain
+                                # modification took place, position is the 
+                                # center of the patch
```
Samahu, line 4: I don't understand what do you mean by "position is the center of the patch", which patch?

ow_dynamic_terrain/src/DynamicTerrainBase.cpp
```diff
@@ -37,6 +37,9 @@ void DynamicTerrainBase::Initialize(const std::string& topic_extension)
   subscribe<modify_terrain_patch>("modify_terrain_patch", on_modify_terrain_patch);
   subscribe<modify_terrain_patch>("modify_terrain_patch/" + topic_extension, on_modify_terrain_patch);
 
+  m_differential_pub = m_node_handle->advertise<modified_terrain_diff>(
+      "/" + m_package_name + "/modification_differential/" + topic_extension, 1);
```
Samahu, line 41: What is the purpose of adding it at the base class here when you only intend to use one of the two topics (the "visual")? Is it to have the option to alternate between the two?

ow_dynamic_terrain/src/TerrainModifier.cpp
```diff
@@ -215,13 +257,35 @@ void TerrainModifier::applyImageToHeightmap(Heightmap* heightmap, const Point2i&
...
+      
+      // if we make it here, flag that a change has occurred
+      change_occurred = true;
     }
+
+    if (change_occurred)
+    {
+      out_diff_image.image    = diff;
+      out_diff_image.encoding = image_encodings::TYPE_32FC1;
+    }
```
Samahu, line 288: Why do you need to check for _change_occured_ before copying the diff? Is it for performance?

ow_regolith/include/RegolithSpawner.h
```diff
@@ -0,0 +1,82 @@
...
+  // must be called before the other functions
+  bool initialize(void);
+
+  // spawn the regolith model just above the tip of the scoop and apply a force
+  // that keeps it in the scoop during the remainder of scooping operation
+  bool spawnRegolithInScoop(tf::Vector3 pushback_direction = tf::Vector3());
+
+  void clearAllPsuedoForces(void);
+
+  void removeAllRegolithModels(void);
```
Samahu:
```suggestion
  void removeAllRegolithModels();
```

Samahu (changes requested):
Sorry Thomas, I haven't finished reviewing your PR. I think am half way. Will continue on Thursday. Nonetheless, I had some few comments that needs to be addressed.

ow_regolith/src/RegolithSpawner.cpp
```diff
@@ -0,0 +1,295 @@
...
+        SRV_SPAWN_MODEL, m_gz_spawn_model, true) ||
+      !createRosServiceClient<DeleteModel>(m_node_handle, 
+        SRV_DELETE_MODEL, m_gz_delete_model, true) ||
+      !createRosServiceClient<ApplyBodyWrench>(m_node_handle, 
+        SRV_APPLY_WRENCH, m_gz_apply_wrench, true) ||
+      !createRosServiceClient<BodyRequest>(m_node_handle, 
+        SRV_CLEAR_WRENCH, m_gz_clear_wrench, true)) {
+    ROS_ERROR("Failed to connect to all required ROS services");
+    return false;
+  }
```
Samahu, line 129: Hardly readable.. please extend the line limit to 120 chars.

ow_regolith/src/RegolithSpawner.cpp
```diff
@@ -0,0 +1,295 @@
...
+  m_active_models.push_back({model_name.str(), body_name.str()});
+
+  return callRosService(m_gz_apply_wrench, wrench_msg);
+}
+
+
+void RegolithSpawner::clearAllPsuedoForces()
+{
+  for (auto &regolith : m_active_models) {
+    BodyRequest msg;
```
Samahu, line 219: I think this would be slightly more efficient (unless the compiler is smart enough):
```suggestion
{
  BodyRequest msg;
  for (auto &regolith : m_active_models) {
```

Note this is okay since BodayRequest has only one field that you overwrite on each iteration

Samahu (changes requested):
more comments .. 70% through

ow_regolith/src/RegolithSpawner.cpp
```diff
@@ -0,0 +1,295 @@
...
+    ROS_ERROR("Regolith node requires the regolith_model_uri paramter.");
+}
+
+bool RegolithSpawner::initialize() 
+{
+  // set the maximum scoop inclination that the psuedo force can counteract
+  // to be 45 degrees (excluding friction)
+  // NOTE: do not use a value that makes cosine zero!
+  constexpr auto MAX_SCOOP_INCLINATION_DEG = 70.0f; // degrees
+  constexpr auto MAX_SCOOP_INCLINATION_RAD
```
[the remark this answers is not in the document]
kmdalal: Our OW coding guidelines suggest 80, with a _max_ of 120.  That has been my personal standard all my career, which I find most readable, provided that the code is properly indented.

ow_regolith/src/RegolithSpawner.cpp
kmdalal: I agree it's hard to read, but because of bad indentation.  I suggest this, using the first clause as an example:
```
if (!createRosServiceClient<GetPhysicsProperties>(m_node_handle, 
                                                                                        SRV_GET_PHYS_PROPS,
                                                                                        gz_get_phys_prop, 
                                                                                        false)
```

which I think more readable to a wider audience than a really long line.

ow_regolith/include/RegolithSpawner.h
AstroStucky: Sorry, I misunderstood. All the methods you mentioned are now explicitly defined or deleted.

AstroStucky:
My solution for particle scooping retention needs more time. I'll let you all know when its been committed.

AstroStucky:
I just pushed a fix for the particle retention while scooping. For the tests I performed all particles remained in the scoop, but please let me know if you see otherwise.

ow_regolith/src/RegolithSpawner.cpp
```diff
@@ -0,0 +1,295 @@
...
+    ROS_ERROR("Timed out waiting for service %s to advertise", 
+      service_path.c_str()
+    );
+    return false;
+  } 
+  return true;
+}
+
+RegolithSpawner::RegolithSpawner(ros::NodeHandle* nh)
+  : m_volume_displaced(0.0), m_node_handle(nh)
```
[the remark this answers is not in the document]
AstroStucky: With the latest version `m_previous_center` has been removed, and the pushback direction is computed from the scoop's orientation.

ow_regolith/src/RegolithSpawner.cpp
```diff
@@ -0,0 +1,295 @@
...
+    // compute scooping direction
+    auto scooping_dir = center - m_previous_center;
+    scooping_dir.setZ(0.0); // flatten against X-Y plane (cheap projection)
+
+    if (!spawnRegolithInScoop(-scooping_dir.normalize()))
+      ROS_ERROR("Failed to spawn regolith in scoop");
+  }
+
+  // cache center so on next call we know in what direction the scoop has moved
+  m_previous_center = center;
```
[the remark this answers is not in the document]
AstroStucky: Saving the previous center less frequently makes sense. Thank you for the suggestions. However, the pushback direction is now computed an entirely different way that I think is much less error prone.

ow_regolith/src/RegolithSpawner.cpp
```diff
@@ -0,0 +1,295 @@
...
+  SpawnModel spawn_msg;
+  
+  spawn_msg.request.model_name                  = model_name.str();
+  spawn_msg.request.model_xml                   = m_model_sdf;
+  spawn_msg.request.robot_namespace             = "/regolith";
+  spawn_msg.request.reference_frame             = "lander::l_scoop_tip";
+
+  spawn_msg.request.initial_pose.position.x     = 0.0;
+  spawn_msg.request.initial_pose.position.y     = 0.0;
+  spawn_msg.request.initial_pose.position.z     = -0.05;
```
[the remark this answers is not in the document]
AstroStucky: This suggestion is compatible with the new way in which the pushback direction is computed, but I want to hold off on applying until we see further issues with particle retention. I worry it could cause particles to spawn on top of previously spawned particles in instances where the scoop is extremely full.

AstroStucky:
@mikedalal and @Samahu please note that I modified the testing procedure by adding a second test which uses non-default trenching parameters. You are welcome to select your own parameters for **Test 2**.

Samahu (approved):
This definitely works better than before. Some samples are still lost but very rare (1 out of 10 based on my testing).

Had some minor optional comments

ow_regolith/src/RegolithSpawner.cpp
```diff
@@ -248,7 +269,25 @@ void RegolithSpawner::removeAllRegolithModels()
...
+      m_scoop_orientation = Quaternion(pose_it->orientation.x, 
+                                       pose_it->orientation.y,
+                                       pose_it->orientation.z,
+                                       pose_it->orientation.w);
+      return;
+    }
+    ++name_it;
+    ++pose_it;
+  }
+  ROS_WARN("Failed to find %s in link states", m_scoop_link_name.c_str());
```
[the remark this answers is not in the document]
Samahu: [required] Use tf::quaternionMsgToTF instead of manually assigning the field. The STL find is still optional.
```suggestion
  auto name_it = std::find(begin(msg->name), end(msg->name), m_scoop_link_name);
  if (name_it == msg->name.end())
  {
    ROS_WARN("Failed to find %s in link states", m_scoop_link_name.c_str());
    return;
  }

  auto pose_it = msg->pose.begin() + (name_it - msg->name.begin());
  tf::quaternionMsgToTF(pose_it->orientation, m_scoop_orientation);
```

Samahu (changes requested):
Utilize `tf::quaternionMsgToTF` instead of handling the data conversion type between `geometry_msgs::Quaternion` and `tf::Quaternion` yourself.

ow_regolith/include/RegolithSpawner.h
```diff
@@ -24,61 +25,74 @@ class RegolithSpawner
...
 
   // loads regolith SDF and computes its mass
-  // must be called before the other functions
+  // NOTE: this must be called before any other functions
   bool initialize();
 
   // spawn the regolith model just above the tip of the scoop and apply a force
   // that keeps it in the scoop during the remainder of scooping operation
-  bool spawnRegolithInScoop(const tf::Vector3 &pushback_direction = tf::Vector3());
+  bool spawnRegolithInScoop(bool with_pushback = true);
```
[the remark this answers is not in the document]
AstroStucky: `with_pushback = false` will be used by a service call which is not yet integrated into `regolith_node`, but I hope to do so later as an improvement. The service would be a simple way for a user to manually spawn regolith particles into the scoop, and the pushback force will either not be used at all when it is called or turning it on will be a parameter of the service (I haven't decided which makes more sense yet).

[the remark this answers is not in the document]
AstroStucky: Thinking twice about this, it really does not need the default argument... I'm kind of on the fence, but I'm leaning towards removing it so all calls to this function explicitly toggle `with_pushback` on or off.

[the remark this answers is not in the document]
Samahu: I recommend removing it. You can add it back if there is a need for it

ow_regolith/src/RegolithSpawner.cpp
```diff
@@ -116,55 +128,56 @@ bool RegolithSpawner::initialize()
...
-      !createRosServiceClient<ApplyBodyWrench>(m_node_handle,
-                                               SRV_APPLY_WRENCH,
-                                               m_gz_apply_wrench,
-                                               true) ||
-      !createRosServiceClient<BodyRequest>(m_node_handle,
-                                           SRV_CLEAR_WRENCH,
-                                           m_gz_clear_wrench,
-                                           true))
+                                                    false)  ||
+      !createRosServiceClient<SpawnModel>(          m_node_handle,
```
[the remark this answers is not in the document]
AstroStucky: I don't really care for any of the options available for this particular code pattern, and I feel when you can't find a way to make the code pretty, just let it be ugly. I'll leave it how it is for now, so it remains under 80 characters in length.

AstroStucky:
Thanks @Samahu! Those were all great suggestions.

ow_regolith/include/RegolithSpawner.h
```diff
@@ -74,7 +74,7 @@ class RegolithSpawner
   // regolith will spawn once this amount of volume is displaced
   double m_spawn_threshold;
   // Gazebo link name of the scoop particles will spawn in
-  std::string m_scoop_link_name;
+  std::string m_scoop_linkname;
```
Samahu, line 77: I think **m_scoop_link_name** is the conformant style, but anyway
```suggestion
  std::string m_scoop_link_name;
```

ow_regolith/include/RegolithSpawner.h
```diff
@@ -30,7 +30,7 @@ class RegolithSpawner
 
   // spawn the regolith model just above the tip of the scoop and apply a force
   // that keeps it in the scoop during the remainder of scooping operation
-  bool spawnRegolithInScoop(bool with_pushback = true);
+  bool spawnRegolithInScoop(bool with_pushback);
```
Samahu, line 33: I suggested to remove argument in its entirety and re-introduce it when it has actual usage in the code but anyway.

Samahu (approved):
Finally the lander can dig and scoop out samples! :1st_place_medal:

kmdalal:
It's probably unrelated to this PR, but the first time I tried delivery, nothing happened, and the output was this:
```
./deliver_action_client.py 
[INFO]: Result: final: 
  x: 0.0
  y: 0.0
  z: 0.0
```
There were no unusual terminal messages.  The preceding dig looked perfect and there were many spheres in the scoop.  The subsequent delivery attempt was fine.

kmdalal (approved):
I tried a couple successive digs, going slightly deeper the second time.  Definitely an improvement in retention of spheres.  I'm foregoing another code review.   Nice work!

ow_regolith/include/RegolithSpawner.h
AstroStucky: I see your point about not developing a feature before it is in use, but since it's already there and this is part of an interface and I am inclined to keep it.

Samahu:
@AstroStucky We need a similar PR against the `noetic-devel` branch. @kmdalal

PR #199, opened 2021-11-18
Oceanwater 850 Fix arm check test calling deliver with parameters

AstroStucky:
## Linked Issues:
| EPIC ⚡ | N / A |
| Jira Ticket 🎟️ | [OCEANWATER-850](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-850) |
| Github :octocat:  | # |

## Summary of Changes
* arm_check test no longer calls deliver service with parameters
* arm_check now normalizes its joint goals before comparing with the normalized joint states
* arm_check uses arm_deliver_final as the goal for the deliver test

## Test
@mogumbo, feel free to just do a code review, if that's all you've got time for.
@anjan1984, when you test this and we merge it into release, then we can count the results as **Tester 3** on the test page.

### Procedure
Perform [OW-TEST-009-3 Sample Collection](https://babelfish.arc.nasa.gov/confluence/display/OCEANWATERS/OW-TEST-009-3+Sample+Collection) using rostest and ensure all tests pass.
```rostest ow_sim_tests arm_check.test```

### DIFF
--- a/ow_sim_tests/test/arm_check.py
+++ b/ow_sim_tests/test/arm_check.py
@@ -38,7 +38,7 @@ class ArmCheck(unittest.TestCase):
 
   def _normalize_angle(self, angle):
     """
-    Convenience method to nomrlaize angles to the range [-pi, pi]
+    Convenience method to normalize angles to the range [-pi, pi]
     @param angle: An angle that might need to be normalized
     @returns: the normalized value of the angle
     """
@@ -66,19 +66,20 @@ class ArmCheck(unittest.TestCase):
 
     rospy.wait_for_service(service_name)
     test_start_time = rospy.get_time()
-    joints_goal_array = np.array(joints_goal)
+    norm_joints_goal_array = np.array(
+      [self._normalize_angle(e) for e in joints_goal])
     rospy.loginfo("""
     test info:
     ----------
     service name: {}
     joints goal: {}
     start time: {}""".format(
-      service_name, np.round(joints_goal_array, 2), test_start_time))
+      service_name, np.round(norm_joints_goal_array, 2), test_start_time))
     arm_activity = rospy.ServiceProxy(service_name, service_type)
     success = arm_activity() if service_args is None else arm_activity(*service_args)
     self.assertTrue(success, "submitted request to " + service_name)
     goal_state_achieved = False
-    joints_abs_diff_array = np.full(joints_goal_array.shape, np.inf)
+    joints_abs_diff_array = np.full(norm_joints_goal_array.shape, np.inf)
 
     elapsed = 0
     while not rospy.is_shutdown() and \
@@ -89,7 +90,7 @@ class ArmCheck(unittest.TestCase):
       norm_joints_state_array = np.array(
           [self._normalize_angle(e) for e in joints_state])
       joints_abs_diff_array = np.abs(
-          joints_goal_array - norm_joints_state_array)
+          norm_joints_goal_array - norm_joints_state_array)
       goal_state_achieved = all(joints_abs_diff_array < tolerance)
       rospy.loginfo("""
       current joints state: {}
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: ow_sim_tests/test/arm_check.py [hunk 1]
```diff
@@ -89,7 +90,7 @@ def _test_activity(self, service_name, service_type,
       norm_joints_state_array = np.array(
           [self._normalize_angle(e) for e in joints_state])
       joints_abs_diff_array = np.abs(
-          joints_goal_array - norm_joints_state_array)
+          norm_joints_goal_array - norm_joints_state_array)
```
mogumbo, line 93: I think it's safer to normalize to {-pi, pi} after this subtraction happens. If you normalize both values before, there's a small chance your difference could end up being close to 2*pi, much greater than the tolerance.

mogumbo (changes requested):

ow_sim_tests/test/arm_check.py
AstroStucky: That makes sense. I wanted to prove it to myself with a test. So setting `l_hand_yaw` to exactly pi for `arm_unstowed` in **lander.rsdf**, I did confirm this can happen, and it happens only half the time, which makes sense. I would expect the error depends on whether final joint state is higher/lower than joint goal.
```[rosout][INFO] 2021-11-23 11:32:58,055: 
      current joints state: [ 0.    1.57 -1.57 -0.   -3.14  0.01]
      diff joints state: [0.   0.   0.   0.   6.28 0.01]
      maximum diff between expected/current: 6.28
[rosout][INFO] 2021-11-23 11:32:58,257: operation /arm/unstow took 30s, time-limit is 30s
```
I get the same fails half of the time result when using -pi.

As of commit 3a22af it now normalizes after taking the difference between the angles.

AstroStucky:
Latest commit fixes a normalize angle edge case. Please retest this.

I also changed the test print statement to include negatives in the difference values. So instead of 
```
[rosout][INFO] 2021-11-23 13:02:36,962: 
      current joints state: [ -0.     1.57   4.71  -0.   -11.     0.02]
      diff joints state: [ 0.    0.    0.    0.   0.   0.02]
      maximum diff between expected/current: 0.02
```
it will show
```
[rosout][INFO] 2021-11-23 13:02:36,962: 
      current joints state: [ -0.     1.57   4.71  -0.   -11.     0.02]
      diff joints state: [ 0.    0.    0.    0.   -0.   -0.02]
      maximum diff between expected/current: -0.02
```
It's a subtle change, but I think the latter gives more information, so is therefore a better choice.

anjan1984:
Ran twice and all the tests are successful 
[ROSTEST]-----------------------------------------------------------------------

[ow_sim_tests.rosunit-arm_check/test_01_unstow][passed]
[ow_sim_tests.rosunit-arm_check/test_02_guarded_move][passed]
[ow_sim_tests.rosunit-arm_check/test_03_unstow][passed]
[ow_sim_tests.rosunit-arm_check/test_04_grind][passed]
[ow_sim_tests.rosunit-arm_check/test_05_dig_circular][passed]
[ow_sim_tests.rosunit-arm_check/test_06_grind][passed]
[ow_sim_tests.rosunit-arm_check/test_07_dig_linear][passed]
[ow_sim_tests.rosunit-arm_check/test_08_deliver_sample][passed]
[ow_sim_tests.rosunit-arm_check/test_09_unstow][passed]
[ow_sim_tests.rosunit-arm_check/test_10_stow][passed]

SUMMARY
 * RESULT: SUCCESS
 * TESTS: 10
 * ERRORS: 0
 * FAILURES: 0

anjan1984 (approved):

mogumbo (approved):

ow_sim_tests/test/arm_check.py
Samahu: @mogumbo You comment is right but doesn't apply to the case of the test. We are checking for equal angle values so when both values are normalized and they are equal they wouldn't exceed the {-pi, pi} otherwise they are not equal to start with. I reverted the changes made to this code block to its original implementation to keep it simple.

PR #206, opened 2021-12-08
OceanWATER-840 Added spawn regolith and remove regolith services

AstroStucky:
## Linked Issues:
| Jira Ticket 🎟️ | [OCEANWATER-840](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-840) |
| Github :octocat:  | # |

## Summary of Changes
* Added `ow_regolith/spawn_regolith_in_scoop` and `/ow_regolith/remove_all_regolith`.

`ow_regolith/spawn_regolith_in_scoop` will spawn a single regolith model just above the tip of the scoop. Unlike when spawning is triggered by terrain modification, there will be no psuedo force applied to the model when this service is called.

`ow_regolith/remove_all_regolith` will delete all regolith models from the simulation.

Both services together provide a minimal interface into regolith_node that can be helpful for both debugging purposes and some specific test cases.

## Test
1. Launch whatever world you'd like.
2. Move the arm into whatever pose you'd like it to receive regolith in. Note that any pose that does not point the scoop up will cause regolith models to fall to the ground because there is no force to keep them in the scoop. This is intended behavior.
3. Use either RQT or the command line to call `/ow_regolith/spawn_regolith_in_scoop`. Call the service as many times as you'd like.
4. Open the **Models** dropdown box in gzclient and verify as many regolith models have spawned as times you called the service from step 3. 
5. Call `/ow_regolith/remove_all_regolith` and verify that all models are removed from the Gazebo world.

### DIFF
--- a/ow_regolith/CMakeLists.txt
+++ b/ow_regolith/CMakeLists.txt
@@ -16,6 +16,7 @@ find_package(catkin REQUIRED COMPONENTS
   cv_bridge
   ow_dynamic_terrain
   ow_lander
+  message_generation
 )
 
 include(FindPkgConfig)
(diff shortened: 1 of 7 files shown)


kmdalal (approved):
Tried several tests, worked as described.    A nice future enhancement may be to specify the number of sphere's collected as a service argument.

AstroStucky:
> Tried several tests, worked as described. A nice future enhancement may be to specify the number of sphere's collected as a service argument.
This might be doable. I would like to avoid spawning models one after another like how spawn_bulk_sample.sh does it, but that would require RegolithSpawner to somehow know the dimensions of the regolith model being used, so it can space their spawn positions apart enough to avoid overlap. If a non-default regolith model is being used, it's not clear how it could determine those dimensions. 🤔 

I will look into this, because I agree it would be nice to have.

Samahu (approved):

PR #212, opened 2021-12-23
Oceanwater 852 fixed regolith spawned during grind

AstroStucky:
## Linked Issues:
| Jira Ticket 🎟️ | [OCEANWATER-852](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-852) |
| Github :octocat:  | # |

## Summary of Changes
* Scoop yaw now has to be at 90 deg before scoop may modify the terrain
  - Fewer unnecessary publishes to `ow_dynamic_terrain/modify_terrain_ellipse/visual` topic
  - Guarded move no longer deforms the terrain
  - Fewer sporadic visual terrain modifications, which sometimes would result in regolith being spawned during a grind operation.

## Caveat
This change does not entirely eliminate the sporadic visual terrain mods, but does reduce them. This is probably because grinding does modify the visual terrain, but in a way that rarely produces any real change to the terrain. 
@Samahu What do you think about removing the visual terrain modification call in the grinder script to ensure this no longer happens? I can't seem to notice the affects of this call in general anyways.

## Test
1. Launch whatever world you'd like.
2. Run a grind action. Ensure no regolith particles are spawned during it.
3. Run a scoop action. Ensure regolith particles are spawned during it.

### DIFF
--- a/ow_dynamic_terrain/scripts/modify_terrain_scoop_pub.py
+++ b/ow_dynamic_terrain/scripts/modify_terrain_scoop_pub.py
@@ -12,6 +12,7 @@ import rospy
 import numpy as np
 from geometry_msgs.msg import Point
 from gazebo_msgs.msg import LinkStates
+from sensor_msgs.msg import JointState
 from tf.transformations import euler_from_quaternion
 from ow_dynamic_terrain.msg import modify_terrain_ellipse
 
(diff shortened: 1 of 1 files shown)


kmdalal:
I will just test, @Samahu will just review code.

AstroStucky:
@Samahu elucidated how the grinder modifies terrain in discussions this morning. The modification is only noticeable when the grinder plunges far enough into the terrain. I think changing this behavior is better saved for a future PR that seeks to improve grinder terrain visual feedback. That does mean that for now deep enough grinds will cause regolith to spawn in the scoop.

kmdalal:
> That does mean that for now deep enough grinds may cause regolith to spawn in the scoop.

Is this PR still valid then?  I'll test successive grinds, each a bit deeper.

kmdalal:
Doing a default Grind with 0.05 depth, followed by another with 0.1 depth, no regolith appeared.  But when I tried with depth 0.15 (the first that visually modified the terrain), two regolith appeared in the Models list.

AstroStucky:
> > That does mean that for now deep enough grinds may cause regolith to spawn in the scoop.
> 
> Is this PR still valid then? I'll test successive grinds, each a bit deeper.

I think it is. We still benefit from restricting when the scoop can modify terrain, in particular during guarded moves.

As for eliminating unwanted regolith spawns, that's still not entirely fixed by this PR, so I suppose I should make a few more changes that address that. Thanks to you, I now know at what grind depth terrain starts getting modified at. On which world did you see modification for 0.15 depth? 

Feel free to hold-off on reviews and testing until I have a solution.

kmdalal:
> On which world did you see modification for 0.15 depth?

Atacama.  This is the only world I tested with.  Ping me when this is ready for another test!

Samahu (changes requested):
Sorry @AstroStucky but I don't think this approach is viable. You are making the assumption that we would only scoop samples from the terrain when the scoop is parallel to the arm, which is not always the case. For example, have you thought about the **dig_circular** with _parallel_ set to `False`?

This may solve one issue for us but it breaks existing capabilities of the arm.

I don't approve this approach.

AstroStucky:
It seems I did not consider/test the parallel argument with this change. Lets discuss other solutions next week.

PR #214, opened 2022-01-06
OCEANWATER 852 - Fix regolith spawning during a grind

AstroStucky:
## Linked Issues:
| EPIC ⚡ | n/a  |
| Jira Ticket 🎟️ | [OCEANWATER-852](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-852) |
| Github :octocat:  | # |

## Summary of Changes
**ow_regolith** now requires that the scoop end-effector's bottom faces downward before regolith will spawn in the scoop as the result of a visual terrain modification. This will ensure deep grinds won't result in regolith.

## Test
In interest of saving time: @Samahu feel free to just review code. @mikedalal feel free to just perform the test.
1. Launch `atacama_y1a`
2. Call a grind action, and set the depth parameter to at least `0.15`, which will cause a deep enough grind to modify the visual terrain. Here's an example command that keeps other variables default 
```./grind_action_client.py 1.65 0.0 0.15```
3. Open the Models dropdown box in the left sidebar in Gazebo, and watch it during the grind to ensure no `reoglith_#` models are spawned.
4. Call a dig action that follows the trench previously made, and ensure `regolith_#` models spawn.

### DIFF
--- a/ow_regolith/include/RegolithSpawner.h
+++ b/ow_regolith/include/RegolithSpawner.h
@@ -93,8 +93,8 @@ private:
   double m_spawn_threshold;
   // Gazebo link name of the scoop particles will spawn in
   std::string m_scoop_link_name;
-  // a vector that describes the forward direction of the scoop
-  tf::Vector3 m_scoop_forward;
+  // vectors that describe the forward and down direction of the scoop
+  tf::Vector3 m_scoop_forward, m_scoop_down;
   // an offset relative to the scoop's frame where regolith will be spawned
   tf::Vector3 m_scoop_spawn_offset;
 
--- a/ow_regolith/include/RegolithSpawner.h
+++ b/ow_regolith/include/RegolithSpawner.h
@@ -25,7 +25,7 @@ public:
   RegolithSpawner()  = delete;
   ~RegolithSpawner() = default;
   RegolithSpawner(const RegolithSpawner&) = delete;
-  RegolithSpawner& operator=(const RegolithSpawner&) = default;
+  RegolithSpawner& operator=(const RegolithSpawner&) = delete;
 
   RegolithSpawner(ros::NodeHandle* nh);
 
@@ -91,12 +91,6 @@ private:
 
   // regolith will spawn once this amount of volume is displaced
   double m_spawn_threshold;
-  // Gazebo link name of the scoop particles will spawn in
-  std::string m_scoop_link_name;
-  // vectors that describe the forward and down direction of the scoop
-  tf::Vector3 m_scoop_forward, m_scoop_down;
-  // an offset relative to the scoop's frame where regolith will be spawned
-  tf::Vector3 m_scoop_spawn_offset;
 
   // regolith model that spawns in the scoop when digging occurs
   std::string m_model_uri;
--- a/ow_regolith/src/RegolithSpawner.cpp
+++ b/ow_regolith/src/RegolithSpawner.cpp
@@ -69,6 +69,14 @@ const static std::string TOPIC_DIG_LINEAR_RESULT     = "/DigLinear/result";
 const static std::string TOPIC_DIG_CIRCULAR_RESULT   = "/DigCircular/result";
 const static std::string TOPIC_DELIVER_RESULT        = "/Deliver/result";
 
+// constants specific to the scoop end-effector
+const static std::string scoop_link_name  = "lander::l_scoop_tip";
+const static Vector3 scoop_forward        = Vector3(1.0, 0.0, 0.0);
+const static Vector3 scoop_downward       = Vector3(0.0, 0.0, 1.0);
+const static Vector3 scoop_spawn_offset   = Vector3(0.0, 0.0, -0.05);
+
+const static Vector3 world_downward = Vector3(0.0, 0.0, -1.0);
+
 template <class T>
 static bool createRosServiceClient(unique_ptr<ros::NodeHandle> &nh,
                                    string service_path, 
@@ -102,12 +110,7 @@ static bool callRosService(ros::ServiceClient &srv, T &msg)
 }
 
 RegolithSpawner::RegolithSpawner(ros::NodeHandle* nh)
-  : m_node_handle(nh), 
-    m_volume_displaced(0.0),
-    m_scoop_forward(1.0, 0.0, 0.0),
-    m_scoop_down(0.0, 0.0, 1.0),
-    m_scoop_spawn_offset(0.0, 0.0, -0.05),
-    m_scoop_link_name("lander::l_scoop_tip")
+  : m_node_handle(nh), m_volume_displaced(0.0)
 {
   // get node parameters
   if (!m_node_handle->getParam("spawn_volume_threshold", m_spawn_threshold))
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: ow_regolith/src/RegolithSpawner.cpp [hunk 1]
```diff
@@ -310,11 +315,19 @@ void RegolithSpawner::onLinkStatesMsg(const LinkStates::ConstPtr &msg)
...
   auto pose_it = begin(msg->pose) + distance(begin(msg->name), name_it);
-  tf::quaternionMsgToTF(pose_it->orientation, m_scoop_orientation);
+  quaternionMsgToTF(pose_it->orientation, m_scoop_orientation);
 }
 
 void RegolithSpawner::onModDiffVisualMsg(const modified_terrain_diff::ConstPtr& msg)
 {
+  // check if visual terrain modification was caused by the scoop
+  Vector3 scoop_bottom(quatRotate(m_scoop_orientation, m_scoop_down));
+  if (tfDot(scoop_bottom, Vector3(0.0, 0.0, -1.0)) < 0.0)
```
Samahu, line 325: Why not save and use a normalized gravity vector that you queried for during initialization?
```suggestion
  if (tfDot(scoop_bottom, gravity_norm) < 0.0)
```
(instead of making the assumption that z is down)

ow_regolith/include/RegolithSpawner.h
```diff
@@ -93,8 +93,8 @@ class RegolithSpawner
   double m_spawn_threshold;
   // Gazebo link name of the scoop particles will spawn in
   std::string m_scoop_link_name;
-  // a vector that describes the forward direction of the scoop
-  tf::Vector3 m_scoop_forward;
+  // vectors that describe the forward and down direction of the scoop
+  tf::Vector3 m_scoop_forward, m_scoop_down;
```
Samahu, line 97: The **m_scoop_down** seems to be a constant rather than an actual value. I would suggest removing and using the value directly in the function call 
```suggestion
  tf::Vector3 m_scoop_forward;
```

ow_regolith/src/RegolithSpawner.cpp
```diff
@@ -100,6 +105,7 @@ RegolithSpawner::RegolithSpawner(ros::NodeHandle* nh)
   : m_node_handle(nh), 
     m_volume_displaced(0.0),
     m_scoop_forward(1.0, 0.0, 0.0),
+    m_scoop_down(0.0, 0.0, 1.0),
```
Samahu, line 108:
```suggestion
```

ow_regolith/src/RegolithSpawner.cpp
```diff
@@ -310,11 +315,19 @@ void RegolithSpawner::onLinkStatesMsg(const LinkStates::ConstPtr &msg)
...
 
   auto pose_it = begin(msg->pose) + distance(begin(msg->name), name_it);
-  tf::quaternionMsgToTF(pose_it->orientation, m_scoop_orientation);
+  quaternionMsgToTF(pose_it->orientation, m_scoop_orientation);
 }
 
 void RegolithSpawner::onModDiffVisualMsg(const modified_terrain_diff::ConstPtr& msg)
 {
+  // check if visual terrain modification was caused by the scoop
+  Vector3 scoop_bottom(quatRotate(m_scoop_orientation, m_scoop_down));
```
Samahu, line 324: Related to comment made in the header file:
```suggestion
  Vector3 scoop_down(0.0, 0.0, 1.0);
  Vector3 scoop_bottom(quatRotate(m_scoop_orientation, scoop_down));
```

ow_regolith/src/RegolithSpawner.cpp
Samahu, line 325: #optional: I think you can be more restrictive, 90 degrees or even 80 is unlikely to result in a valid scoop act
```suggestion
  max_potential_scooping = 0.3; // search for a reasonable value
  if (tfDot(scoop_bottom, Vector3(0.0, 0.0, -1.0)) < max_potential_scooping)
```

Samahu (changes requested):
Improvements to the code

ow_regolith/include/RegolithSpawner.h
Samahu: related comment: https://github.com/nasa/ow_simulator/pull/214#discussion_r782345160

AstroStucky: Rather than making `m_scoop_down` a constant, I suggest I keep it as a member variable and expose it as a node parameter. That way users can use their own scoop models and will just have to reconfigure this node appropriately for the new model. The same should be done for `m_scoop_forward`, `m_scoop_link_name`, and `m_scoop_spawn_offset`. The parameters will be optional, and their default value will be the same as what's currently in the constructor.

Samahu: I don't know about that ..  to me this seems over engineering a workaround.

Samahu: https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it

kmdalal:
The test failed on the first try.  I used the given grind parameters.  Two regolith (_0 and _1) appeared in the models list.

ow_regolith/include/RegolithSpawner.h
kmdalal: All of  m_scoop_forward, m_scoop_spawn_offset, m_scoop_link_name are constants.  My preference for encoding these, in order is:

1. as const file variables
2. as static class members

I do recommend abstraction over using the raw values in function calls.

kmdalal:
Minor comment outside the scope of this PR: it seems odd to have an assignment operator but not a copy constructor.  I'd think you want neither, and also that this class is a singleton.

ow_regolith/src/RegolithSpawner.cpp
AstroStucky: I actually did it this way at first, but I realized assuming that the scoop bottom has to be aligned with the gravity vector is equally as restricting as assuming it must align with `-z`, so I used the less complicated method. Furthermore, since `ow_dynamic_terrain` uses essentially the same method to decide if collision terrain deformation should occur, this approach is more consistent with approaches used in other parts of our code. This will also have to be revisited at some point anyways when/if we implement sloped scooping.

AstroStucky:
> I am "On branch OCEANWATER-852_regolith_spawned_during_grind Your branch is up to date with 'origin/OCEANWATER-852_regolith_spawned_during_grind'."

Understandable mistake. That is the branch for the now closed PR #212. I wasn't sure what best practice was around reusing the name of a deleted branch name, so the branch for this PR is `OCEANWATER-852_regolith_spawned_during_grind_2` just to be on the safe side.

The fact that testing the old branch fails just reaffirms the need to close PR #212 because clearly it did not solve the bug.

AstroStucky:
> Minor comment outside the scope of this PR: it seems odd to have an assignment operator but not a copy constructor. 

I agree, it doesn't need assignment. 

> ... also that this class is a singleton.

It is a class instanced only ever once. Instancing it twice will cause ROS to report an error when the second instance tries to advertise the same services as the first instance, but aside from that error it doesn't seem to break much. It turns out instancing multiple versions of this class could even be a hacky way to spawn regolith models of different types at the same time.

Since it is not absolutely necessary that this instance be unique, I am hesitant to make it a singleton.

ow_regolith/src/RegolithSpawner.cpp
AstroStucky: I still prefer the viable dig arc being 180 degrees. There are cases where a dig linear that starts low to the grind will remove terrain at the very start of the first circular trajectory, when the scoop's bottom is 90 degrees relative to the horizontal. I think in this particular case any volume removed by the scoop should still be tracked by the regolith node.

AstroStucky:
Some suggested changes made. Please review code and retest if needed.

Samahu (approved):

ow_regolith/src/RegolithSpawner.cpp
Samahu: `ow_dynamic_terrain` didn't have the gravity vector at hand. anyway like you said this entire thing is workaround. no need to overthink it.

kmdalal:
> Understandable mistake. That is the branch for the now closed PR #212. I wasn't sure what best practice was around reusing the name of a deleted branch name, so the branch for this PR is OCEANWATER-852_regolith_spawned_during_grind_2 just to be on the safe side.

Whoops, I had missed the "_2".  I'd suggest using a more easily distinguishable string after the Jira number, but no big deal.  

Retested with the correct branch and worked fine!

kmdalal:
> Since it is not absolutely necessary that this instance be unique, I am hesitant to make it a singleton.

I don't suggest that, either.  A class effectively being a singleton doesn't mean it should be forced to be.

kmdalal (approved):

AstroStucky:
> I don't suggest that, either. A class effectively being a singleton doesn't mean it should be forced to be.

Pardon me, I misunderstood the point you were making. I think, semantically speaking, a class instanced only once is not necessarily a singleton, and to be called a singleton it must additionally require that only one instance of the class may ever be instanced.

AstroStucky:
Thank you both for the helpful reviews and testing. :smiley:

kmdalal:
> > I don't suggest that, either. A class effectively being a singleton doesn't mean it should be forced to be.
> 
> Pardon me, I misunderstood the point you were making. I think, semantically speaking, a class instanced only once is not necessarily a singleton, and to be called a singleton it must additionally require that only one instance of the class may ever be instanced.

You're right, I used 'singleton' too loosely.  This is specifically a pattern that prohibits multiple instances.

PR #217, opened 2022-01-20
Removed extraneous joint_state_publisher.

mogumbo:
## Linked Issues:
| Jira Ticket 🎟️ | [OCEANWATER-888](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-888) |

## Summary of Changes
* Removed a joint_state_publisher from spawn.launch. It was unnecessary because /_original/joint_states is already being published by /gazebo due to specifying a joint_state_controller in ros_controllers.launch. The joints published by joint_state_controller are in alphabetical order while joint_state_publisher publishes them in the order they appear in the urdf, which is why something like `/joint_states/position[0]` shows different values from each of these programs.

## Test
* Use `rostopic info` to ensure that `/joint_states` and `/_original/joint_states` each have only one publisher.
* Run your favorite plexil plan to see if the sim still works as expected.

### DIFF
--- a/ow_lander/launch/spawn.launch
+++ b/ow_lander/launch/spawn.launch
@@ -64,12 +64,6 @@
     <arg name="load_robot_description" value="false"/>
   </include>
 
-  <!-- We do not have a robot connected, so publish fake joint states -->
-  <node name="joint_state_publisher" pkg="joint_state_publisher" type="joint_state_publisher">
-    <param name="use_gui" value="$(arg use_gui)"/>
-    <rosparam param="source_list">[/joint_states]</rosparam>
-  </node>
-
   <!-- Convert joint states from Gazebo to tf-tree for rviz -->
   <node name="robot_state_publisher" pkg="robot_state_publisher" type="robot_state_publisher" >
     <param name="publish_frequency" value="30"/>


ow_lander/launch/spawn.launch
```diff
@@ -64,12 +64,6 @@
     <arg name="load_robot_description" value="false"/>
   </include>
 
-  <!-- We do not have a robot connected, so publish fake joint states -->
```
mogumbo, line 67: I understand we need this hotfix fast, but I don't understand this comment or why we originally had this joint_state_publisher. Ussama wrote these lines of code, so remind me to ask him about it when he returns.

mogumbo:
> Tests passed. Studying the code, this fix seems right. Perhaps @Samahu knows why a joint_state_publisher was instantiated in the first place. The remapping scheme as I understand it should have Gazebo remap /joint_states to /_original_joint_states, which only the faults node subscribes to, and faults solely then publishes /joint_states.

The remap now appears to happen near the top of `ow/launch/common.launch` using an extra argument to gazebo. I hadn't seen remapping without the `<remap>` tag before, but I guess it can be done this way since gazebo is the one outputting the topic.

anjan1984:
I can confirm the oscillations in the value are gone and the controllers are working fine!

anjan1984 (approved):

ow_lander/launch/spawn.launch
Samahu: The change you made is legit, the **joint_state_publisher** node needs not to be published when Gazebo is launched.

Samahu:
> Tests passed. Studying the code, this fix seems right. Perhaps @Samahu knows why a joint_state_publisher was instantiated in the first place. The remapping scheme as I understand it should have Gazebo remap /joint_states to /_original_joint_states, which only the faults node subscribes to, and faults solely then publishes /joint_states.

The **joint_state_publisher** node needs not to be defined when Gazebo is used.

PR #218, opened 2022-01-21
Oceanwater 843 Reinstate persistent service clients and code clean-up

AstroStucky:
## Linked Issues:
| Jira Ticket 🎟️ | [OCEANWATER-843](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-843) |
| Github :octocat:  | # |

## Summary of Changes
* Persistent service clients are implemented as a facade pattern that encapsulates auto-reconnect logic.
  * Results in improved performance when calling commonly called services.
* Addressed a miss-use of `std::unique_ptr`. Now using `std::shared_ptr` to support the new persistent service clients.
* Numerous code clean-ups, including:
  * ALL_CAPS constant names
  * Node parameters now grabbed in `initialize()` function allowing handling of parameter errors.

## Test
Similar to past tests of this node. Everything should work the same.
1. Launch `atacama_y1a` world.
2. Call `/ow_lander/scripts/grind_action_client.py`
3. Call `/ow_lander/scripts/dig_linear_action_client.py`
    Look for particles falling into and staying in the scoop
4. Call `/ow_lander/scripts/deliver_action_client.py` or `/ow_lander/scripts/discard_action_client.py`
    Look for particles vanishing upon completion of either action

### DIFF
--- a/ow_regolith/include/RegolithSpawner.h
+++ b/ow_regolith/include/RegolithSpawner.h
@@ -8,6 +8,8 @@
 #include <ros/ros.h>
 #include <tf/tf.h>
 
+#include <ServiceClientFacade.h>
+
 #include <gazebo_msgs/LinkStates.h>
 
 #include <gazebo_msgs/SpawnModel.h>
@@ -15,17 +17,15 @@
 #include <gazebo_msgs/ApplyBodyWrench.h>
 #include <gazebo_msgs/BodyRequest.h>
 
-#include "ow_dynamic_terrain/modified_terrain_diff.h"
+#include <ow_dynamic_terrain/modified_terrain_diff.h>
 
-#include "ServiceClientFacade.h"
+#include <ow_lander/DigLinearActionResult.h>
+#include <ow_lander/DigCircularActionResult.h>
+#include <ow_lander/DeliverActionResult.h>
+#include <ow_lander/DiscardActionResult.h>
 
-#include "ow_lander/DigLinearActionResult.h"
-#include "ow_lander/DigCircularActionResult.h"
-#include "ow_lander/DeliverActionResult.h"
-#include "ow_lander/DiscardActionResult.h"
-
-#include "ow_regolith/SpawnRegolithInScoop.h"
-#include "ow_regolith/RemoveAllRegolith.h"
+#include <ow_regolith/SpawnRegolithInScoop.h>
+#include <ow_regolith/RemoveAllRegolith.h>
 
 namespace ow_regolith {
 
--- a/ow_regolith/include/RegolithSpawner.h
+++ b/ow_regolith/include/RegolithSpawner.h
@@ -8,9 +8,16 @@
 #include <ros/ros.h>
 #include <tf/tf.h>
 
-#include "gazebo_msgs/LinkStates.h"
+#include <gazebo_msgs/LinkStates.h>
+
+#include <gazebo_msgs/SpawnModel.h>
+#include <gazebo_msgs/DeleteModel.h>
+#include <gazebo_msgs/ApplyBodyWrench.h>
+#include <gazebo_msgs/BodyRequest.h>
 
 #include "ow_dynamic_terrain/modified_terrain_diff.h"
+
+#include "PersistentServiceClient.h"
 
 #include "ow_lander/DigLinearActionResult.h"
 #include "ow_lander/DigCircularActionResult.h"
@@ -19,6 +26,8 @@
 
 #include "ow_regolith/SpawnRegolithInScoop.h"
 #include "ow_regolith/RemoveAllRegolith.h"
+
+namespace ow_regolith {
 
 class RegolithSpawner
 {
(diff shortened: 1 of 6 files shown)


ow_regolith/include/ServiceClientFacade.h
```diff
@@ -0,0 +1,80 @@
...
+
+template <class T>
+class ServiceClientFacade
+{
+public:
+  ServiceClientFacade() : m_node_handle(nullptr)
+  {
+    // do nothing
+  }
+
```
kmdalal, line 22: Good to 'default' or 'delete' the copy constructor, assignment operator, and destructor, as applicable, if you're not defining them.

ow_regolith/include/ServiceClientFacade.h
```diff
@@ -0,0 +1,80 @@
...
+  bool connect(std::shared_ptr<ros::NodeHandle> &nh, std::string service_path,
```
kmdalal, line 23: minor: service_path can be const&

ow_regolith/include/ServiceClientFacade.h
```diff
@@ -0,0 +1,80 @@
...
+
+#include <ros/ros.h>
+
+namespace ow_regolith {
+
+template <class T>
+class ServiceClientFacade
+{
+public:
+  ServiceClientFacade() : m_node_handle(nullptr)
```
kmdalal, line 18: Good to initialize m_persistent as well.

ow_regolith/include/RegolithSpawner.h
```diff
@@ -28,7 +37,7 @@ class RegolithSpawner
   RegolithSpawner(const RegolithSpawner&) = delete;
   RegolithSpawner& operator=(const RegolithSpawner&) = delete;
 
-  RegolithSpawner(ros::NodeHandle* nh);
+  RegolithSpawner(std::string node_name);
```
kmdalal, line 40: node_name could be const&

ow_regolith/include/RegolithSpawner.h
```diff
@@ -8,17 +8,26 @@
 #include <ros/ros.h>
 #include <tf/tf.h>
 
-#include "gazebo_msgs/LinkStates.h"
+#include <ServiceClientFacade.h>
```
kmdalal, line 11: Minor: #include "ServiceClientFacade.h"` since it's filed in your package.

ow_regolith/include/ServiceClientFacade.h
AstroStucky: Thank you for the consistent class method hygiene reminders! I'll catch onto this one eventually. :laughing:

ow_regolith/include/ServiceClientFacade.h
```diff
@@ -0,0 +1,84 @@
+// The Notices and Disclaimers for Ocean Worlds Autonomy Testbed for Exploration
+// Research and Simulation can be found in README.md in the root directory of
+// this repository.
+
```
kmdalal, line 4: A brief explanation of this class and its motivation would be helpful here.   I may try it in ow_plexil.

ow_regolith/include/ServiceClientFacade.h
```diff
@@ -0,0 +1,84 @@
...
+      return false;
+    }
+
+    if (!m_client.call(message)) {
+      ROS_ERROR("Failed to call service %s", m_client.getService().c_str());
+      return false;
+    }
+    return true;
+  }
+
```
kmdalal, line 64: The `connect` and `call` functions are rather long for inline functions.  The RegolithSpawner has just one Service but if there were many services (in this or another user), there would be some code bloat.   Suggest either un-inlining or explaining the rationale for inlining in a comment.

ow_regolith/include/ServiceClientFacade.h
AstroStucky: Class explanation added for both ServiceClientFacade and RegolithSpawner

ow_regolith/include/ServiceClientFacade.h
AstroStucky: Inlining was unnecessary. They are now implemented in the same header file below the class definition. The original reason for them being implemented in the same header file was because it was a template class; however, in our conversation this morning we both realized `ServiceClientFacade` need not be a template class at all, only its member functions need to be. This is now the case. However, I decided to keep member function implementation inside the header file because this makes the class more portable to other software packages, and you expressed interest in using it in **ow_autonomy**.

Making `ServiceClientFacade` a non-template class has also made the class more useful since the message type no longer needs to be decided upon its instantiation, but instead can be decided or even changed for a single instance each time the connect method is called. This allows a single `ServiceClientFacade` to handle connections to multiple different ROS Services, so long as they are not concurrent connections.

kmdalal (approved):
Looks good!  I'll try out ServiceClientFacade in ow_plexil after this is merged.

ow_regolith/include/ServiceClientFacade.h
```diff
@@ -0,0 +1,117 @@
...
+  if (!m_client.call(message)) {
+    ROS_ERROR("Failed to call service %s", m_client.getService().c_str());
+    return false;
+  }
+  return true;
+};
+
+template <typename T>
+bool ServiceClientFacade::attempt_reconnect()
```
mogumbo, line 108: Should be called attemptReconnect per our coding guidelines.

ow_regolith/include/ServiceClientFacade.h
```diff
@@ -0,0 +1,117 @@
...
+namespace ow_regolith {
+
+class ServiceClientFacade
+{
+public:
+  ServiceClientFacade() : m_node_handle(nullptr), m_persistent(false)
+  {
+    // do nothing
+  }
+  ServiceClientFacade(const ServiceClientFacade&) = delete;
```
mogumbo, line 24: Please comment on why the copy constructor and assignment operator are explicitly prohibited, rather than just not implementing them.

mogumbo (changes requested):

ow_regolith/include/ServiceClientFacade.h
kmdalal: This was in response to my comment when these weren't there.  I believe in automatically generated methods being implemented, defaulted, or deleted, i.e. being explicit about the semantics here.

The pre-C++-11 way to disallow copy, etc was to declare these private and then not implement them.

In this case I'm not sure if copying needs to be prohibited, but it doesn't matter until a use case arises.

AstroStucky: The underline reason to delete both copy and assignment is that they both require copying an instance, and supporting that safely would require testing of the ROS service API. Because I cannot think of a use case for this anyways, it will go untested for now. We can always add this feature down the line if it's needed.

mogumbo (approved):

PR #221, opened 2022-02-02
Oceanwater 898 Fix ow_lander actions failing if called too early

AstroStucky:
## Linked Issues:
| EPIC ⚡ | N/A |
| Jira Ticket 🎟️ | [OCEANWATER-898](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-898) |
| Github :octocat:  | # |

Fixes a bug whereby any of the **ow_lander** action servers could fail if they were initialized while an action client was already waiting for them to exist. 
Each action server was passing a member function to SimpleActionClient, which spun-off a thread that could call that member function before the action server's `__init__` function finished initializing the remainder of the action server's member variables. This appears to only occur if there was an action client already waiting for the action server to exist.

## Summary of Changes
* SimpleActionClient is now called at the very end of each action server's `__init__` function, ensuring all member variables are initialized before any other member function is called.

## Test
The bug could be consistently replicated by calling a `*_action_client.py` script before calling `roslaunch` to launch a world. Each action defined in **ow_lander** has already been tested under these conditions, and they executed fine.

It is sufficient to perform the following procedure on several different **ow_lander** actions:
1. Call any of the `*_action_client.py` scripts. You will see a message saying that the master node is not running yet, this is just a warning and can be ignored.
2. In another terminal call `roslaunch ow atacama_y1a use_rviz:=false rqt_gui:=false`, or use another world if you prefer.
3. Ensure the action executes fine after the world is booted, and no `AttributeError` is thrown by its server.

### DIFF
--- a/ow_lander/scripts/antenna_pan_tilt_action_server.py
+++ b/ow_lander/scripts/antenna_pan_tilt_action_server.py
@@ -19,9 +19,6 @@ class AntennaPanTiltActionServer(object):
 
     def __init__(self, name):
         self._action_name = name
-        self._server = actionlib.SimpleActionServer(
-            self._action_name, ow_lander.msg.AntennaPanTiltAction, execute_cb=self.on_antenna_action, auto_start=False)
-        self._server.start()
         # Action Feedback/Result
         self._fdbk = ow_lander.msg.AntennaPanTiltFeedback()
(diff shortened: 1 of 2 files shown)


anjan1984 (approved):
Looks good to me

PR #222, opened 2022-02-02
Oceanwater 891 implement ros action stop

anjan1984:
## Linked Issues:

| Jira Ticket 🎟️ | [OCEANWATER-XXX](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-891) |
| Github :octocat:  | # |

## Summary of Changes
* Adding Arm Stop action 

## Test 
In 3 seperate terminlas. 

launch any of the action files: 
* python3 unstow_action_client.py 

Stop the motion during the movement. 
* python3 stop_action_client.py


ow_lander/scripts/arm_action_servers.py
```diff
@@ -175,6 +171,34 @@ def now_from_start(start):
...
 
+class StopActionServer(object):
+
+    def __init__(self, name):
+        self._action_name = name
+        self._server = actionlib.SimpleActionServer(self._action_name,
+                                                    ow_lander.msg.StopAction,
+                                                    execute_cb=self.on_stop_action,
+                                                    auto_start=False)
+        self._server.start()
```
AstroStucky, line 182: Please move these two lines to the end of `__init__` to avoid the bug that's addressed by #221

ow_lander/CMakeLists.txt
```diff
@@ -38,6 +38,7 @@ add_action_files(
   FILES GuardedMove.action
   FILES AntennaPanTilt.action
   FILES Discard.action
+  FILES Stop.action
```
kmdalal, line 41: I think a more specific name than "Stop" would be good, but let's hold off on renaming until we address naming across the whole OW/OWLAT interface.

ow_lander/action/Stop.action
```diff
@@ -0,0 +1,8 @@
+#goal definition
+std_msgs/Empty goal
+---
+# result
+geometry_msgs/Point final     # arm final position
+---
+# feedback
+geometry_msgs/Point current   # arm current position
```
kmdalal, line 8: OWLAT's ARM_STOP has no result or feedback.  We should follow suit, or argue that these are useful.  If the arm's current position is available in telemetry, it may not be so useful here.

kmdalal:
I just committed an executable version of the client script.  I also accidentally committed the Saturn ephemeris hotfix, since I needed it to build this branch, and undid that with two more commits.

kmdalal:
I did many tests stopping the arm during both Stow and Unstow, and it works well.  However, if one calls Stop when the arm is at rest in between operations, the following error appears:

`[ERROR]: Received comm state PREEMPTING when in simple state DONE with SimpleActionClient in NS /arm_controller/follow_joint_trajectory`

And when I called it when the arm is at rest in mid-operation (i.e. after stopping it once), I get this error:

`[ERROR]: Received comm state RECALLING when in simple state DONE with SimpleActionClient in NS /arm_controller/follow_joint_trajectory`

kmdalal:
I think two reviewers is probably sufficient for this change.  @Samahu, unless you want to review you can take yourself off since Thomas and I have been on it.

AstroStucky:
> I think two reviewers is probably sufficient for this change. @Samahu, unless you want to review you can take yourself off since Thomas and I have been on it.

Either way works for me. I should have mentioned in my code comment that I was not intending to do a full review. I just needed to make one comment that relates to a PR #221, but the only way to do that on GitHub is to submit a review. I'll still need the chance to look over the remainder of the code and perform tests.

AstroStucky:
Launched _atacama_y1a_. Called the `unstow_action_client.py` script, followed by the `stop_action_client.py` in another terminal. The arm stopped mid-unstow as intended, and both action client scripts terminated. I then called `grind_action_client.py` and it produced the following error. 
```
[ERROR] [1643918124.228572445, 1916524961.019999981]: Can't accept new action goals. Controller is not running.
```

AstroStucky:
> Launched _atacama_y1a_. Called the `unstow_action_client.py` script, followed by the `stop_action_client.py` in another terminal. The arm stopped mid-unstow as intended, and both action client scripts terminated. I then called `grind_action_client.py`. Nothing happened for a few seconds, until the following error message came up in the terminal running roslaunch
> 
> ```
> [ERROR] [1643918124.228572445, 1916524961.019999981]: Can't accept new action goals. Controller is not running.
> ```
> 
> Then a few seconds later the script terminated.

Calling `dig_linear_action_client.py` after all that worked just fine. Maybe only some actions cause this error :thinking:

ow_lander/scripts/arm_action_servers.py
```diff
@@ -67,10 +67,10 @@ def on_unstow_action(self, goal):
             return
         success = False
```
AstroStucky, line 69: Lines 31 and 32 defining the trajectory_async_executer member variable should be removed as it was for the other action servers.

AstroStucky: Same for the stow action server: lines 110 and 111 should be removed.

AstroStucky (changes requested):
Some member variables were left behind UnstowActionServer and StowActionServer that should be cleaned up.

ow_lander/scripts/arm_action_servers.py
anjan1984: @AstroStucky thanks for the review. Some of the problems you are seeing happened as I merged noetic devel in this branch. Since there were changes in this file from your changes some parts if the file were over written. Will work on it next week

anjan1984:
> > Launched _atacama_y1a_. Called the `unstow_action_client.py` script, followed by the `stop_action_client.py` in another terminal. The arm stopped mid-unstow as intended, and both action client scripts terminated. I then called `grind_action_client.py`. Nothing happened for a few seconds, until the following error message came up in the terminal running roslaunch
> > ```
> > [ERROR] [1643918124.228572445, 1916524961.019999981]: Can't accept new action goals. Controller is not running.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Then a few seconds later the script terminated.
> 
> Calling `dig_linear_action_client.py` after all that worked just fine. Maybe only some actions cause this error thinking

Thank you for pointing this out. I have addressed this

anjan1984:
> I did many tests stopping the arm during both Stow and Unstow, and it works well. However, if one calls Stop when the arm is at rest in between operations, the following error appears:
> 
> `[ERROR]: Received comm state PREEMPTING when in simple state DONE with SimpleActionClient in NS /arm_controller/follow_joint_trajectory`
> 
> And when I called it when the arm is at rest in mid-operation (i.e. after stopping it once), I get this error:
> 
> `[ERROR]: Received comm state RECALLING when in simple state DONE with SimpleActionClient in NS /arm_controller/follow_joint_trajectory`

Updates have been made to address this

kmdalal:
I've tested a lot, stopping all the arm operations.  In one case when stopping a DigCircular, I got this error:
```
Exception in thread /Stop/result:
  File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/opt/ros/noetic/lib/python3/dist-packages/rospy/impl/tcpros_pubsub.py", line 185, in robust_connect_subscriber
    conn.receive_loop(receive_cb)           
  File "/opt/ros/noetic/lib/python3/dist-packages/rospy/impl/tcpros_base.py", line 846, in receive_loop
    self.close()
  File "/opt/ros/noetic/lib/python3/dist-packages/rospy/impl/tcpros_base.py", line 858, in close
    self.socket.close()
AttributeError: 'NoneType' object has no attribute 'close'
```

kmdalal:
On several occasions, the arm does not stop moving.  It seems to depend on where in the operation the stop command is issued.  I can't yet characterize it.

AstroStucky:
The error I encountered earlier is not resolved. 

Similar to my last test, with _atacama_y1a_ opened I called `unstow_action_client.py`, and it stopped fine when `stop_action_client.py` was called. I then called `unstow_action_client.py` and it resumed fine and completed the action. I then continued normal operation and tried to call `grind_action_client.py`, nothing happened for a while, then I saw the same error.
```
[ERROR] [1645219350.801963890, 1916524899.075000048]: Can't accept new action goals. Controller is not running.
```
Calling other action clients at this point result in the same error, but for whatever reason `dig_linear_action_client.py` is still unaffected by it, and can be called fine.

AstroStucky:
> On several occasions, the arm does not stop moving. It seems to depend on where in the operation the stop command is issued. I can't yet characterize it.

I can verify this. In my case it would not stop when when the grinder was moving through the terrain, but does stop if stop action is called before that phase of the operation. 

> I've tested a lot, stopping all the arm operations. In one case when stopping a DigCircular, I got this error:
> 
> ```
> Exception in thread /Stop/result:
> Traceback (most recent call last):
>   File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
>     self.run()
>   File "/usr/lib/python3.8/threading.py", line 870, in run
>     self._target(*self._args, **self._kwargs)
>   File "/opt/ros/noetic/lib/python3/dist-packages/rospy/impl/tcpros_pubsub.py", line 185, in robust_connect_subscriber
>     conn.receive_loop(receive_cb)           
>   File "/opt/ros/noetic/lib/python3/dist-packages/rospy/impl/tcpros_base.py", line 846, in receive_loop
>     self.close()
>   File "/opt/ros/noetic/lib/python3/dist-packages/rospy/impl/tcpros_base.py", line 858, in close
>     self.socket.close()
> AttributeError: 'NoneType' object has no attribute 'close'
> ```

I also saw this error occur once. It was thrown by `stop_action_client.py`. The arm still managed to stop despite the error.

anjan1984:
Please review the stop actions. The error still happens sometime. I think it is best to open a separate ticket to handle the threading error.

AstroStucky:
During testing I had one instance where the stop action did not stop a dig linear in the middle of the linear portion, but I have not been able to replicate this behavior, so I'm going to call it a pass. No errors occurred including the one thrown by threading.py.

I just want to note that the stop action does not prevent an action from starting if it's called in the planning phase of the action (after the action client is called, but before the arm starts moving). I don't think this is a problem, since stopping an action in the planning phase does not seem all that useful, except maybe in the case a user accidentally sends an action when they didn't intend to.

ow_lander/scripts/trajectory_async_execution.py
```diff
@@ -11,114 +11,111 @@
...
+        goal.goal_time_tolerance = self._goal_time_tolerance
+        goal.trajectory = trajectory
+        self._client.send_goal(goal, done_cb, active_cb, feedback_cb)
+
+    def stop(self):
+        """
+        Stops the execution of the last trajectory submitted for executoin
+        """
+        goal_active = self.is_active()
+        if self._connected and goal_active == 1:
```
AstroStucky, line 94:
```suggestion
        if self._connected and goal_active == GoalStatus.ACTIVE:
```

AstroStucky (approved):
Test worked. Just had one code comment. Well done!

kmdalal:
I've tried several stows and unstows where I stop the action immediately, i.e. during planning.  I expected the real arm to still move, assuming that a stop during planning is ignored.  However, the real arm did _not_ move, as one would desire.  But also (not desired):

- the planning motion continued to completion in Rviz
- the stow/unstow action scripts returned with the final position of the arm as though it had moved

kmdalal:
My last comment pointed out a problem in what is printed by the client scripts when they finish, after a stop was given.

A more general issue: can the client scripts know that the action was stopped and convey that when they return?  (This could be a follow-on Jira).

kmdalal:
Regarding the problem of not being able to _stop_ planning, the only idea I have is to make planning a process rather than a blocking thread, so that it can be killed.  (Follow-on Jira).  That said, I'm not fluent in Python techniques.

kmdalal (approved):
This is working pretty well per the limitations we've discussed in this ticket and in our meetings.  Improvements are follow-on work.

PR #225, opened 2022-03-09
Oceanwater 907 stop planning using stop action

anjan1984:
## Linked Issues:
| EPIC ⚡ | [OCEANWATER-907_Stop_planning_using_stop_action
](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-907) |
| Jira Ticket 🎟️ | [OCEANWATER-XXX](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-XXX) |
| Github :octocat:  | # |

## Summary of Changes
* Adding Arm Stop action function during the planning phase. Now the stop arm action can immediately stop the arm during both the planning and execution phase. 

## Test
* In 3 seperate terminlas.

launch any of the action files:
./grind_action_client.py
-you can check for all other action clients like dig_circular, dig_linear etc... 

Stop the motion during the planning phase itself movement.
./stop_action_client.py


ow_lander/scripts/action_trajectories.py
```diff
@@ -0,0 +1,1033 @@
...
+        if (is_shou_yaw_goal_in_range(joint_goal) == False):
+            return False
+
+        joint_goal[constants.J_SCOOP_YAW] = 0
+
+        move_arm.set_joint_value_target(joint_goal)
+        _, plan, _, _ = move_arm.plan()
+        return plan
+
+    def dig_circular(self, move_arm, move_limbs, robot, moveit_fk, args, server_stop):
```
kmdalal, line 218: Is it possible to access server_stop without passing it in?  E.g. by importing it from its source file?

ow_lander/scripts/action_trajectories.py
```diff
@@ -0,0 +1,1033 @@
...
+        y_start = args.y_start
+        depth = args.depth
+        parallel = args.parallel
+        ground_position = args.ground_position
+
+        if not parallel:
+
+            server_stop.get_state()
+            if server_stop.stopped is True:
+                return False
```
kmdalal, line 239: These three lines are repeated many times.  Three suggestions:

1. Factor this out into a boolean-valued utility function.
2. Issue an Info message when the operation is stopped, otherwise hard to distinguish from normal completion in the output.
3. Minor, and I'm not a Python expert: I think `==` is more appropriate than `is' here, and `not ...` reads even better.

Samahu: Actually it would be totally fine (and completely readable) to just use:
```suggestion
            server_stop.get_state()
            if server_stop.stopped:
                return False
```

kmdalal:
I have tested Stop many times on several operations in various stages of planning and execution.  There was one case where Stow did not stop during execution, but I can't reproduce it.

Left some minor code comments.  I'll approve on my end and leave this in the hands of the more experienced Python folks.

kmdalal:
Got this error from GuardedMove upon stopping it.
```
[ERROR]: Exception in your execute callback: cannot unpack non-iterable bool object                                        
 File "/opt/ros/noetic/lib/python3/dist-packages/actionlib/simple_action_server.py", line 289, in executeLoop             
   self.execute_callback(goal)                                                                                            
 File "/usr/local/home/kdalal/ow/src/ow_simulator/ow_lander/scripts/arm_action_servers.py", line 410, in on_guarded_move_
action                                                                                                                     
   self._update_motion(goal)                                                                                              
 File "/usr/local/home/kdalal/ow/src/ow_simulator/ow_lander/scripts/arm_action_servers.py", line 395, in _update_motion   
   self.guarded_move_traj, self._guarded_move_plan_ratio = action_trajectories.guarded_move_plan(                         
TypeError: cannot unpack non-iterable bool object            
```

anjan1984:
> 



> Got this error from GuardedMove upon stopping it.

New commit should get rid of the GuardedMove error

ow_lander/scripts/action_trajectories.py
AstroStucky: I agree the repetition of this statement is not ideal, but I think I have a better way to address it, which I think would be better covered in another PR. It involves some changes to arm_action_servers.py, so I would be happy to implement it myself as part of my refactor ticket, https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-904. My proposal is as follows... 

As best as I can tell the `server_stop.stopped` check has to occur before any call to `move_arm.plan()`. So rather than calling a function each time before the call to `plan()`, I suggest we wrap the `plan()` function in an `*ActionServer` version of the `plan()` function that does two things: **1)** Checks the state of `server_stop` and throws and exception if it is stopped, and **2)** calls the `MoveGroupCommander`'s version of `plan()` returning its return values. Then all of these large blocks of movement code simply have to be put inside of a try-catch block, and the case where stop is called is handled inside the catch.

AstroStucky: I agree the repetition of this statement is not ideal, but I think I have a better way to address it, which I think would be better covered in another PR. It involves some changes to arm_action_servers.py, so I would be happy to implement it myself as part of my refactor ticket, https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-904. My proposal is as follows... 

As best as I can tell the `server_stop.stopped` check has to occur before any call to `move_arm.plan()`. So rather than calling a function each time before the call to `plan()`, I suggest we wrap the `plan()` function in an `*ActionServer` version of the `plan()` function that does two things: **1)** Checks the state of `server_stop` and throws and exception if it is stopped, and **2)** calls the `MoveGroupCommander`'s version of `plan()` returning its return values. Then all of these large blocks of movement code simply have to be put inside of a try-catch block, and the case where stop is called is handled inside the catch.

ow_lander/scripts/action_trajectories.py
AstroStucky: I have a proposal in your other comment thread that would make this unnecessary, but it would be best implemented in another PR.

ow_lander/scripts/action_trajectories.py
```diff
@@ -0,0 +1,1033 @@
...
+        :type move_arm: class 'moveit_commander.move_group.MoveGroupCommander'
+        :type moveit_fk: class moveit_msgs/GetPositionFK
+        '''
+        #joint_names: [j_shou_yaw, j_shou_pitch, j_prox_pitch, j_dist_pitch, j_hand_yaw, j_scoop_yaw]
+        # robot full state name: [j_ant_pan, j_ant_tilt, j_shou_yaw, j_shou_pitch, j_prox_pitch, j_dist_pitch, j_hand_yaw,
+        # j_grinder, j_scoop_yaw]
+
+        # get joint states from the end of the plan
+        joint_states = plan.joint_trajectory.points[len(
+            plan.joint_trajectory.points)-1].positions
```
AstroStucky, line 668:
```suggestion
        joint_states = plan.joint_trajectory.points[-1].positions
```

ow_lander/scripts/action_trajectories.py
```diff
@@ -0,0 +1,1033 @@
...
+        # construct robot state at the end of the plan
+        robot_state = robot.get_current_state()
+        # adding antenna (0,0) and j_scoop_yaw (0.1) which should not change
+        new_value = (0, 0) + joint_states[:6] + (0.1740,)
+        # modify current state of robot to the end state of the previous plan
+        robot_state.joint_state.position = new_value
```
AstroStucky, line 674:
```suggestion
        # modify current state of robot to the end state of the previous plan
        robot_state.joint_state.position = (0, 0) + joint_states[:6] + (0.1740,)
```

ow_lander/scripts/action_trajectories.py
```diff
@@ -0,0 +1,1033 @@
...
+        # Move to pre move position, align shoulder yaw
+        joint_goal = move_arm.get_current_joint_values()
+        joint_goal[constants.J_DIST_PITCH] = 0
+        joint_goal[constants.J_HAND_YAW] = 0
+        joint_goal[constants.J_PROX_PITCH] = -math.pi/2
+        joint_goal[constants.J_SHOU_PITCH] = math.pi/2
+        joint_goal[constants.J_SHOU_YAW] = alpha + beta
+
+        # If out of joint range, abort
+        if (is_shou_yaw_goal_in_range(joint_goal) == False):
```
AstroStucky, line 857:
```suggestion
        if not is_shou_yaw_goal_in_range(joint_goal):
```

ow_lander/scripts/action_trajectories.py
```diff
@@ -0,0 +1,1033 @@
...
+            return False, False
+
+        _, plan_c, _, _ = move_arm.plan()
+
+        guarded_move_traj = self.cascade_plans(pre_guarded_move_traj, plan_c)
+
+        pre_guarded_move_end_time = pre_guarded_move_traj.joint_trajectory.points[len(
+            pre_guarded_move_traj.joint_trajectory.points)-1].time_from_start
+        guarded_move_end_time = guarded_move_traj.joint_trajectory.points[len(
+            guarded_move_traj.joint_trajectory.points)-1].time_from_start
```
AstroStucky, line 918:
```suggestion
        pre_guarded_move_end_time = pre_guarded_move_traj.joint_trajectory.points[-1].time_from_start
        guarded_move_end_time = guarded_move_traj.joint_trajectory.points[-1].time_from_start
```

ow_lander/scripts/action_trajectories.py
```diff
@@ -0,0 +1,1033 @@
...
+        robot_state = robot.get_current_state()
+        move_arm.set_start_state(robot_state)
+        x_discard = args.discard.x
+        y_discard = args.discard.y
+        z_discard = args.discard.z
+
+        # after sample collect
+        mypi = 3.14159
+        d2r = mypi/180
+        r2d = 180/mypi
```
AstroStucky, line 939:
```suggestion
        d2r = math.pi / 180
        r2d = 180 / math.pi
```

ow_lander/scripts/action_trajectories.py
anjan1984: made the requested change

anjan1984:
@AstroStucky  committed all the suggested changes. I really like your suggestion about the refactoring of the code.

kmdalal:
GuardedMove is always returning True, whether it succeeds, is stopped during planning, or stopped during execution.  I see no way to distinguish a completed from a stopped action.

ow_lander/scripts/action_trajectories.py
```diff
@@ -0,0 +1,1026 @@
...
+
+
+class ActionTrajectories:
+    '''
+    Computes all the trajectories used by the lander
+    '''
+
+    def __init__(self):
+        self.stopped = False
+        self.complete_trajectory = None
```
kmdalal, line 30: Are these used?

ow_lander/scripts/action_trajectories.py
```diff
@@ -0,0 +1,1026 @@
...
+        circ_traj = RobotTrajectory()
+
+        x_start = args.x_start
+        y_start = args.y_start
+        depth = args.depth
+        parallel = args.parallel
+        ground_position = args.ground_position
+        server_stop.get_state()
+        if server_stop.stopped:
+                return False
```
kmdalal, line 233: This is a little improvement over the previous code, but unless @AstroStucky 's refactor (OCEANWATER-904) is going to happen soon, I still think we should issue a ROS message when an action is stopped.  Instead of repeating that line many times, the simple refactor I proposed last time would be easy, e.g. a function like:
```
 def check_for_stop(self, action_name, server_stop):
        server_stop.get_state()
        if server_stop.stopped:
            rospy.loginfo('%s was stopped.', action_name);
            return True
        return False
````
and calling it instead, e.g.
``` 
if self.check_for_stop("dig_circular", server_stop):
            return False
```

**Comment by kmdalal** [comment]
For complete functionality I think the action server should issue a message when an action is stopped (simple approach suggested in code comment), and so should the action client.  At present there is no easy way to distinguish a completed action from a stopped one except by watching the graphics (or _knowing_ that it was stopped!).

**Review by AstroStucky — COMMENTED** [review]
I think Michael's suggested refactor of the stopped check should be implemented. I agree some kind of message or log notification should occur to acknowledge a stop request.

File: ow_lander/scripts/action_trajectories.py (hunk 8 above)
Comment (anjan1984) [reply ow_lander/scripts/action_trajectories.py:30]: self.complete_trajectory = None removed

File: ow_lander/scripts/action_trajectories.py [hunk 10]
```diff
@@ -0,0 +1,1033 @@
...
+        inputs:  current plan, robot, arm interface, and moveit forward kinematics object
+        outputs: goal_pose, robot state and joint states at end of the plan
+        '''
+        #joint_names: [j_shou_yaw, j_shou_pitch, j_prox_pitch, j_dist_pitch, j_hand_yaw, j_scoop_yaw]
+        # robot full state name: [j_ant_pan, j_ant_tilt, j_shou_yaw, j_shou_pitch, j_prox_pitch, j_dist_pitch, j_hand_yaw,
+        # j_grinder, j_scoop_yaw]
+
+        # get joint states from the end of the plan
+        joint_states = plan.joint_trajectory.points[len(
+            plan.joint_trajectory.points)-1].positions
```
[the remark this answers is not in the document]
anjan1984: Done!

ow_lander/scripts/arm_action_servers.py
```diff
@@ -207,6 +211,7 @@ def on_stop_action(self, goal):
         trajectory_async_executer.stop()
 
         rospy.loginfo('%s: Succeeded' % self._action_name)
```
kmdalal, line 213: I would think that if an action is stopped, it does not succeed.  Thoughts?

ow_lander/scripts/arm_action_servers.py
```diff
@@ -207,6 +211,7 @@ def on_stop_action(self, goal):
...
+        rospy.loginfo('Current Arm Action was stopped.')
```
kmdalal, line 214: Since there's also an action-specific "was stopped" message, this one seems redundant.

kmdalal:
In my browser, the code view of action_trajectories.py shows resolved conversations, but is still showing the code prior to the resolution and not after.   I've never seen this happen.  Is anyone else seeing it?

kmdalal:
I retested the stopping of most arm actions and all behaves well, though there is a design issue remaining.  Currently, when an action is stopped it is still marked Succeeded, but I think it should Fail, and that an action failure should be reported as a ROS message in the action client scripts....

ow_lander/scripts/arm_action_servers.py
anjan1984: This is the stop action succeeding

ow_lander/scripts/arm_action_servers.py
anjan1984: action specific stop could only be issued during planning phase... During execution all the plans use the same trajectory_async_executer. I added this to clarify that the current action was stopped and Stop Action succeeded.

anjan1984:
> I retested the stopping of most arm actions and all behaves well, though there is a design issue remaining. Currently, when an action is stopped it is still marked Succeeded, but I think it should Fail, and that an action failure should be reported as a ROS message in the action client scripts....

In my code I set the action server as "aborted". I can further investigate why it still shows succeeded.

AstroStucky (approved):
Recent code changes implemented suggestions and they look good to me. I did not retest, but two testers did not look necessary for these commits. Approved! :confetti_ball:

ow_lander/scripts/arm_action_servers.py
kmdalal: Oops, my bad :)

ow_lander/scripts/arm_action_servers.py
kmdalal: Got it.  It may be useful to add "during execution" to this message, and "during planning" in the specific messages.

kmdalal:
> > I retested the stopping of most arm actions and all behaves well, though there is a design issue remaining. Currently, when an action is stopped it is still marked Succeeded, but I think it should Fail, and that an action failure should be reported as a ROS message in the action client scripts....
> 
> In my code I set the action server as "aborted". I can further investigate why it still shows succeeded.

Got it.  Then the problem is that the action clients don't check the state, they just print the result.   If an action was aborted, the client should know and report that -- _instead_ of a result I would think.

anjan1984:
> > > I retested the stopping of most arm actions and all behaves well, though there is a design issue remaining. Currently, when an action is stopped it is still marked Succeeded, but I think it should Fail, and that an action failure should be reported as a ROS message in the action client scripts....
> > 
> > 
> > In my code I set the action server as "aborted". I can further investigate why it still shows succeeded.
> 
> Got it. Then the problem is that the action clients don't check the state, they just print the result. If an action was aborted, the client should know and report that -- _instead_ of a result I would think.

Upon investigating further I think it will be best handled in a separate Jira ... This involves the state of the in-built joint trajectory action.

ow_lander/scripts/arm_action_servers.py
anjan1984: > Got it. It may be useful to add "during execution" to this message, and "during planning" in the specific messages.

The problem is during Stow and Unstow there is no separate planning and execution phase.

kmdalal (approved):
Left a small suggestion for improving the stop message.  For a more polished package, please create a Jira for action clients to detect and report when their action was stopped.  Otherwise, looks good!

PR #239, opened 2022-05-10
Update to Plexil 4.6

kmdalal:
The change is only an update to the installation instructions for Plexil.

The test is to update your Plexil installation to 4.6 (see my email "Please update PLEXIL" of 5/10) and then run the ReferenceMission1 plan.  It should behave as before.

### DIFF
--- a/oceanwaters/doc/setup_dev_env.md
+++ b/oceanwaters/doc/setup_dev_env.md
@@ -61,14 +61,14 @@ git clone --branch releases/plexil-4.6 https://git.code.sf.net/p/plexil/git plex
 ```
 
 NOTE: OceanWATERS has been tested with a version of this branch tagged
-`OceanWATERS-v9.1`.  Its git commit hash begins with `c54173c4`.  Newer
+`OceanWATERS-v10.0`.  Its git commit hash begins with `cde7d07`.  Newer
 versions of this branch should work with the newest version of the
 `master` branches of OceanWATERS, but you can revert your PLEXIL to
 this tag if there are problems:
 
 ```
 cd $PLEXIL_HOME
-git checkout OceanWATERS-v9.1
+git checkout OceanWATERS-v10.0
 make squeaky-clean
 make universalExec plexil-compiler checkpoint
 ```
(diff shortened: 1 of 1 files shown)


AstroStucky:
Running ReferenceMission1.plx I get the following failure
```
[ INFO] [1652893103.705607489, 1916524867.602499962]: IdentifySampleLocation started...
[ INFO] [1652893103.930887236, 1916524867.827500105]: Possible sample location identified at (1.778633, 0.203339, -0.219355)
[ERROR] [1652893104.007181659, 1916524867.904999971]: PLEXIL: Failed to find ground, aborting.
[ INFO] [1652893104.007277613, 1916524867.904999971]: PLEXIL: Reference Mission 1, Sol 0 complete.
```

AstroStucky:
> 

The failure repeated after a second run. Could it have to do with the guarded move bug described in https://github.com/nasa/ow_simulator/pull/241?

AstroStucky:
FYI, I ran ReferenceMission1.plx in the atacama_y1a world each attempt.

kmdalal:
> The failure repeated after a second run. Could it have to do with the guarded move bug described in #241?

No, and the message is misleading: it does not even do a GuardedMove because it couldn't find a good sample location.  I will change the message.

And yes, ReferenceMission1 is designed for the Atacama world only.   I'll make sure this is documented.

AstroStucky (changes requested):
Unable to complete the test at present. Will await a work around or fix.

kmdalal:
Sure enough, Plexil 4.6 seems to be the cause of ReferenceMission1 breaking.  Because I noticed this problem around the same time as the GuardedMove bug, I thought it was a case of the latter, especially because of PLEXIL's misleading message. 

I'm creating a fix branch on ow_autonomy for this issue, and suspending this PR for now.

kmdalal:
Please re-review this, concurrently with https://github.com/nasa/ow_autonomy/pull/81
That PR has a test to be tried before and after updating your Plexil installation.  So the order to satisfy both PRs is:
1. Do the first autonomy PR test with your current Plexil install.
2. Update your Plexil install (per instructions in last week's email).
3. Do the same test again.

AstroStucky (approved):
Performed the tests described in [ow_autonomy - PR#81](https://github.com/nasa/ow_autonomy/pull/81), and they passed. Approved!

PR #240, opened 2022-05-11
changing planner to RRT* for consistent deliver and discard behavior

anjan1984:
## Linked Issues:
| Jira Ticket 🎟️ | [OCEANWATER-932](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-932) |
| Github :octocat:  | # |

## Summary of Changes
* Changed default planner for discard and deliver to RRT* for constant behavior

## Test
*roslaunch ow atacama_y1a.launch 

In a separate terminal :
*./grind_action_client.py 
* ./dig_circular_action_client.py  (you can also use the dig_linear service)
* ./discard_action_client.py
*  ./dig_circular_action_client.py (you can also use the dig_linear service)
* ./deliver_action_client.py

### DIFF
--- a/ow_lander/scripts/action_trajectories.py
+++ b/ow_lander/scripts/action_trajectories.py
@@ -930,7 +930,7 @@ class ActionTrajectories:
         goal_pose.orientation = Quaternion(q[0], q[1], q[2], q[3])
 
         move_arm.set_pose_target(goal_pose)
-
+        move_arm.set_planner_id("RRTstar")
         if self.check_for_stop("discard_sample", server_stop):
             return False
         _, plan_a, _, _ = move_arm.plan()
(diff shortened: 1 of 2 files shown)


kmdalal:
This fixes the problem, but takes us back to where we were, with the much slower planner.   Is it possible to add constraints in the other planner, such as that the scoop never tilts sideways?  This might correct the bad movements.

anjan1984:
> This fixes the problem, but takes us back to where we were, with the much slower planner. Is it possible to add constraints in the other planner, such as that the scoop never tilts sideways? This might correct the bad movements.

I had tried to use constrained planning before... Had mixed success with it. RRT* gave much robust solutions.

kmdalal (approved):
Per discussion today, performance improvement can be future work.

AstroStucky:
The additional time RRT* requires for planning had to be accounted for in the deliver and discard max durations in the terrain_interaction test, so I went ahead and added a commit for that. This change requires no additional testing from reviewers.

AstroStucky:
The prescribed test passes for me, and I performed the terrain_interaction test 3 times with no failures.

PR #243, opened 2022-05-20
Oceanwater 918 ingest sample action

AstroStucky:
## Linked Issues:
| EPIC ⚡ | |
| Jira Ticket 🎟️ | [OCEANWATER-918](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-918) |
| Github :octocat:  | # |

## Summary of Changes
* Added ingest sample action to ow_lander, which allows users to call an action that removes any regolith particles inside the sample dock. 
    * **NOTE** The actual implementation removes any regolith contacting the sample dock, and terminates after 3.0 seconds without new regolith contacts. 
* ContactSensorPlugin added to the lander's sample dock to facilitate detection of particles inside it.
* Included the ingest sample action in the terrain_interaction test.
    * **NOTE** This test is marked expected to fail for now due to regolith delivery not always staying in the dock.
* Fixed a bug in ow_regolith that would sometimes cause all sample to be removed when none should be.
* Moved the sample dock slightly closer to the lander body.

## Test
If you want to use the terrain_interaction test for this, feel free, but you will have to watch it and verify regolith is removed from the dock at the very end as the output for the final unit test cannot be trusted. 

Alternatively, you can do this...
1. Launch a world of your choice.
2. Call `/src/ow_simulator/ow_lander/scripts/dock_ingest_sample_action_client.py`
3. Nothing should happen, but after a brief moment the action client should output
```
Result: sample_ingested: False
```
5. Call `rosservice call /ow_regolith/spawn_regolith '[0.55, -0.4, 0.8]' 'lander::base_link'` as many times as you'd like.
6. Call `rosservice call /ow_regolith/spawn_regolith '[0.4, -0.4, 0.8]' 'lander::base_link'` as many times as you'd like.
7. There should now be several regolith particles in both the sample dock and on the lander body.
8. Call `./src/ow_simulator/ow_lander/scripts/dock_ingest_sample_action_client.py`
9. Verify that the particles in the sample dock have been removed, but the ones on the lander body remain, and the action client outputs
```
Result: sample_ingested: True
```


AstroStucky:
Sorry for the long chain of commits. This branch was originally based on OCEANWATER-918, so it includes redundant commits that are already in noetic-devel. I'm open to suggestions on how to clean this up or if we should even bother.

kmdalal (approved):
I tried the manual test several times, worked well.  Gave the code a light review, mainly for readability.  Nice new feature!

anjan1984 (approved):
default_gzclient_camera(1)-2022-05-24T16_02_15 285682


Great work.... This looks a little odd though !

AstroStucky:
> default_gzclient_camera(1)-2022-05-24T16_02_15 285682
> 
> Great work.... This looks a little odd though !

The lander just likes to show off its balance :laughing: 

Thanks for the speedy reviews!

PR #248, opened 2022-06-29
Oceanwater 894 restrict power faults to one per simulation

Lumsterling:
## Linked Issues:
| EPIC ⚡ | [OCEANWATER-894](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-894) |
| Jira Ticket 🎟️ | [OCEANWATER-894](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-894) |
| Github :octocat:  | # |

## Summary of Changes
* Power fault activation is limited to one per simulation run. Additional activation attempts should be ignored and a warning displayed.

## Test
* In another terminal, navigate to the same directory and activate one of the three power faults:
* rosrun dynamic_reconfigure dynparam set /faults low_state_of_charge_power_failure True
* rosrun dynamic_reconfigure dynparam set /faults instantaneous_capacity_loss_power_failure True
* rosrun dynamic_reconfigure dynparam set /faults thermal_power_failure True
* The first fault activated should apply normally, however any additional faults should output a warning to the original terminal instead. This includes if you deactivate the first fault and then try to reactivate it again in the same simulation
* Faults can still be deactivated by using the same command to set them to False. This should prevent them from outputting future warnings

Note that the checkboxes in the dynamic reconfigure window will still show as checked off if you attempt to activate a fault, even if the fault is not applied.

I am unsure as to if the changes to FaultInjector.cpp are actually necessary; it seems like the primary change was done in power_system_node.cpp. I'm also unsure if the changes I made actually prevent the fault from affecting the simulation; to my understanding, they should, but I can't guarantee that.

### DIFF
--- a/ow_power_system/src/power_system_node.cpp
+++ b/ow_power_system/src/power_system_node.cpp
@@ -24,6 +24,10 @@ const string FAULT_NAME_THERMAL = "thermal_power_failure";
 //
 static constexpr int TEMPERATURE_INDEX = 1;
 
+// GLOBAL VARS (should implementation be changed to avoid this?)
+bool one_fault_injected = false;
+string one_fault_name = "";
+
 PowerSystemNode::PowerSystemNode()
 { }
 
@@ -242,9 +246,22 @@ void PowerSystemNode::injectFault (const string& fault_name,
 
   if (!fault_activated && fault_enabled)
   {
-    ROS_INFO_STREAM(fault_name << " activated!");
-    index = 0;
-    fault_activated = true;
+    // Only activate the fault if no other fault has been activated in the simulation thus far.
+    if (one_fault_injected)
+    {
+      ROS_WARN_STREAM
+        (fault_name << ": cannot activate; only one fault activation per simulation is supported.");
+    }
+    else 
+    {
+      ROS_INFO_STREAM(fault_name << " activated!");
+      index = 0;
+      fault_activated = true;
+
+      // TEST
+      one_fault_injected = true;
+      one_fault_name = fault_name;
+    }
   }
   else if (fault_activated && !fault_enabled)
   {
@@ -254,6 +271,13 @@ void PowerSystemNode::injectFault (const string& fault_name,
 
   if (fault_activated && fault_enabled)
   {
+    // Check and ensure the fault is the original one injected.
+    // Any additional faults are to be ignored due to one-fault-per-simulation.
+    if (fault_name != one_fault_name) 
+    {
+      return;
+    }
+
     // TODO: Unspecified how to handle end of fault profile, which is
     // unlikely.  For now, reuse the last entry.
     if (index + m_profile_increment >= sequence.size()) {
--- a/ow_power_system/src/power_system_node.cpp
+++ b/ow_power_system/src/power_system_node.cpp
@@ -249,7 +249,7 @@ void PowerSystemNode::injectFault (const string& fault_name,
     // Only activate the fault if no other fault has been activated in the simulation thus far.
     if (one_fault_injected)
     {
-      ROS_WARN_STREAM
+      ROS_WARN_STREAM_ONCE
         (fault_name << ": cannot activate; only one fault activation per simulation is supported.");
     }
     else 
--- a/ow_power_system/src/power_system_node.cpp
+++ b/ow_power_system/src/power_system_node.cpp
@@ -250,7 +250,9 @@ void PowerSystemNode::injectFault (const string& fault_name,
     if (one_fault_injected)
     {
       ROS_WARN_STREAM_ONCE
-        (fault_name << ": cannot activate; only one fault activation per simulation is supported.");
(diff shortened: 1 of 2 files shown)


kmdalal:
@Lumsterling I tested a bit and all works correctly so far.  I suspect that the changes in power_system_node are all were needed.  Could you please verify this?  I'll make some minor comments on the code later.

Lumsterling:
I reverted the edits to FaultInjector.cpp and it seems like nothing changed, as suspected.

Let me know if/when you make those minor comments, as it's something I want to improve on for the next time.

ow_power_system/src/power_system_node.cpp
```diff
@@ -24,6 +24,10 @@ const string FAULT_NAME_THERMAL = "thermal_power_failure";
 //
 static constexpr int TEMPERATURE_INDEX = 1;
 
+// GLOBAL VARS (should implementation be changed to avoid this?)
+bool one_fault_injected = false;
```
kmdalal, line 28: First, this needn't be a global variable, just a file variable.  Prefix a `static` to the declaration and remove the comment.

Second, per a (proposed) coding guideline on Confluence, a better name would be OneFaultInjected to help distinguish this variable from function variables and arguments.

ow_power_system/src/power_system_node.cpp
```diff
@@ -24,6 +24,10 @@ const string FAULT_NAME_THERMAL = "thermal_power_failure";
...
+string one_fault_name = "";
```
kmdalal, line 29: I'll suggest a clearer name for this one, `SelectedFaultName`.  Also add `static` to make this a file variable.

ow_power_system/src/power_system_node.cpp
```diff
@@ -242,9 +246,24 @@ void PowerSystemNode::injectFault (const string& fault_name,
...
-    ROS_INFO_STREAM(fault_name << " activated!");
-    index = 0;
-    fault_activated = true;
+    // Only activate the fault if no other fault has been activated in the simulation thus far.
+    if (one_fault_injected)
+    {
+      ROS_WARN_STREAM_ONCE
+        (fault_name << ": cannot activate; only one fault activation per simulation is supported."
+         << " Any additional faults for the remainder of this simulation will not be injected,"
+         << " even if they are shown as being activated.");
```
kmdalal, line 255: I suggest a shorter and more informative message, something like "Ignoring activation of fault FOO.  Only one power fault is permitted per simulation, and BAR has already been injected."

Note that FOO and BAR might be the same.

kmdalal:
> I reverted the edits to FaultInjector.cpp and it seems like nothing changed, as suspected.

I'm glad this was the case.  I've left code comments.  Will give the feature a more thorough test on Friday.

Lumsterling:
I spent a little time trying to brainstorm a way to get the warning to appear every time the user tried to activate a new fault, but I don't think it would be worth allocating even a tiny bit more memory for variables used only for the warning and nothing else. For context, if ROS_WARN_STREAM is used instead of ROS_WARN_STREAM_ONCE, the warning repeats endlessly given how the function is called every time a fault is injected. I was looking to see if I could have it output twice if the user attempted activating additional faults twice without making the feature too complicated, but I think having the one warning explain that future activations will also fail should be enough.

Comments have been updated based on your specifications. I think I'll take a look at the Confluence documents to get a better feel for how I should be writing them.

kmdalal:
Getting a warning every time the user activates a new fault would be good, and shouldn't be too hard.  The memory usage of variables is insignificant in this context, and in general.  

I assigned to you the Jira for consolidating the 3 faults into one.  You could delay this improvement until then, so you'd have less code to rip out. 

I'll give this feature a final test later this morning - have been tied up with some other things.

ow_power_system/src/power_system_node.cpp
```diff
@@ -24,6 +24,10 @@ const string FAULT_NAME_THERMAL = "thermal_power_failure";
 //
 static constexpr int TEMPERATURE_INDEX = 1;
 
+// Vars used to handle preventing more than one fault activation.
+static bool OneFaultInjected = false;
+static string SelectedFaultName = "";
```
kmdalal, line 29: After noticing that these variables are only used in one function, an improvement would be moving these definitions into that function (`injectFault`) as static _function_ variables.  This keeps the definitions as local as possible.   As function variables, their names should go back to lowercase/underscore.

kmdalal:
I am satisfied with testing.  After the small code change I have suggested (which I already tested), we can call this done!

Lumsterling:
I've pushed the small change.

To confirm just in case, should I be the one merging the pull request once the branch has been cleared (i.e. now), or should I leave that to you and/or others reviewing the pull request?

kmdalal (approved):
Nice work!  You may merge this branch and then delete it.  I usually use the web interface, which is easier than command line.

PR #249, opened 2022-07-05
Oceanwater 993 combine power fault injections

Lumsterling:
## Linked Issues:
| EPIC ⚡ | N/A |
| Jira Ticket 🎟️ | [OCEANWATER-993](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-993) |
| Github :octocat:  | # |

## Summary of Changes
* The 3 original power faults (thermal, low SOC, & instantaneous capacity loss) have been removed, given they all had nearly identical effects (inducing either a 24W or 25W power draw), and replaced with a single 'high power draw' power fault.
* This power fault can be tweaked in the RQT window or terminal. In the Dynamic Reconfiguration window under faults, the high power draw fault can be modified to induce any value between 0 and 30W. Checking the 'activate_high_power_draw' box beneath the slider injects the fault.
* The fault can be changed during runtime and will adapt to the new values on the slider.

## Test
* In another terminal, navigate to the same directory and modify/activate the new fault:
* rosrun dynamic_reconfigure dynparam set /faults high_power_draw_power_failure [0.0-30.0]
* rosrun dynamic_reconfigure dynparam set /faults activate_high_power_draw True
* If the fault is set to 25W, it should mimic the behavior of the previous power faults

## Potential Issue(s)
* The value and activation of the new fault seems to be saved between simulation runs for some reason despite Faults.cfg setting them to 0.0 and False at the start. It seems that other faults with a slider value share this behavior, but I don't recall this happening with checkbox-related faults previously (though it's happening now). Can anyone else verify this behavior on their computer?
* With this change, much of the original code regarding power faults (e.g. reading from .CSV files and updating power, voltage, and temperature based on the numbers in the file) is now unused. Should this code be kept for future use should someone down the line want to add a new custom fault? I can draft up a basic explanation for creating and adding a new fault to the simulator to accommodate this.

## Additional Notes
* The ticket also requests updating the user documentation on the ow_simulator wiki, as well as the Release 10 power tests. Unsure if I possess the knowledge to update the wiki, but I am working on drafting changes to the power tests. 
* Below is a draft of an explanation describing how to add a new power fault for future use based on the old .CSV setup. If it is easy enough to follow (and not missing any important steps), it could perhaps be added to the wiki or other user documentation if it is decided that the .CSV related code be kept.

## Adding a new power fault (Draft 1):
**_1._** Create a new .CSV file under **ow_simulator/ow_power_system/data**. It should be formatted with 4 columns, and named ``[FAULT_NAME]_power_failure.csv``. Here's a short example that adds a 25W power draw:
```
Timestamp, power, temperature, voltage
0, 25, 0, 0
1, 25, 0, 0
2, 25, 0, 0
3, 25, 0, 0
```

Timestamp should start at 0 and act as a column count.

**_2._** In **ow_simulator/ow_faults_injection/cfg/Faults.cfg**, a new power fault needs to be defined in the ``# POWER FAULTS`` section. The new line should follow this format:
```
power_faults.add("[FAULT_NAME]__power_failure",	bool_t,	0, "This is the mouse-hover description of the power fault", False)
```
'bool_t' sets the fault to a simple enable/disable checkbox in the RQT window, and False sets its initial value to unchecked.

**_3._** In **ow_simulator/ow_power_system_src/power_system_node.cpp**, multiple changes need to be made:

**_3a._** Towards the top of the program next to the other ``const string`` lines, a new string needs to be created with the same string as ``[FAULT_NAME]_power_failure``. The string's name should follow expected naming conventions. The ``const string`` will be called as the first argument in step 3c.

**_3b._** In ``PowerSystemNode::loadFaultPowerProfiles()``, two new lines need to be added:
```
path = ros::package::getPath("ow_power_system") + "/data/[FAULT_NAME]_power_failure.csv";
m_[FAULT_NAME]_power_failure_sequence = loadPowerProfile(path);
```

**_3c._** In ``PowerSystemNode::injectFaults(...)``, a new function call needs to be added:
```
injectFault(FAULT_NAME_NEW_FAULT,
      m_[FAULT_NAME]_power_failure_activated,
      m_[FAULT_NAME]_power_failure_sequence,
      m_[FAULT_NAME]_power_failure_sequence_index,
      power, voltage, temperature);
```

Said function call should be added for every individual power fault available, regardless of whether or not it is active at the time or not.

## Updating Release 10 power tests in Confluence:
It looks like all that would need to be changed in the Confluence test procedures would be to replace steps 9.2.3 - 9.2.5 with a single step utilizing the high power draw fault. I tested the ``noetic-devel`` branch and this current branch and found the results to be pretty much the same if I used a 25W power draw here. Curiously, I noticed that the ``remaining_useful_life`` metric seems to suddenly drop to 0 when a fault is injected (and eventually return to normal numbers if the fault is deactivated). Unsure if this is intended behavior or not, but it's consistent across both branches.


ow_faults_injection/cfg/Faults.cfg
```diff
@@ -48,8 +48,8 @@ ft_sensor_faults.add("signal_noise_failure", double_t, 0,
 
 # POWER FAULTS
 power_faults = gen.add_group("power_faults")
-power_faults.add("low_state_of_charge_power_failure", bool_t,   0, "Low state of charge power failure", False)
-power_faults.add("instantaneous_capacity_loss_power_failure", bool_t,   0, "Instantaneous capcity loss power failure", False)
-power_faults.add("thermal_power_failure", bool_t,   0, "Thermal power failure", False)
+power_faults.add("high_power_draw_power_failure",   double_t, 0,
+    "High power draw power failure", 0.0, 0.0, 30.0)
```
kmdalal, line 52: Just "high power draw" would be better, since the power draw itself isn't a failure -- it's potential effects are.

ow_power_system/src/power_system_node.cpp
```diff
@@ -238,6 +232,7 @@ void PowerSystemNode::injectFault (const string& fault_name,
   static string selected_fault_name = "";
 
   bool fault_enabled = false;
+  double hpd_value = 0.0;
```
kmdalal, line 235: Suggestion: rename to reflect the unit, e.g. `hpd_wattage`.

ow_faults_injection/cfg/Faults.cfg
Lumsterling: Sounds good. I was naming them based on the previous power fault naming scheme (where 'power_failure' was appended to every name) even if it seemed a little convoluted to name something 'high power draw power failure'.

Having the minimum value set to 0.0 was a holdover from a previous iteration before there was the 'enable/disable' checkbox for power draw activation. Will fix that too.

Lumsterling: As a heads up, the code will not build if I don't follow the set naming conventions. Attempting to name the slider "High power draw" instead of "high_power_draw" caused an error. I updated the names and mouse-hover labels within the naming convention limitations.

ow_power_system/src/power_system_node.cpp
```diff
@@ -15,9 +15,8 @@ using namespace std;
 using namespace std::chrono;
 using namespace std_msgs;
 
-const string FAULT_NAME_LOW_SOC = "low_state_of_charge_power_failure";
-const string FAULT_NAME_ICL     = "instantaneous_capacity_loss_power_failure";
-const string FAULT_NAME_THERMAL = "thermal_power_failure";
+const string FAULT_NAME_HPD           = "high_power_draw_power_failure";
```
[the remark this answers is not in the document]
Lumsterling: You're right, that was an oversight. Pushing the change shortly

ow_power_system/include/power_system_node.h
```diff
@@ -101,18 +101,10 @@ class PowerSystemNode
...
-  PrognoserVector m_instantaneous_capacity_loss_power_failure_sequence;
-  size_t m_instantaneous_capacity_loss_power_failure_sequence_index = 0;
-
-  bool m_thermal_power_failure_activated = false;
-  PrognoserVector m_thermal_power_failure_sequence;
-  size_t m_thermal_power_failure_sequence_index = 0;
+  
+  bool m_high_power_draw_power_failure_activated = false;
+  PrognoserVector m_high_power_draw_power_failure_sequence;
+  size_t m_high_power_draw_power_failure_sequence_index = 0;
```
kmdalal, line 107: These are not needed, since by definition this is a 'parametric' fault and does not use a predefined profile.

ow_power_system/src/power_system_node.cpp
```diff
@@ -278,44 +273,47 @@ void PowerSystemNode::injectFault (const string& fault_name,
...
+  //       The failure_sequence, failure_sequence_index,
+  //       voltage, and temperature
+  //       are no longer used as of JIRA 993. Should they
+  //       be left in for future use in case someone wants
+  //       to create a fault profile with additional modifiers?
+  injectFault(FAULT_NAME_HPD_ACTIVATE,
+        m_high_power_draw_power_failure_activated,
+        m_high_power_draw_power_failure_sequence,
+        m_high_power_draw_power_failure_sequence_index,
+        power, voltage, temperature);
```
kmdalal, line 316: Yes, the sequence references should be removed, but leave the rest of the framework as is for now.  If my proposal for a new custom power fault profile is agreed on, this could be added in easily under a new Jira.

ow_power_system/src/power_system_node.cpp
```diff
@@ -130,14 +129,9 @@ PrognoserVector PowerSystemNode::loadPowerProfile(const string& filename)
...
-  path = ros::package::getPath("ow_power_system") + "/data/low_state_of_charge_power_failure.csv";
-  m_low_state_of_charge_power_failure_sequence = loadPowerProfile(path);
 
-  path = ros::package::getPath("ow_power_system") + "/data/instantaneous_capacity_loss_power_failure.csv";
-  m_instantaneous_capacity_loss_power_failure_sequence = loadPowerProfile(path);
-
-  path = ros::package::getPath("ow_power_system") + "/data/thermal_power_failure.csv";
-  m_thermal_power_failure_sequence = loadPowerProfile(path);
+  path = ros::package::getPath("ow_power_system") + "/data/high_power_draw_power_failure.csv";
+  m_high_power_draw_power_failure_sequence = loadPowerProfile(path);
```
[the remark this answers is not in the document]
Lumsterling: The new CSV file was a holdover from before, when I was creating the new HPD fault based on the previous system. It was simply injecting a 25W fault like the old power faults. I'll remove it momentarily with the other changes.

ow_power_system/src/power_system_node.cpp
Lumsterling: I commented out the now-irrelevant parameters/variables without fully removing them. I imagine following additional feedback these could be removed completely, but until then I figured it would be best practice to leave them in as comments (though let me know if that isn't the case).

kmdalal (approved):
There's a small vestige of the power profile processing still here, the m_profile_increment member variable, but let's not worry about it.  I like the thorough job in describing the commented out parts of this functionality.

I don't see this fault activation and value persist across simulation runs as described in the PR.

The new fault feature seems to work very well.  Great work!

Lumsterling:
> I don't see this fault activation and value persist across simulation runs as described in the PR.

Very strange; I went back again and tested it and the value persisting issue is gone on my end too. I have no clue what caused it to begin with, nor why it disappeared again, but if you didn't notice anything then I suppose it must have been addressed at some point.

PR #251, opened 2022-07-07
OCEANWATER-974 terrain interaction improvement test

imynnnn:
## Linked Issues:
| Jira Ticket 🎟️ | [OCEANWATER-974](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-974) |
| Github :octocat:  | # |

## Summary of Changes
* In the test_06_deliver, following the call to _assert_scoop_regolith_containment, added additional assert that checks all regolith is now contained within the volume of the sample dock
* Approach to check point inside the volume: special case (axis aligned), this will not work well when lander land on sloped/rocky surface. Removed axis aligned approach. Added general case (oriented rectangle box) apply dot product method to check if the point inside the dock

## Test
* run the following command in the Terminal
* ```rostest ow_sim_tests terrain_interaction.test```

### DIFF
--- a/ow_sim_tests/test/terrain_interaction.py
+++ b/ow_sim_tests/test/terrain_interaction.py
@@ -183,22 +183,22 @@ class TerrainInteraction(unittest.TestCase):
   """
   Check a point inside a volume (dot product method)
   """
-  def check_point_in_volume_dot(self, dock, p):
+  def check_point_in_volume_dot(self, p):
     # define point as vector
     point = (p.x, p.y, p.z)
-    # define 4 vertices P1 P2 P4 P5
-    P1 = ( dock.x - SAMPLE_DOCK_LENGTH/2, dock.y - SAMPLE_DOCK_WIDTH/2,  dock.z - SAMPLE_DOCK_HEIGHT/2 )
-    P2 = ( dock.x - SAMPLE_DOCK_LENGTH/2, dock.y + SAMPLE_DOCK_WIDTH/2,  dock.z - SAMPLE_DOCK_HEIGHT/2 )
-    P4 = ( dock.x + SAMPLE_DOCK_LENGTH/2, dock.y - SAMPLE_DOCK_WIDTH/2,  dock.z - SAMPLE_DOCK_HEIGHT/2 )
-    P5 = ( dock.x - SAMPLE_DOCK_LENGTH/2, dock.y - SAMPLE_DOCK_WIDTH/2,  dock.z + SAMPLE_DOCK_HEIGHT/2 )
+    # define 4 vertices P1 P2 P3 P4
+    P1 = ( 0.6 ,   -0.425,  -6.53 )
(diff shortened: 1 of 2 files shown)


ow_sim_tests/test/terrain_interaction.py
```diff
@@ -176,6 +182,40 @@ def _assert_scoop_regolith_containment(self, assert_regolith_in_scoop=True):
               )
         )
 
+  """
+  Check a point inside a volume (dot product method)
+  """
+  def check_point_in_volume_dot(self, dock, p):
```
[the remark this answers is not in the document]
imynnnn: Sorry, wrong explanation. The error occur because the vector direction not consistent. Now it works

ow_sim_tests/test/terrain_interaction.py
```diff
@@ -22,6 +23,9 @@
 GROUND_POSITION = -0.155
 DISCARD_POSITION = Point(1.5, 0.8, 0.65) # default for discard action
 REGOLITH_CLEANUP_DELAY = 5.0 # seconds
+SAMPLE_DOCK_LENGTH = 0.27
+SAMPLE_DOCK_WIDTH  = 0.08
+SAMPLE_DOCK_HEIGHT = 0.04
```
AstroStucky, line 28: No longer used, and should be removed. 
```suggestion
```

ow_sim_tests/test/terrain_interaction.py
```diff
@@ -427,13 +464,13 @@ def test_06_deliver(self):
```
AstroStucky, line 464: The new assert works great! However, due to unreliable delivery of sample into the sample dock this test should be decorated with `@unittest.expectedFailure` at the very top of it. This is a temporary measure until we solve the delivery issue.

ow_sim_tests/test/terrain_interaction.py
```diff
@@ -176,6 +180,39 @@ def _assert_scoop_regolith_containment(self, assert_regolith_in_scoop=True):
               )
         )
 
+  """
+  Check a point inside a volume (dot product method)
+  """
+  def check_point_in_volume_dot(self, p):
```
AstroStucky, line 186: I'd like to see a function name here that has "dock" or "sample_dock" in it since the function is now specific to checking if a point is in the particular volume enclosed by the dock.

AstroStucky (commented):
The new assert works great! I was able to test both a success of the `test_06_deliver` and a failure. Both behave as expected. Just a few minor changes to the code suggested, and I'd like to see the `expectedFailure` decoration on `test_06_deliver` for the time being.

Awesome work! This new assert will enable automated testing of the deliver action with sample, which will go a long way towards making delivery more precise and keeping it that way. Just a few changes, and I'll approve this right away.

AstroStucky (approved):
Looks great! Well done.

PR #252, opened 2022-07-12
Oceanwater 947 arm move joints action

anjan1984:
## Linked Issues:
| Jira Ticket 🎟️ | [OCEANWATER-947](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-947) |
| Github :octocat:  | # |

## Summary of Changes
* Added ros action support to command all the joints to particular angle.

## Test
terminal 2:
to see all the option:
./arm_move_joints_client.py -h

The dafalt move is to set all the angles to 0.1 Run 
./arm_move_joints_client.py

other options:
python3 arm_move_joints_client.py false 0.1 0.1 0.1 0.1 0.1 0.1

To check the relative option you can try: 
python3 arm_move_joints_client.py true -0.5 0.0 0.0 0.0 0.0 0.0
python3 arm_move_joints_client.py true 0.5 0.0 0.0 0.0 0.0 0.0


kmdalal:
I just committed an executable version of the client script.

I did a fair amount of testing beyond the given examples.   Worked well, but the simplest case had a problem:
```
$ ./arm_move_joints_client.py
usage: arm_move_joints_client.py [-h] [relative] angles angles angles angles angles angles
arm_move_joints_client.py: error: the following arguments are required: angles
```
Also, the usage message could be improved.  Instead of multiple "angles" arguments (which should be singular), the joint name would be helpful.   

And I'm not sure how `relative` could be optional (as indicated by the brackets) since its position is not last.  This applies to some other client scripts -- I didn't catch it in previous reviews.

ow_lander/scripts/arm_move_joints_client.py
```diff
@@ -0,0 +1,60 @@
...
+    client = actionlib.SimpleActionClient(
+        'ArmMoveJoints', ow_lander.msg.ArmMoveJointsAction)
+
+    client.wait_for_server()
+
+    goal = ow_lander.msg.ArmMoveJointsGoal()
+
+    goal.relative = bool(args.relative)
+    #goal.joint = args.joint
+    goal.angles = args.angles
```
AstroStucky, line 38: This combination of an optional boolean argument and required array of floats is always required when put in this order. I have not been able to produce a situation where either of them are truly optional. I think it would make sense to instead expand `angles` out into the various named joint angles in order of their index. 
```suggestion
    parser.add_argument('relative', type=strtobool,
                        help='Move joint relative to the current position', nargs='?', default='False', const=0)

    parser.add_argument('j_shou_yaw', type=float,
        help='Shoulder yaw joint angle', nargs='?', default=0.1, const=0)
    parser.add_argument('j_shou_pitch', type=float,
        help='Shoulder pitch joint angle', nargs='?', default=0.1, const=0)
    parser.add_argument('j_prox_pitch', type=float,
        help='Proximal pitch joint angle', nargs='?', default=0.1, const=0)
    parser.add_argument('j_dist_pitch', type=float,
        help='Distal pitch joint angle', nargs='?', default=0.1, const=0)
    parser.add_argument('j_hand_yaw', type=float,
        help='Hand yaw joint angle', nargs='?', default=0.1, const=0)
    parser.add_argument('j_scoop_yaw', type=float,
        help='Scoop yaw joint angle', nargs='?', default=0.1, const=0)
    args = parser.parse_args()
    print_arguments(args)

    client = actionlib.SimpleActionClient(
        'ArmMoveJoints', ow_lander.msg.ArmMoveJointsAction)

    client.wait_for_server()

    goal = ow_lander.msg.ArmMoveJointsGoal()

    goal.relative = bool(args.relative)
    goal.angles = [
        args.j_shou_yaw,
        args.j_shou_pitch,
        args.j_prox_pitch,
        args.j_dist_pitch,
        args.j_hand_yaw,
        args.j_scoop_yaw
    ]
```

AstroStucky (changes requested):
I'd like to see my suggested improvement to the client arguments. The test works great, I tried a few different ranges of values. 

Is it possible to include the allowable ranges for a joint in the descriptions of each joint angle argument? I believe this would be a lander.xacro dependency, so I wonder if there is a way for Python to grab the values and insert them into the argument descriptions.

kmdalal:
Regarding whether the client script has individually named joints as parameters, or a single array of joint values, I'm not sure what users would prefer more.  I'd prefer the array, which is also what @MilanAdd is using in the Plexil plan interface.  Also, the ROS action takes an array, however that doesn't mean that user interfaces also need to.

I do strongly agree that the joint angle ranges should be documented, and ideally enforced in the client script.  It's too easy to create an invalid command.

ow_lander/scripts/arm_move_joints_client.py
anjan1984: Done!

anjan1984: as discussed we will have a separate jira to introduce the joint limits for all the joints and then we can revisit the problem of range of allowed angles for the joints.

AstroStucky (approved):
I like how the new help message reads with the joint angles expanded out. Test still works. Looks great!

kmdalal:
I tried the default invocation of the client script, and it successfully moved the arm into an extended position.  I wanted to move it back to the stowed position, and figured that this would do it:

./arm_move_joints_client.py true -0.1 -0.1 -0.1 -0.1 -0.1 -0.1

but it failed with an error setting joint target.   I also tried:

./arm_move_joints_client.py false 0 0 0 0 0 0

and this fails with the same error.  I assumed that any joint could be moved to 0 radians.

anjan1984:
> I tried the default invocation of the client script, and it successfully moved the arm into an extended position. I wanted to move it back to the stowed position, and figured that this would do it:
> ./arm_move_joints_client.py true -0.1 -0.1 -0.1 -0.1 -0.1 -0.1
> but it failed with an error setting joint target. I also tried:

In the default client script relative is false; so this wont get you back to stowed position. After default movement you can try 
./arm_move_joints_client.py true 0.1 0.1 0.1 0.1 0.1 0.1
and then 
./arm_move_joints_client.py true -0.1 -0.1 -0.1 -0.1 -0.1 -0.1
Then you can see it gets back to the first postion. 

> ./arm_move_joints_client.py false 0 0 0 0 0 0
> 
> and this fails with the same error. I assumed that any joint could be moved to 0 radians.

This also does not work as j_shou_pitch has a lower limit of 0.05

kmdalal (approved):
Thanks for the clarification on my last comment.  Works well with the tests I've done so far, manually and via PLEXIL.  It'll be good to have a more comprehensive test suite once the limits are known.

PR #253, opened 2022-07-13
Replaced all known instances of power_system_node

Lumsterling:
## Linked Issues:
| EPIC ⚡ | N/A |
| Jira Ticket 🎟️ | [OCEANWATER-1001](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1001) |
| Github :octocat:  | # |

## Summary of Changes
* Using 'git grep' and 'git mv' I renamed the relevant power system files and then went through the repository and renamed any direct calls to the .cpp ro, to fit naming conventions. Preliminary tests show nothing changed functionally, as should be expected.

* Calling 'rostopic echo' to view relevant power system statistics is unchanged.

## Test
* In another terminal, navigate to the same directory and activate high power draw (or use the Dynamic Reconfigure tab in rqt):
```
rosrun dynamic_reconfigure dynparam set /faults high_power_draw_power_failure [0.0-30.0]
rosrun dynamic_reconfigure dynparam set /faults activate_high_power_draw True
```
* In 3 other terminals, monitor the power system statistics using the same name and ensure they update properly:
```
```


kmdalal (changes requested):
My tests pass, and the class filename change looks good, but this work also renames the _topics_, which was not part of the Jira.  Topic names are unrelated to C++ names.  

While prefixing the topics `PowerSystemNode` makes everything consistent, this change will require an edit to the user docs.  Plus, the topics will need to get renamed again when we do the telemetry unification with JPL.  And they might get renamed as a result of the battery pack model. 

It would be best to revert the topic prefix to `power_system_node`.  Does this make sense?

Lumsterling:
I think the wording in the JIRA ticket of modifying "all dependencies throughout the system" is why I thought all instances needed to be updated. I've gone back through and reverted most of the name changes. The only ones that remain now are the file names and the calls to those files (which is only ~3 changes overall).

I'm not totally certain on the distinction of topics and names, but hopefully this change addresses the issue. If not, I'm guessing I didn't change enough on my end (although the simulation builds and seemingly runs correctly).

kmdalal (approved):
Thanks for the revert on the topic names.  These 3 files and simple changes are all I was expecting, and this was intended to be a trivial change.  It was co-incidental that the topic names were prefixed with the old filename.  I meant dependencies on the filenames in the Jira -- should have been more specific.   Looks good!

PR #254, opened 2022-07-20
Implemented custom fault support

Lumsterling:
## Linked Issues:
| EPIC ⚡ | N/A |
| Jira Ticket 🎟️ | [OCEANWATER-1002](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1002) |
| Github :octocat:  | # |

## Summary of Changes
* User now has a new option under RQT -> Dynamic Reconfigure -> faults. A custom fault can be added to a specific directory, and when activated, will inject the fault using the same logic from before the changes in [OW-993](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-993).
* Basic error catching prevents the program from terminating if the user inputs an invalid file name, either not found in the _ow_power_system/profiles_ directory, and/or improper formatting.
* ~~Similar to [OW-894](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-894)'s original changes, the user is restricted to using either the provided high power draw or their own custom fault, never both at once. Activating one fault prevents the other from being activated for the remainder of that simulation and appropriate warnings are provided should the user attempt this.~~
  * The restriction is no longer implemented, and the user can activate any combination of the provided high power draw as well as their own custom faults. However, only one custom fault at a time can be active.
* Multiple different custom faults can be deactivated and reactivated in the same simulation, as long as the provided file name is valid (though only one custom fault can be active at a time).
* If the custom fault file runs its course (i.e. the end of the file is reached), the fault injection ceases. If the user deactivates and reactivates the custom fault without changing the file name, the fault will be re-loaded and restart at index 0. Otherwise, whenever the custom fault is deactivated and then reactivated with a new file name in the box, the previous fault is unloaded and the new one loaded.
* A blank example .CSV file has been added under _ow_power_system/profiles/example_fault.csv_. This file is provided to show potential users how to format their own custom faults.
  * The file does not necessarily have to use the .csv extension, but its formatting must be identical to the provided example.
* NOTE: Additional testing was done to see if a more user-friendly method of fault selection was available (e.g. using a file selection dialog, a drop-down menu, etc.) but such options do not seem feasible under the Dynamic Reconfigure tab.

## Test
* catkin build
* In 3 other terminals, run one of the 3 following commands in each terminal to monitor statistics:
```
```
* Either use the provided _example_test.csv_ file for the next step, or create your own custom fault .csv file using the same format as seen in _example_test.csv_ and place it in the _ow_power_system/profiles_ directory
  * One example is to make a new CSV where every index has a 'wattage' value of 25, simulating a 25W high power draw. This should have identical results as a 25W high power draw using the non-custom feature
* Navigate to the Dynamic Reconfigure tab -> faults and enter the full name of the fault file to be tested in the _custom_fault_name_ section
* Check the box for _activate_custom_fault_ to activate the custom fault. Monitor battery statistics to ensure the custom fault's behavior is expected.
* ~~Ensure high power draw cannot be applied in the current simulation (since a valid custom power fault has already been applied) by checking the box for _activate_custom_fault_ and observing the console running the workspace for warnings.~~
  * High power draw can now be used simultaneously with custom power faults.
* Confirm the new fault feature can be utilized from the terminal:
  * Modify the information in the file path text box using a console command in a new terminal:
    * rosrun dynamic_reconfigure dynparam set /faults custom_fault_name [file name]
  * rosrun dynamic_reconfigure dynparam set /faults activate_custom_fault True

## Potential Issue(s)
* ~~Error-catching logic does not prevent user from attempting to activate a .CSV file containing formatting unrecognized by the old code.~~
  * Error detection during file loading has been implemented.
* A better system for file designation (rather than manual entry of the desired file path) would be to use an open-file dialog, however I was unable to find any information on whether or not this is possible within the RQT window. It seems possible, given there are buttons to select _.perspective_ files to upload a perspective to the window, but I couldn't find anything regarding adding one of those buttons within the faults tab of Dynamic Reconfigure.
* Updating the user documentation and release tests will be done after the PR is ready to merge.


Lumsterling:
I've spent some more time looking at the dynamic reconfigure tutorials, and at this point I'm not too sure designating a file directory in the same tab is possible, at least in C++. I have some ideas for maybe editing the _.cfg_ python file and filling an _enum_ with options based on the number of CSV files found in a designated directory though, but nothing concrete. I'll update this once I've come to a better conclusion on that end.

Lumsterling:
After more experimentation, I'm unsure that it's possible to get the file names of each file in a given directory to show up in Dynamic Reconfigure using the _enum_ option. The setup for the RQT window must be done on startup and in a .cfg file, which doesn't support getting the current directory and parsing through files (to my understanding). It is possible to find a directory and parse through its files later in PowerSystemNode.cpp (that's what the old system did), but this part of the code isn't able to modify the RQT window (again, to my understanding). I'm thinking it is possible to designate, say, 5 custom fault 'slots' on startup in the RQT window and have added custom files use those 'slots', but since those slot names can't be adjusted after launch it could lead to confusion if the user uploads 2+ custom faults at once (all they would see in the window are slots labeled 1-5, so unless they name their fault files alphabetically they might expect one file to be in slot 1 when it is actually in slot 2). However, ROS_INFO output could be used to notify the user which faults are activated by name, so they would know if they activated the wrong custom fault, but only after activating it. If the user is not allowed to activate more than one type of fault per simulation (custom or otherwise) though, they would still have to restart the entire simulation.

The bottom line is that this other implementation contains a lot of problems too. Some of these could be mitigated by restricting the user to one uploaded custom fault, but I think we talked some time ago about that being an arbitrary restriction that we shouldn't implement.

kmdalal:
My Ubuntu has been crashing since yesterday, but I finally got to test this a few times.  I tried my own custom fault profile, and the feature mostly worked.  Here is my feedback so far.

With respect to profile selection:

- I appreciate the time you've put into searching for file-browsing capability in RQT, but I'm fine with giving up on this since it doesn't look doable.  However, file path type-in is just too cumbersome, and it may be hard to predict how relative paths will be handled.  I suggest a middle ground: type-in of the just the filename, which must end with `.csv`.  The file must be placed in a designated directory.
- Name the designated directory `profiles` instead of `TEMP_DATA`
- I'll suggest that test_fault.csv just contain a pageful of lines, all 0s, instead of a 25W draw, but we can get other opinions.

For profile processing:
- The specified profile should be read when the fault is injected.  This way the user can change files during a simulation.
- This fault should be re-injectable.  This didn't seem to work for me.  It worked correctly the first time, though.

Lumsterling:
I took the suggestions from the previous comment and this new version should work properly. I'd forgotten to set a flag on fault re-activation so that's why it wasn't working before.

I'm unsure how much detail the user needs when it comes to console output, but I figured an "attempting activation" message was pertinent because there are a multitude of reasons why the activation could fail. Warning messages to notify the user that custom faults can cause unexpected outputs (as well as notifying them that high power draw cannot be used simultaneously with custom faults) only print on the first successful fault activation and not on subsequent ones. Let me know if these additional messages are useful or not.

ow_faults_injection/cfg/Faults.cfg
```diff
@@ -103,5 +103,7 @@ power_faults.add("high_power_draw",   double_t, 0,
     "High power draw increment", 1.0, 1.0, 30.0)
 power_faults.add("activate_high_power_draw",    bool_t, 0,  "Activate high power draw at the selected wattage", False)
 
+power_faults.add("custom_fault_name",   str_t, 0, "Enter the name of a desired CSV file within the ow_power_system/profiles directory", "Enter file name here (must end with '.csv')")
```
kmdalal, line 106: Wording suggestions:
-  "custom_fault_profile"
- "Enter the name of the desired CSV file in ow_power_system/profiles"

Have the initial string be "example_fault.csv".  This conveys the required format, and they can use this file if they edited it.

kmdalal:
I retested a few times, and this is definitely improving!  My main comments for now:

- It didn't seem like reactivation of the custom profile works.  I used a 30-line CSV that I expected to restart from line 1, but it behaved like the last line was still being reused.
- I now have second thoughts about prohibiting high power draw in combination with this!  I ran a 30-second custom fault profile.  After that I wanted to inject some more power and couldn't.  We can discuss this in the group.

I have some more testing to do.

kmdalal:
Minor: I'd mentioned this before, but I think a much shorter (30-60 line) example_fault.csv would be more usable.  I suspect most users would either use their own file, or try simple experiments with this one, which currently means deleting a lot of rows.

Lumsterling:
> * It didn't seem like reactivation of the custom profile works.  I used a 30-line CSV that I expected to restart from line 1, but it behaved like the last line was still being reused.

I misinterpreted this earlier: I thought you were talking about how the fault activation was unable to reactivate a fault if you unchecked the box then rechecked it without changing anything. I fixed that in the last commit, thinking that was what you were talking about. I can easily add this too, it's just a matter of changing a single line of code in the old fault activation logic.

> * I now have second thoughts about prohibiting high power draw in combination with this!  I ran a 30-second custom fault profile.  After that I wanted to inject some more power and couldn't.  We can discuss this in the group.

I can remove the limitation, I was just concerned because in my tests I was seeing some odd behavior at the time. I had the custom fault inject a 25W power draw and I was also testing a 30W high power draw. The simulation said it would cap the power draw at 30W, but there was something else going on with it (that I can't recall now) that didn't look right in the battery statistics.

> * Minor: I'd mentioned this before, but I think a much shorter (30-60 line) example_fault.csv would be more usable. I suspect most users would either use their own file, or try simple experiments with this one, which currently means deleting a lot of rows.

Yup, I forgot about this and it goes hand-in-hand with restarting the fault back at index 0 when the end of the CSV file is reached. I can address all these things after the meeting today.

kmdalal:
Small request: merge noetic-devel into your branch and re-push.

Lumsterling:
> Small request: merge noetic-devel into your branch and re-push.

Done (unsure if it notifies you when I push new changes or not, so I'm sending this comment to be sure)

ow_power_system/profiles/example_fault.csv
```diff
@@ -0,0 +1,51 @@
+Timestamp, power, temperature, voltage
```
kmdalal, line 1: First column would be better labeled something like "Index" .  How the profile is read and explained to the user can be left for a followon Jira.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -263,30 +268,155 @@ void PowerSystemNode::injectFault (const string& fault_name,
...
+        }
+        catch (int err_val)
+        {
+          switch(err_val)
+          {
+            case -1:
+              ROS_WARN_STREAM("Invalid file name entered; Deactivate fault and try again.");
+              break;
+            case -2:
+              ROS_WARN_STREAM("Could not find CSV in the 'profiles' directory with the provided name. Deactivate fault and try again.");
```
kmdalal, line 371: Would be helpful to include the filename in these messages.  If this is hard or impractical, don't worry about it.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -263,30 +268,155 @@ void PowerSystemNode::injectFault (const string& fault_name,
...
+        // Skip loading and continue from previous index.
+        ROS_INFO_STREAM(saved_file.substr(0, saved_file.size() - 4) << " re-activated!");
+        custom_fault_ready = true;
+      }
+    }
+    fault_activated = true;
+  }
+  else if (fault_activated && !fault_enabled)
+  {
+    ROS_INFO_STREAM("Custom power fault de-activated!");
```
kmdalal, line 391: In this and all messages that refer to the "custom" fault, it would be more helpful to identify it by the profile name.  If this is hard, don't worry about it.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -263,30 +268,155 @@ void PowerSystemNode::injectFault (const string& fault_name,
...
+      if (!end_fault_warning_displayed)
+      {
+        ROS_WARN_STREAM
+          ("custom_fault: reached end of fault profile. Fault disabled, but will restart if re-enabled.");
+        end_fault_warning_displayed = true;
       }
-      else index += m_profile_increment;
 
+      // Probably unneeded, but makes index explicit.
+      //index = sequence.size() - 1;
```
kmdalal, line 411: You can just delete these lines, and any others similar.

ow_power_system/profiles/example_fault.csv
kmdalal: Could as well label the second column "wattage", followed by "temperature (Celsius)", to make all the units clear.

ow_power_system/include/PowerSystemNode.h
```diff
@@ -102,12 +106,13 @@ class PowerSystemNode
 
   std::uniform_real_distribution<double> m_temperature_dist;
   
-  bool m_high_power_draw_power_failure_activated = false;
+  bool m_high_power_draw_activated = false;
   // DEPRECATED (7/12/22)
   // Part of the old .CSV system for power faults. It will be revisited
   // in the coming days/weeks to either fully remove or reinstantiate.
```
kmdalal, line 112: This comment can be deleted now.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -18,6 +18,10 @@ using namespace std_msgs;
 const string FAULT_NAME_HPD           = "high_power_draw";
 const string FAULT_NAME_HPD_ACTIVATE  = "activate_high_power_draw";
 
+// GLOBAL VARS?
+bool CUSTOM_FAULT_ACTIVATED           = false;
+bool HPD_FAULT_ACTIVATED              = false;
```
kmdalal, line 23: These can be file (static) variables.  Suggest CamelCase which is easier on the eyes.  All caps are for constants.
```suggestion
static bool CustomFaultActivated = false;
static bool HPDFaultActivated      = false;
```

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -244,15 +232,32 @@ void PowerSystemNode::injectFault (const string& fault_name,
...
+      }
+    }
+    else
+    {
+      ROS_INFO_STREAM(fault_name << " activated!");
+      //ROS_WARN_STREAM_ONCE("Note that high power draw cannot be combined with custom faults.");
+      // TEMP: Uncomment this and the other 'TEMP' line code to re-enable restricting power faults:
+      // When uncommented, the user will only be able to use either high power draw or the custom
+      // fault feature, never both in a single simulation. Behavior of both at once is unpredictable.
+      //HPD_FAULT_ACTIVATED = true;
```
kmdalal, line 252: These comments can be deleted.  I'm not envisioning a need to bring restrictions back, especially since the GUI cannot enforce them.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -244,15 +232,32 @@ void PowerSystemNode::injectFault (const string& fault_name,
...
+    }
     fault_activated = true;
   }
   else if (fault_activated && !fault_enabled)
   {
     ROS_INFO_STREAM(fault_name << " de-activated!");
```
kmdalal, line 258: Minor: "deactivated" is a more common spelling, I think.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -263,30 +268,155 @@ void PowerSystemNode::injectFault (const string& fault_name,
...
+  if (!fault_activated && fault_enabled)
+  {
+    // Only continue if no other faults have been injected.
+    ROS_INFO_STREAM("Attempting custom fault activation...");
+    if (HPD_FAULT_ACTIVATED)
+    {
+      if (!custom_warning_displayed)
+      {
+        ROS_WARN_STREAM("Cannot activate custom fault: high power draw already used this simulation.");
+        custom_warning_displayed = true;
```
kmdalal, line 305: This warning is no longer valid, and yet this code seems unreachable.  At least I don't get this warning when I perform what it says it cannot do.

Lumsterling: Yup, it was deprecated as of commenting out the flags that were set upon activating either a custom fault or high power draw. I've removed them in the incoming commit (along with all of the other changes you've requested).

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -263,30 +268,155 @@ void PowerSystemNode::injectFault (const string& fault_name,
...
+      }
+    }
+    else
+    {
+      // Get user-entered file directory
+      if (! ros::param::getCached("/faults/custom_fault_profile", designated_file))
+      {
+        ROS_ERROR("Failed to fetch custom fault name.");
```
kmdalal, line 313: I presume this should never happen.  In any case, the check is fine but maybe call it "custom fault profile name".

Lumsterling: I don't think it should happen either; this check was simply part of the old code we're repurposing. I can remove it if you'd like, otherwise I'll modify the error message.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -263,30 +268,155 @@ void PowerSystemNode::injectFault (const string& fault_name,
...
+      
+      if (current_fault_directory != saved_fault_directory || end_fault_warning_displayed)
+      {
+        // It's a different fault profile from whatever the previous stored directory was, OR the
+        // previous fault ran its course, so reset index.      
+        // Load a new fault profile from the provided directory.
+        try
+        {
+          ifstream file(current_fault_directory);
+          if (current_fault_directory.substr(current_fault_directory.size() - 4) != ".csv")
```
kmdalal, line 328: Minor: There may be an API for getting the extension from a pathname, but this is okay and clear.

Lumsterling: It's definitely a bit of a hack, but since we are only looking for a single type of file extension, I figured it wasn't a top priority.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -263,30 +268,155 @@ void PowerSystemNode::injectFault (const string& fault_name,
...
+          {
+            ROS_INFO_STREAM("Loading " << designated_file << "...");
+          }
+          loadCustomFaultPowerProfile(current_fault_directory);
+          custom_fault_ready = true;
+          end_fault_warning_displayed = false;
+          // TEMP: Uncomment this and the other 'TEMP' line code to re-enable restricting power faults:
+          // When uncommented, the user will only be able to use either high power draw or the custom
+          // fault feature, never both in a single simulation. Behavior of both at once is unpredictable.
+          //CUSTOM_FAULT_ACTIVATED = true;
```
kmdalal, line 354: More that can be deleted.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -263,30 +268,155 @@ void PowerSystemNode::injectFault (const string& fault_name,
...
+          saved_fault_directory = current_fault_directory;
+          saved_file = designated_file;
+          index = 0;
+
+          ROS_WARN_STREAM_ONCE("Custom power faults may exhibit unexpected results. Caution is advised.");
+          //ROS_WARN_STREAM_ONCE("Note that custom power faults cannot be combined with high power draw.");
```
kmdalal, line 360: And another.

kmdalal:
I've tested the following, in one simulation session, and all worked well and as expected:
 - multiple instances of high power draw
 - multiple instances of custom fault
 - multiple instances of both of the above combined, in varying orders of starting and stopping the fault
 - entering a non-existent profile name

Aside from some small suggested code changes, this looks good to go!  However, given it's a substantial change to power fault injection, it should get an additional reviewer -- I'll add one.

ow_power_system/src/PowerSystemNode.cpp
Lumsterling: I've actually removed these two variables, since they were tied to the whole 'restrict-faults-to-one-or-the-other' feature that has since been removed.

ow_power_system/src/PowerSystemNode.cpp
kmdalal: Leave the check in (adds robustness), just modify the message, thanks!

ow_power_system/src/PowerSystemNode.cpp
kmdalal: That's fine, you can leave it like this.

kmdalal:
I'm done with my code review and testing.  Have requested an additional review from @AstroStucky .   Please ping me (re-request review) when he's finished, and I'll give it another quick test .

AstroStucky:
When I provide a CSV with incorrect formatting (specifically an inconsistent number of commas in rows) it results in the following exception.
```
[ INFO] [1659391299.597410495, 1916526486.707499981]: Loading example_fault.csv...
terminate called after throwing an instance of 'std::invalid_argument'
  what():  stod
[power_system_node-20] process has died [pid 34086, exit code -6, cmd /usr/local/home/tstucky/ow/workspace/devel/lib/ow_power_system/power_system_node __name:=power_system_node __log:=/usr/local/home/tstucky/.ros/log/945c08b6-11e1-11ed-aa41-57d23f8a5070/power_system_node-20.log].      
log file: /usr/local/home/tstucky/.ros/log/945c08b6-11e1-11ed-aa41-57d23f8a5070/power_system_node-20*.log
```
I would like to see this exception caught and result in a more descriptive error message for the user.

kmdalal:
> When I provide a CSV with incorrect formatting (specifically an inconsistent number of commas in rows) it results in the following exception.

I believe this work doesn't touch the existing CSV-reading code, which had the same lack of robustness, partly because it used only our provided profiles which can be assumed valid.  It would be good to add a bit of robustness here, but detailed checking of the CSV may be best skipped or future work.

Lumsterling:
> When I provide a CSV with incorrect formatting (specifically an inconsistent number of commas in rows) it results in the following exception.

Like Mike said, this error stems from the original code which didn't check if the file was formatted properly because the file was provided by us. However, it's true that this is a type of error that could easily be raised by a user trying to load their own file, so I can take a look at the old code and see if I can implement some form of error catching there. It won't be ready today though.

Lumsterling:
I've pushed changes that should prevent the process from terminating if an invalid CSV file is loaded. To my understanding, it should be able to detect if there is a row with too many columns, too few columns, and/or if the data in a column is invalid (e.g. letters). Unsure if there are other potential avenues of error in a CSV file, but in general if an exception is thrown while loading the file the program should now handle it.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -263,30 +277,151 @@ void PowerSystemNode::injectFault (const string& fault_name,
...
+      {
+        ifstream file(current_fault_directory);
+        if (current_fault_directory.substr(current_fault_directory.size() - 4) != ".csv")
+        {
+          throw ERR_INVALID_FILE_NAME;
+        }
+        else if (file.fail())
+        {
+          throw ERR_NO_CSV_FOUND;
+        }
```
AstroStucky, line 335: This ifstream is only created here for the sole purpose of interrogating whether the file exists, but creating an ifstream and checking its existence is already performed by `PowerSystemNode::loadPowerProfile`, which is called later in the try block. 

I suggesting removing these lines and the ifstream above.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -263,30 +277,151 @@ void PowerSystemNode::injectFault (const string& fault_name,
...
+      // It's a different fault profile from whatever the previous stored directory was, OR the
+      // previous fault ran its course, so reset index.      
+      // Load a new fault profile from the provided directory.
+      try
+      {
+        ifstream file(current_fault_directory);
+        if (current_fault_directory.substr(current_fault_directory.size() - 4) != ".csv")
+        {
+          throw ERR_INVALID_FILE_NAME;
+        }
```
AstroStucky, line 331: I differ from Michael here, and I don't think it's at all necessary to enforce file extension. A ".profile" file could work perfectly fine too, so long as it's contents are properly formatted.

In either case, if we are to check the file extension, I think that is best performed inside the `PowerSystemNode::loadPowerProfile` function, where file existence and formatting checks also occur.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -263,30 +277,151 @@ void PowerSystemNode::injectFault (const string& fault_name,
...
+        custom_fault_ready = false;
+      }
+    }
+    else
+    {
+      // It's the same fault profile as previously stored, meaning the same fault was reactivated.
+      // Skip loading and continue from previous index.
+      ROS_INFO_STREAM(saved_file.substr(0, saved_file.size() - 4) << " re-activated!");
+      custom_fault_ready = true;
```
AstroStucky, line 386: From discussion I read in the PR, shouldn't this case trigger a reload of the fault profile file?

AstroStucky (changes requested):
Some code changes requested.

In my tests, the custom fault profiles appear to work, but I don't know exactly what to look for in the power node output to verify they're doing the right thing. If I inject a high wattage, then I can see that the remaining useful life goes down and sometimes even collapses to zero (could this be an unrelated bug?), and the temperature rises steadily. If I inject a high voltage, I don't see any difference. If I inject a high temperature, I don't see any difference, but I would have expected `/power_system_node/battery_temperature` to increase by the same amount as in my profile.  

I haven't been able to inject a fault profile, modify that profile, and re-inject it. This seems to require a reboot of the simulation for the modified profile to be reloaded, but I'm also not sure because as I said I find it difficult to tell if my fault profile does much of anything.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -78,65 +79,84 @@ bool PowerSystemNode::loadSystemConfig()
...
+        continue;
+
+      stringstream line_stream(line);
+      string cell;
+
+      // Confirm the line contains only 4 columns.
+      stringstream line_stream_check(line);
+      int cols = 0;
+      while (getline(line_stream_check, cell, ','))
+        cols++;
```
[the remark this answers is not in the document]
AstroStucky: The plus one is because this counts the number of commas, but being a delimiter, there should be one extra column following the final comma.

Lumsterling:
> Some code changes requested.

I'll be going through the feedback you posted shortly.

> In my tests, the custom fault profiles appear to work, but I don't know exactly what to look for in the power node output to verify they're doing the right thing.

To be completely honest, I don't know for certain if the other types of fault injection (temp & voltage) work, especially given your testimony that the temperature entry doesn't seem to affect the battery temperature at all. The way they are implemented is from whoever wrote the previous code, as I simply took that and built the file injector around it. I'm in the same boat as you when it comes to uncertainty on testing results outside of power draw; it's definitely something that needs to be looked at, but my initial thought is that addressing it is out of the scope of this current PR (which focuses on the implementation of custom fault profiles).

> If I inject a high wattage, then I can see that the remaining useful life goes down and sometimes even collapses to zero (could this be an unrelated bug?)

It is a bug, and last I'd heard Chetan was working on it. It's outside of the scope of this PR to my understanding.

> I haven't been able to inject a fault profile, modify that profile, and re-inject it. This seems to require a reboot of the simulation for the modified profile to be reloaded, but I'm also not sure because as I said I find it difficult to tell if my fault profile does much of anything.

Essentially, the current logic is that if you activate a file, deactivate it before the injector reaches the end of the file, and then reactivate it, it continues from the index it left off on. However, if the injector reaches the end of the file and then you deactivate & reactivate the same file, it will reload it (which will load any new values you changed in the file itself). In practice, you're probably right in thinking this isn't the kind of behavior to be expected, so I can change it to fully reload the file every time it is reactivated instead.

kmdalal:
> In my tests, the custom fault profiles appear to work, but I don't know exactly what to look for in the power node output to verify they're doing the right thing.

As @Lumsterling mentioned, fault profile behavior is outside the scope of this PR.  I too failed to observe interesting (or any) results when injecting voltage and temperature, and will create a Jira to investigate this.

kmdalal:
> If I inject a high wattage, then I can see that the remaining useful life goes down and sometimes even collapses to zero (could this be an unrelated bug?)

This is a long-standing bug (and Jira that @chetankul has been working on) that  I've brought up in many meetings.  Again, outside the scope of this PR.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -263,30 +250,144 @@ void PowerSystemNode::injectFault (const string& fault_name,
...
+    return;
+  }
+
+  if (!fault_activated && fault_enabled)
+  {
+    // Multiple potential points of failure, so alert user the process has started.
+    ROS_INFO_STREAM("Attempting custom fault activation...");
+
+    // Get user-entered file directory.
+    if (! ros::param::getCached("/faults/custom_fault_profile", designated_file))
```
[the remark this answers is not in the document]
kmdalal: Good catch, @AstroStucky.  @Lumsterling I wonder if this might explain some of the "flukey" behavior we've seen before.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -78,65 +79,84 @@ bool PowerSystemNode::loadSystemConfig()
...
-    Datum<double> voltage(stod(cell));
-    voltage.setTime(timestamp);
-
-    data.insert({ MessageId::Watts, power });
-    data.insert({ MessageId::Centigrade, temperature });
-    data.insert({ MessageId::Volts, voltage });
-
-    result.push_back(data);
+    ROS_ERROR_STREAM("Failed to read " << csv << ": Improper formatting detected." << endl << 
+                     "Confirm " << csv << " follows the exact format of example_fault.csv before retrying.");
```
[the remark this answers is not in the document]
kmdalal: Nice idea, though I recommend not putting too much more time into error checking, to have time for the battery pack.  Adding robustness to the profile reading could be a follow-on Jira.

ow_power_system/src/PowerSystemNode.cpp
kmdalal: Good observation, @AstroStucky - I guess the file extension doesn't really matter!  Even a file without extension seems fine.

ow_power_system/src/PowerSystemNode.cpp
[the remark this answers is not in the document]
Lumsterling: Interesting. This is another holdover from the original code; I just left it in without modifying it since it worked. I assumed it stored the parameter if found and returned true if so, and false if it couldn't find it. If it doesn't affect the behavior of the program at all I can remove it.

kmdalal:
> Essentially, the current logic is that if you activate a file, deactivate it before the injector reaches the end of the file, and then reactivate it, it continues from the index it left off on. However, if the injector reaches the end of the file and then you deactivate & reactivate the same file, it will reload it (which will load any new values you changed in the file itself). In practice, you're probably right in thinking this isn't the kind of behavior to be expected, so I can change it to fully reload the file every time it is reactivated instead.

Yes, I think the profile should start from the top every time it is (re) activated.  There may be a use case for wanting continuation, but this seems most intuitive to me.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -263,30 +250,144 @@ void PowerSystemNode::injectFault (const string& fault_name,
...
+
+    // Get user-entered file directory.
+    if (! ros::param::getCached("/faults/custom_fault_profile", designated_file))
+    {
+      ROS_ERROR("Failed to fetch custom fault profile name.");
+      return;
+    }
+
+    // Append the current fault directory to the stored file path.
+    current_fault_directory = ros::package::getPath("ow_power_system") + "/profiles/" + designated_file;
```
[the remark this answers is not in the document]
Lumsterling: This was a holdover from before when I was attempting to get directories full of fault profiles instead of just looking into a designated profiles directory. I'll change the variable name.

AstroStucky:
@Lumsterling Please don't forget to re-request a review from @kmdalal and myself when you're finished integrating requested changes.

Lumsterling:
> @Lumsterling Please don't forget to re-request a review from @kmdalal and myself when you're finished integrating requested changes.

My mistake; I quickly pushed those changes at the end of the day and forgot to come back to it in the morning.

ow_power_system/src/PowerSystemNode.cpp
Lumsterling: (resolved in another comment; reactivation now fully reloads the file)

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -78,65 +80,97 @@ bool PowerSystemNode::loadSystemConfig()
...
+PrognoserVector PowerSystemNode::loadPowerProfile(const string& filename, string csv)
 {
   ifstream file(filename);
+
+  // Checking for .csv extension isn't strictly necessary as long as the
+  // formatting is correct, but for now, CSV files are expected.
+  if (filename.substr(filename.size() - 4) != ".csv")
+  {
+    throw ERR_INVALID_FILE_NAME;
+  }
```
kmdalal, line 92: This can be removed -- I see no need to enforce a filename.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -17,6 +18,12 @@ using namespace std_msgs;
 
 const string FAULT_NAME_HPD           = "high_power_draw";
 const string FAULT_NAME_HPD_ACTIVATE  = "activate_high_power_draw";
+const int CSV_EXPECTED_COLS           = 4;
+
+// Error flags.
+const int ERR_CSV_FORMAT              = -1;
+const int ERR_INVALID_FILE_NAME       = -1;
```
kmdalal, line 25: Can be removed.  See related comment below.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -78,65 +80,97 @@ bool PowerSystemNode::loadSystemConfig()
...
+      if (line.empty())
+        continue;
+
+      stringstream line_stream(line);
+      string cell;
+
+      // Confirm the line contains the expected number of columns.
+      auto cols = std::count(line.begin(), line.end(), ',') + 1;
+
+      ROS_INFO_STREAM("COLS: " << cols);
```
kmdalal, line 126: Oops, left this in.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -78,65 +80,97 @@ bool PowerSystemNode::loadSystemConfig()
...
-    Datum<double> voltage(stod(cell));
-    voltage.setTime(timestamp);
-
-    data.insert({ MessageId::Watts, power });
-    data.insert({ MessageId::Centigrade, temperature });
-    data.insert({ MessageId::Volts, voltage });
-
-    result.push_back(data);
+    ROS_ERROR_STREAM("Failed to read " << csv << ": Improper formatting detected on line " << line_number << "." << endl << 
+                     "Confirm " << csv << " follows the exact format of example_fault.csv before retrying.");
```
kmdalal, line 159: Please shorten these lines.  Our coding guidelines suggest 80 columns, with a max of 120 if shorter is not practical.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -288,102 +300,86 @@ void PowerSystemNode::injectCustomFault(bool& fault_activated,
...
+      {
+        case ERR_INVALID_FILE_NAME:
+          ROS_WARN_STREAM("Invalid file name '" << designated_file << "' entered; Deactivate fault and try again.");
+          break;
+        case ERR_NO_CSV_FOUND:
+          ROS_WARN_STREAM("Could not find a CSV in the 'profiles' directory with name '" << designated_file << "'. Deactivate fault and try again.");
+          break;
+        default:
+          break;
+      }
```
AstroStucky, line 381: This try-catch does not seem necessary, and I think these two errors are better handled without the use of `throw` inside `loadPowerProfile` itself, which should result in the function returning an empty `PrognoserVector`, which then propagates to `loadCustomFaultPowerProfile` returning false, which then propagates to `injectCustomFault`. And I think that works fine because all `injectCustomFault` really needs to know is whether or not the profile was successfully loaded (`true` / `false`). There's no need to have it be bogged down by the nitty gritty details of error handling that could just as easily be handled inside the function that's closer to the actual file input implementation.

try-catch blocks are useful in the right context, but here it only serves the purpose of separating file input error handling from the actual file input code.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -263,30 +289,135 @@ void PowerSystemNode::injectFault (const string& fault_name,
...
+  if (fault_activated && fault_enabled && custom_fault_ready)
+  {
+    // TODO: Unspecified how to handle end of fault profile. For now, simply disable
+    // the fault from updating any parameters.
+    if (index >= sequence.size())
+    {
+      if (!end_fault_warning_displayed)
+      {
+        ROS_WARN_STREAM
+          (saved_file.substr(0, saved_file.size() - 4) << ": reached end of fault profile. Fault disabled, but will restart if re-enabled.");
```
kmdalal, line 410: Another line to shorten.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -263,30 +289,135 @@ void PowerSystemNode::injectFault (const string& fault_name,
...
+    }
+    catch (int err_val)
+    {
+      switch(err_val)
+      {
+        case ERR_INVALID_FILE_NAME:
+          ROS_WARN_STREAM("Invalid file name '" << designated_file << "' entered; Deactivate fault and try again.");
+          break;
+        case ERR_NO_CSV_FOUND:
+          ROS_WARN_STREAM("Could not find a CSV in the 'profiles' directory with name '" << designated_file << "'. Deactivate fault and try again.");
```
kmdalal, line 377: Another line to shorten.

AstroStucky (changes requested):
I tested the CSV format error line number reporting, and it works great! I think that's a very helpful detail to have added to the error message. 

I had one significant change to request, but everything else looks like lines that were accidentally left in from older versions or debugging. 

Great work so far! It's nearly there.

ow_power_system/src/PowerSystemNode.cpp
AstroStucky: Just to clarify myself, I think the try-catch block inside `loadPowerProfile` that contains the while loop that iterates over file lines is good and should remain. But I feel the one inside `injectCustomFault` is unnecessary so long as `loadPowerProfile` is modified to handle those errors itself.

ow_power_system/src/PowerSystemNode.cpp
AstroStucky: Glad we agree. Just want to chime in that if the ".csv" check is removed then the print statements on lines 356, 360, 410 should be made agnostic to the file extension length.

kmdalal:
The file ow_power_system/config/system.cfg needs an update to the comment for `profile_increment`.   Here's a suggestion:

```
# Power fault profiles (found in ../profiles) are CSV files whose rows
# may represent any time interval.  It is typical for the interval to
# be 1Hz.  To synchronize the consumption of these profiles with
# GSAP's processing rate of .5 Hz, we use every Nth profile line,
# where N = profile interval/gsap_rate.  For now, we avoid additional
# user specification and computation by providing the required value
# of N here.
```

I tried to commit this myself but have misplaced my git key and don't have it on the machine at hand :(

kmdalal:
I have tested a number of different combinations of custom profile and high power draw, including re-running profiles, editing them in between injections, and trying bad file formats.  It's all working as expected.  Aside from the minor code comments I've left I'm satisfied with this work.

Lumsterling:
I've implemented the changes you both requested again. It required a small amount of shifting the code around, which changed some behavior slightly, but I went through and (hopefully) fixed this. One final pass would help to confirm it.

AstroStucky (approved):
This newest version looks good to me. Well done! :confetti_ball:

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -78,65 +78,87 @@ bool PowerSystemNode::loadSystemConfig()
...
   // Skip header line
   file.ignore(numeric_limits<streamsize>::max(), '\n');
 
   auto now = system_clock::now();
 
   PrognoserVector result;
-  while (file.good())
+
+  // Line number starts at 2 instead of 1 because the first line is the header.
+  int line_number = 2;
```
kmdalal, line 100: This will need to be a user requirement to work.   Ideally, the line number would start at the first line of content, however this is fine for this Jira.

kmdalal (approved):
I redid some previous tests, and new ones, involving varied combinations of the two power faults, and all seemed to work correctly.  I've left only some minor additional code suggestions that don't need my re-review if they are addressed.  This is a slick new feature - excellent work!

ow_power_system/src/PowerSystemNode.cpp
Lumsterling: This is based primarily off of me reading files in VS code; it labels each line in the file with a number and starting at 2 means the line number provided in an error message is consistent with the total number of lines in the entire file. I'll leave it as is, but whether or not this starts at 2, 1, or even 0 is primarily up to preference. It can easily be changed later anyways.

PR #256, opened 2022-07-26
Oceanwater 538 ow dynamic terrain modifies normal maps

mogumbo:
## Linked Issues:
| EPIC ⚡ | [OCEANWATER-703](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-703) |
| Jira Ticket 🎟️ | [OCEANWATER-538](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-538) |

## Summary of Changes
* Ogre's internal normal map is applied to any terrain using texture unit `ow_dynamic_terrain_normal_map`. It is updated every time the terrain is modified.

## Test
* You must also use the branch from the ow_europa sibling PR: https://github.com/nasa/ow_europa/pull/20
* Run anything that modifies the atacama or terminator_workspace terrains. You should now see the normals and, consequently, the shading being updated on the terrain. Don't confusing shading with shadows; shadows were already working.

### DIFF
--- a/ow_dynamic_terrain/src/DynamicTerrainVisual.cpp
+++ b/ow_dynamic_terrain/src/DynamicTerrainVisual.cpp
@@ -4,10 +4,12 @@
 
 #include "DynamicTerrainBase.h"
 #include "TerrainModifier.h"
+#include <gazebo/rendering/RenderingIface.hh>
 
 using namespace std;
 using namespace gazebo;
 using namespace rendering;
+using namespace Ogre;
 
 namespace ow_dynamic_terrain
 {
@@ -61,6 +63,41 @@ private:
       terrain->updateDerivedData(false, Ogre::Terrain::DERIVED_DATA_NORMALS | Ogre::Terrain::DERIVED_DATA_LIGHTMAP);
 
       m_differential_pub.publish(diff_msg);
+
+      // Override custom material normal map labeled "ow_dynamic_terrain_normal_map"
+      // with Ogre::Terrain's internal normal map.
+      // Ogre::Terrain makes a copy of a user-specified normal map and the copy
+      // is the one that is updated by Terrain::updateDerivedData(), not the
+      // original normal map specified in a custom material file.
+      // Note: If you are not using a custom material this code will have no
+      // effect, but Ogre should be using the correct normal map and a generated
+      // material and shader in that case.
+      // Caveat: The normal map copy is always resized to the terrain's DEM
+      // resolution, so using a larger texture for more detail is not possible
+      // with this plugin. It might be possible to update the original normal map
+      // by computing all necessary heights at the original resolution and then
+      // computing and uploading normals instead of relying on Ogre to do it. 
+      ResourceManager::ResourceMapIterator res_it = MaterialManager::getSingleton().getResourceIterator();
+      while (res_it.hasMoreElements())
+      {
+        ResourcePtr resource = res_it.getNext();
+        MaterialPtr material = resource.staticCast<Material>();
+        Material::TechniqueIterator tech_it = material->getTechniqueIterator();
+        while (tech_it.hasMoreElements())
+        {
+          Technique* technique = tech_it.getNext();
+          Technique::PassIterator pass_it = technique->getPassIterator();
+          while (pass_it.hasMoreElements())
+          {
+            Pass* pass = pass_it.getNext();
+            TextureUnitState* tus = pass->getTextureUnitState("ow_dynamic_terrain_normal_map");
+            if (tus != 0)
+            {
+              tus->setTexture(terrain->getTerrainNormalMap());
+            }
+          }
+        }
+      }
     }
   }
 
--- a/ow_dynamic_terrain/src/DynamicTerrainVisual.cpp
+++ b/ow_dynamic_terrain/src/DynamicTerrainVisual.cpp
@@ -23,6 +23,13 @@ public:
   void Load(VisualPtr /*visual*/, sdf::ElementPtr /*sdf*/) override
   {
     Initialize("visual");
+
+    m_on_prerender_connection = gazebo::event::Events::ConnectPreRender([this]() {
+      // Initialize terrain with Ogre-generated normal map.
+      applyNormalMap();
+      // Disconnect so this method will not be called again.
+      m_on_prerender_connection.reset();
+    });
   }
 
 private:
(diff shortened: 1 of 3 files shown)


AstroStucky:
Results of running `rostest ow_sim_tests terrain_interaction.test`
Screenshot_20220729_114550
Manually tested on atacama_y1a as well. 

The terrain itself looks notably different under this branch.
| **noetic-devel** branch | this branch |
|-|-|
| Screenshot_20220729_121113 | Screenshot_20220729_115512 |

AstroStucky:
FYI, contributed my own commit to clarify error messages across the two plugins.

kmdalal:
Here's the result of the default LinearDig.  The terrain looks less defined as in @AstroStucky's screenshot.  The trench seems clearer (need to compare with noetic-devel).   

I could use a more detailed description of what to look for (I'm not sure how 'normals' appear), or reference before/after shots...
LinearDig1

ow_dynamic_terrain/src/DynamicTerrainVisual.cpp
```diff
@@ -78,6 +88,53 @@ class DynamicTerrainVisual : public VisualPlugin, public DynamicTerrainBase
...
+      MaterialPtr material = resource.staticCast<Material>();
+      Material::TechniqueIterator tech_it = material->getTechniqueIterator();
+      while (tech_it.hasMoreElements())
+      {
+        Technique* technique = tech_it.getNext();
+        Technique::PassIterator pass_it = technique->getPassIterator();
+        while (pass_it.hasMoreElements())
+        {
+          Pass* pass = pass_it.getNext();
+          TextureUnitState* tus = pass->getTextureUnitState("ow_dynamic_terrain_normal_map");
```
AstroStucky, line 127: [optional] These declarations could be `auto`, but I'm not picky about it :shrug:

AstroStucky (commented):
The test works. Shading definitely looks way better following a dig. No required code changes, but I am curious to hear a response about significant difference in appearance in `atacama_y1a`.

mogumbo:
> The terrain itself looks notably different under this branch.

The original atacama normal map is 2048x2048, but the one generated and applied by Ogre is 1025x1025 so there's less detail in the shading. This is an unfortunate but necessary limitation of updating normals; we can only have normals at the resolution of the terrain geometry.

ow_dynamic_terrain/src/DynamicTerrainVisual.cpp
mogumbo: I prefer to only use auto to replace really complex or unknowable types. Showing types usually makes code easier to understand, in my opinion.

mogumbo:
> I could use a more detailed description of what to look for

That screenshot you posted clearly shows the curve of the bottom of the trench; the side facing the sun the most directly is the brightest. If you dig the same trench without these new branches, you won't see lighting that describes that curved shape. Instead the lighting will look splotchy with a lot of black spots because the normals from the original surface will be applied to the bottom of the trench.

kmdalal:
> The original atacama normal map is 2048x2048, but the one generated and applied by Ogre is 1025x1025 so there's less detail in the shading. This is an unfortunate but necessary limitation of updating normals; we can only have normals at the resolution of the terrain geometry.

I'd like to understand this better.  It seems to me that while the trenches look better, there's a big loss of resolution in the terrain that could degrade image analysis.

mogumbo:
> > The original atacama normal map is 2048x2048, but the one generated and applied by Ogre is 1025x1025 so there's less detail in the shading. This is an unfortunate but necessary limitation of updating normals; we can only have normals at the resolution of the terrain geometry.
> 
> I'd like to understand this better. It seems to me that while the trenches look better, there's a big loss of resolution in the terrain that could degrade image analysis.

When Ogre recomputes surface normals based on the terrain height data, it only knows how to compute normals at the same resolution as the terrain geometry. You can only have one terrain normal map, so if you want to be able to update normals as the terrain changes shape then that normal map must be the same resolution as the terrain geometry.

We are already using a detail normal map to add some extra shading detail, but it looks a little weird using that same detail normal map on the surface and in trenches. The other way we can add detail is with higher-resolution albedo maps. We're already doing that on the atacama terrain.

ow_dynamic_terrain/src/DynamicTerrainVisual.cpp
```diff
@@ -78,6 +88,53 @@ class DynamicTerrainVisual : public VisualPlugin, public DynamicTerrainBase
...
+          if (tus != 0)
+          {
+            tus->setTexture(terrain->getTerrainNormalMap());
```
AstroStucky, line 130: Is this `setTexture` function capable of being handed a normal map that's larger in resolution than the heightmap, or would that throw and error?

AstroStucky:
> When Ogre recomputes surface normals based on the terrain height data, it only knows how to compute normals at the same resolution as the terrain geometry. You can only have one terrain normal map, so if you want to be able to update normals as the terrain changes shape then that normal map must be the same resolution as the terrain geometry.

Am I right to assume that the 2048x2048 normal map we were originally using for atacma_y1a was computed from the FROST dataset?

If the heightmap is all Ogre has to go off of to compute the normal map, then it makes sense that it cannot produce a normal map that's twice its resolution. 

It seems we must either settle for incorrect shading in trenches or a lower visual fidelity workspace environment. But I want to purpose a third, hybrid, option that I suspect would be enough work so as to warrant an additional ticket. Assuming the OGRE API allows for it, the 1025x1025 normal map that's returned by `Ogre::Terrain::getTerrainNormalMap` could be up-sampled and merged with the 2048x2048 normal map. I don't have a good idea for how much this would impact performance, but could it work?

ow_dynamic_terrain/src/DynamicTerrainVisual.cpp
mogumbo: Yes, but I think I'm not being clear. If you use a high-res normal map, and then later modify normal maps (which requires a normal map at the same resolution as terrain geometry) then there will be a visual change across the entire workspace as you switch from a high-res map to a lower-res map.

kmdalal (approved):
After our discussion on the tradeoffs in Dev Day, I'm good with this solution.  I appreciate your perseverance in finding it!

AstroStucky (approved):
Thanks for the discussion on terrain detail. I think it's more important that trenches are rendered correctly than having a 2048x2048 normal map, but I'm glad we discussed some ways to still get that with further development. Namely, the method I discussed earlier or simply using a heightmap that is itself 2048x2048.

Great work! It's nice to see this longstanding issue finally resolved.

PR #257, opened 2022-07-26
 adding goal tolerance such that plan doenot fail for uncertain ground

anjan1984:
## Linked Issues:
| EPIC ⚡ | [OCEANWATER-XXX](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-986) |
| Jira Ticket 🎟️ | [OCEANWATER-XXX](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-986) |
| Github :octocat:  | # |

## Summary of Changes
* Adding a goal tolerance in guarded move, so that planning does not fail when ground is uncertain. 

## Test
* Test guarded move in different locations like: 
*  ./guarded_move_action_client.py 1.78 0.20 -0.12 0.00 0.00 1.00 0.50
*  ./guarded_move_action_client.py

You can also run the Reference mission 1 to verify that guarded move finds the ground.


ow_lander/scripts/action_trajectories.py
```diff
@@ -882,6 +882,7 @@ def guarded_move_plan(self, move_arm, robot, moveit_fk, args, server_stop):
         goal_pose.position.z -= direction_z*search_distance
 
         move_arm.set_pose_target(goal_pose)
+        move_arm.set_goal_tolerance(0.05)
```
kmdalal, line 885: What is the unit for goal tolerance?  If meters, this is about 2 inches.  I have no idea if this is "close enough" according to the sampling location criteria, though it seems reasonable.

anjan1984: The terrain varies a lot in the atacama terrain. This was the smallest tolerance that worked for me

kmdalal:
I still get the same failure in RM1 (see below).  I'm on the correct branch and see that action_trajectories.py has the goal tolerance set.  It's a clean build.  The scoop actually rams into the ground and then wiggles.

```
[INFO]: Guarded move activity started
[Wrn] [Publisher.cc:135] Queue limit reached for topic /gazebo/default/pose/local/info, deleting message. This warning is printed only once.
[ERROR]: arm/arm: Unable to sample any valid states for goal tree
[INFO]: Ground Detected ? False
[INFO]: GuardedMove: Failed
```

kmdalal:
On another attempt I got a different error, but wasn't paying close attention to the arm.
```
[INFO]: Guarded move activity started
[Wrn] [Publisher.cc:135] Queue limit reached for topic /gazebo/default/pose/local/info, deleting message. This warning is printed only once.
[ERROR]: arm/arm: Unable to sample any valid states for goal tree
[INFO]: Ground Detected ? False
[INFO]: GuardedMove: Failed
```

kmdalal:
Here are some GuardedMove invocations from RefMission1 that failed:

Starting GuardedMove(x=1.77, y=0.25, z=-0.11, dir_x=0.00, dir_y=0.00,dir_z=1.00, search_dist=0.50)

Starting GuardedMove(x=1.28, y=0.05, z=-0.06, dir_x=0.00, dir_y=0.00,dir_z=1.00, search_dist=0.50)

mogumbo (changes requested):
I get failures with these:
./guarded_move_action_client.py 1.78 -0.20 -0.12 0.00 0.00 1.00 0.50
./guarded_move_action_client.py 1.5 -0.20 -0.12 0.00 0.00 1.00 0.50

One time the arm contorted badly and then gazebo crashed, but that may be a separate problem.

I thought `set_goal_tolerance()` indicated how far your final position was allowed to be from the intended final position. If that were true, then the tolerance should be set to `search_distance`. But I tried that and it didn't help.

As I recall, finding the ground worked best after Ussama added libgazebo_ros_ft_sensor.so. It might be a good idea to figure out what has changed since then.

kmdalal:
Note that ReferenceMission1 uses image analysis to choose a sampling location, so the calls to GuardedMove will vary.  (I'd like to understand why...lighting conditions at different clock times?).  It may be that it works sometimes.

anjan1984:
Please review this branch again. after checking the plots of FT sensors, i have increased the threshold for ground detection. This simplifies our lives a lot.

kmdalal:
The behavior for the following call was unsatisfactory:

`./guarded_move_action_client.py 1.28 0.05 -0.06 0.00 0.00 1.00 0.50`

The simulator reported:
```
[ERROR]: arm/arm: Unable to sample any valid states for goal tree
[INFO]: Ground Detected ? False
[INFO]: GuardedMove: Failed
```
perhaps rightfully so, but instead of remaining still as I would expect, the scoop made a strange spiraling move towards the ground.

kmdalal:
Another problematic case:

`./guarded_move_action_client.py 1.28 0.05 0.06 0.00 0.00 1.00 0.50`

This time there were no errors, ground was reported as found, and the script returned with success.  However the scoop spinned around a while, ended in a strange position, with the arm bouncing indefinitely (had to kill the sim).

This problem and the one in my last comment were repeats of earlier failed attempts.  I retried all the other reported failed cases (mine and @mogumbo's) and they now work.

anjan1984:
> `./guarded_move_action_client.py 1.28 0.05 0.06 0.00 0.00 1.00 0.50`
I dont see this in my system. Please try the case without goal tolerance. I committed that change so you can just pull. You will get this error though [ERROR]: arm/arm: Unable to sample any valid states for goal tree

kmdalal:
Hmm, both the failures I reported above still fail.   This one fails the same way as described earlier:

`./guarded_move_action_client.py 1.28 0.05 -0.06 0.00 0.00 1.00 0.50`

The following call now ends aborted/failed in both simulator and script, the arm makes the same spirals, but does not end up bouncing:

`./guarded_move_action_client.py 1.28 0.05 0.06 0.00 0.00 1.00 0.50`

Note that the only difference is the sign on the third argument.  I'm using the Atacama world, preceding each call with an Unstow.  I've checked that my branch is current as of 9:30am Wednesday and clean.

mogumbo:
Please merge noetic-devel. This branch is incompatible with ow_europa after my own recent merge.

Previous guarded moves work for me now. However, I also get the strange messages @mikedalal is seeing:
```
[ERROR] [1659636181.012655095, 1916524923.029999971]: arm/arm: Unable to sample any valid states for goal tree
[ WARN] [1659636181.034058839, 1916524923.049999952]: Fail: ABORTED: No motion plan found. No execution attempted.
```

The new guarded moves with x = 1.28 also make a strange spiral maneuver on my computer. I think this is probably a separate problem. It appears we are asking the system to do a guarded move so close to the lander that the arm needs to fold up in a strange way to reach the target. Unfortunately, this can never work because when the arm is folded that way there is a part of the arm that is lower than the scoop and will contact the ground first.

mogumbo:
Tested a little more this morning. Without your most recent commit 598c01152b56183e04b40d70f7bee435fcdd85c2
`​./guarded_move_action_client.py​` touches the ground and reports failure, and then ​`./guarded_move_action_client.py 1.78 0.20 -0.12 0.00 0.00 1.00 0.50`​ lifts the arm off the ground and reports success. So all that behavior is very broken.

With the recent commit both commands report success, but the arm never moves when doing the second command.

So that last commit shouldn't have been pushed as-is. I think you should test a lot more carefully before doing anymore commits, and I think there's still some pretty scary logic problem that needs to be worked out.

mogumbo (approved):
Overall, the behavior is much better now. I can still get the error
`[ERROR] [1660675317.252082576, 1916524902.987499952]: arm/arm: Unable to sample any valid states for goal tree
[ WARN] [1660675317.266900238, 1916524903.002500057]: Fail: ABORTED: No motion plan found. No execution attempted.`
if I try to do `./guarded_move_action_client.py 1.7 -0.8 -0.12 0.00 0.00 1.00 0.50`, but I think that's probably a separate problem that should be handled in another ticket.

AstroStucky:
Almost all my calls to guarded move passed, except this one provided by @kmdalal 
```
./guarded_move_action_client.py 1.28 0.05 -0.06 0.0 0.0 1.0 0.50
```
It resulted in this error 
```
[INFO] [1660685271.201091, 1916525167.465000]: Guarded move activity started
[ERROR] [1660685276.390660169, 1916525172.652499914]: arm/arm: Unable to sample any valid states for goal tree
[ WARN] [1660685276.406348427, 1916525172.667500019]: Fail: ABORTED: No motion plan found. No execution attempted.
[INFO] [1660685295.578143, 1916525191.782500]: Ground Detected ? False
[INFO] [1660685295.583737, 1916525191.830000]: GuardedMove: Failed
```
and the arm twisting itself into this configuration
Screenshot_20220816_152817
I suspect this is an unrelated bug.

Working under the assumption that guarded move is intended to return a failure when the ground is not found within the search distance, I ran some tests with smaller and smaller search distances. No matter the value of search distance, the arm would just proceed downward until the ground is found. Even if I set search distance to zero, this was the case. Isn't the arm supposed to stop and give looking for the ground once it has moved downward farther than the search distance?

AstroStucky:
The following, provided by @mogumbo, also failed.
```
./guarded_move_action_client.py 1.5 -0.20 -0.12 0.00 0.00 1.00 0.50
```
It looked as though the arm continued to push through the terrain, even though contact had already occurred.

AstroStucky (approved):
In general all my tests worked out. The only failures were for cases where the arm twists in a weird way, which has already been pointed out by other reviewers, and is likely unrelated to the bug at hand.

Made some comment typo corrections.

anjan1984:
Thank you for all the comments . Created a jira to capture all the comments 

https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1031

PR #261, opened 2022-08-18
Oceanwater 1006 - Material grid implemented as a Visual Plugin

AstroStucky:
## Linked Issues:
| Jira Ticket 🎟️ | [OCEANWATER-1006](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1006) |
| Github :octocat:  | # |

## Sibling PRs
[ow_europa - PR#21](https://github.com/nasa/ow_europa/pull/21)

## Context
This plugin does not do anything of use yet. This is just foundational work that I'd like to have eyes on. Eventually this visual plugin will be responsible for applying colors/albedo/texture to the terrain corresponding to materials that intersect the heightmap mesh. 

## Summary of Changes
* **MaterialDistributionPlugin** created in the new **ow_materials** package. 
* Created the foundation for volumetricly distributed materials
  * Axis-aligned grid that can be given corner boundaries in world coordinates and populates the volume in between with cells capable of containing a material ID. 
  * Material struct for grouping material parameters. 
  * Material database that is populated by a YAML file in **ow_materials** that defines each material's parameter. 

## Test
1. Switch both **ow_simulator** and **ow_europa** to this branch.
2. View a live feed of gzserver's log using `tail -F ~/.gazebo/server-####/default.log`
3. In another terminal launch the atacama_y1a world `roslaunch ow atacama_y1a.launch`
4. The following output should show in the log
```
(1660852776 254772976) MaterialDistributionPlugin: Material grid centered at (0.65, 0, 0.23) with dimensions 1.3 x 2 x 0.14 meters. 
(1660852776 260661465) MaterialDistributionPlugin: Materials database populated with 3 materials.
(1660852776 260695728) MaterialDistributionPlugin: Successfully loaded!
```
Feel free try other plugin and material parameters by modifying `ow_europa/models/atacama_y1a/model.sdf` or `ow_simulator/ow_materials/materials/default.yaml`, respectively.


AstroStucky:
Placing this PR on-hold while we await the completion of Release 10 and have further discussion about material data structure requirements.

ow_materials/include/Material.h
```diff
@@ -0,0 +1,44 @@
...
+#include <limits>
+
+namespace ow_materials
+{
+
+using MaterialID = uint8_t;
+
+struct Color
+{
+  uint8_t r, g, b;
```
mogumbo, line 17: Use float or double so colors can be blended at high precision. They should only be reduced to uint8 when we want to use them in a texture map.

There is also a chance that material colors differ enough that we would require higher dynamic range and use a floating point texture.

There's also a chance that colors of blended materials would be derived from other properties rather than just directly mixing original colors.

ow_materials/CMakeLists.txt
```diff
@@ -0,0 +1,132 @@
...
+# add_executable(${PROJECT_NAME}_node src/dynamic_terrain_node.cpp)
+
+## Rename C++ executable without prefix
+## The above recommended prefix causes long target names, the following renames the
+## target back to the shorter version for ease of user use
+## e.g. "rosrun someones_pkg node" instead of "rosrun someones_pkg someones_pkg_node"
+# set_target_properties(${PROJECT_NAME}_node PROPERTIES OUTPUT_NAME node PREFIX "")
+
+# Gazebo plugin that gives a terrain a material distribution
+add_library(MaterialDistributionPlugin
```
mogumbo, line 95: Since all you are building is a plugin, there's a strong argument that this should be part of ow_gazebo_plugins instead of in a new package.

ow_materials/include/MaterialDistributionPlugin.h
```diff
@@ -0,0 +1,37 @@
...
+{
+  public:
+  void Load(gazebo::rendering::VisualPtr parent, sdf::ElementPtr sdf);
+
+  private:
+  gazebo::rendering::VisualPtr m_visual;
+
+  std::unique_ptr<MaterialDatabase> m_material_db;
+
+  std::unique_ptr<AxisAlignedGrid<MaterialID>> m_grid;
```
mogumbo, line 29: In order to blend between materials, I think this should be a grid of a type other than MaterialID. Maybe something like this:

`class MaterialBlend {
  // How much of each material type.
  // This could take up a lot of space if there were 256 materials, but having the ability to blend materials means we should be able to keep the material count low.
  float m_fractions[num_materials];

  // You might want terramechanics properties like this derived from the blended material that are required during grinding and digging. This would take more space in the data structure but would save computation time if there are frequent queries.
  float m_hardness;

  // We probably don't need a color member because those would be computed once and put into a 3D texture map.

  // Make all fractions sum to 1.0
  void normalize();
};`

ow_materials/include/AxisAlignedGrid.h
```diff
@@ -0,0 +1,63 @@
...
+
+#include <vector>
+#include <memory>
+
+#include <ignition/math/AxisAlignedBox.hh>
+
+namespace ow_materials {
+
+template <typename T>
+class AxisAlignedGrid {
```
mogumbo, line 16: Are there existing data structures that could be used here, such as an OpenCV Mat? Some investigation might reveal useful features you did not think to add to this class.

mogumbo (changes requested):

ow_materials/include/AxisAlignedGrid.h
mogumbo: In addition to OpenCV Mat there is Boost MultiArray. We would want to do interpolation (weighted averages of 8 voxels), but it doesn't appear that either of these data structures would help with that. It might still be worth evaluating these libraries for performance and API convenience.

ow_materials/include/AxisAlignedGrid.h
```diff
@@ -0,0 +1,63 @@
...
+    return m_cells[index(i, j, k)];
+  };
+
+private:
+
+  inline size_t index(size_t const i, size_t const j, size_t const k) const {
+    assert(i < m_dimensions.X());
+    assert(j < m_dimensions.Y());
+    assert(k < m_dimensions.Z());
+    return i + j * m_dimensions.X() + k * m_dimensions.X() * m_dimensions.Y();
```
kmdalal, line 47: I'm always leery of `assert` in its ability to crash the system.  Is it perhaps overkill for checking dimensions?

ow_materials/CMakeLists.txt
AstroStucky: Following up on our discussion in Teams, I'm open to moving all of this into ow_gazebo_plugins assuming the multi-material terrain system winds up being only Gazebo plug-ins. As we discussed, for now we'll keep it as its own package and move later as needed.

ow_materials/include/Material.h
AstroStucky: This is why I am glad to have your eyes on this. I'll change them to double, thanks.

ow_materials/include/AxisAlignedGrid.h
AstroStucky: I get the impression there is a particular time and place for `assert`, but it looks to me our project's build system does not differentiate between debug and release builds, which is where assert is useful. In a release build, an `assert` does nothing because it is only intended as a debug tool. I do not know if our build system builds with `NDEBUG` defined or not. Additionally, there is a chance the public method that provides `i`, `j`, and `k` will be callable by the user, so there is a chance this should be a user input error as opposed to an internal error, which is the only kind `assert` should ever check for. 

I'll remove them for now, but may replace them with an if-statement and exception throw later as I come to better understand who exactly needs to call `getCellValue`.

_Source_ https://stackoverflow.com/questions/12062365/is-using-assert-in-c-bad-practice

AstroStucky:
noetic-devel has been merged into the epic branch and this branch. Suggested changes that were low effort have been incorporated.

ow_materials/include/AxisAlignedGrid.h
AstroStucky: `assert`s removed

ow_materials/include/Material.h
AstroStucky: Changed to double

ow_materials/include/AxisAlignedGrid.h
mikedalal: Thanks and good question - I don't know if a debug mode as you describe is supported in our build.  You could create a Jira, and this may be a good intern task.

kmdalal (approved):
This simply duplicates my approval under my other account -- will try to stick to one in the future!

AstroStucky:
I have merged noetic-devel into this and the epic branch, and addressed all comments. The major change since the previous review is that contents of a grid cell are now a blend of MaterialIDs, rather than a single MaterialID, as suggested by @mogumbo.

This, and its sibling PR [ow_europa #21](https://github.com/nasa/ow_europa/pull/21) are now ready for a re-review.

kmdalal:
Here's more of the log -- see if there are problems, I can't tell...
```
(1686156541 125484025) MaterialDistributionPlugin: Material grid centered at (0.65, 0, 0.23) with dimensions 1.3 x 2 x 0.14 meters. 
(1686156541 146668518) MaterialDistributionPlugin: Materials database populated with 3 materials.
(1686156541 146815504) MaterialDistributionPlugin: Successfully loaded!
(1686156541 148849847) [Wrn] [SystemPaths.cc:459] File or path does not exist [""] [__default__]
(1686156541 149124961) [Wrn] [SystemPaths.cc:459] File or path does not exist [""] [__default__]
(1686156541 173353487) [Err] [Dem.cc:236] EXCEPTION: Unable to transform terrain coordinate system to WGS84 for coordinates (0,0)

(1686156541 173515220) [Wrn] [Dem.cc:133] Failed to automatically compute DEM size. Please use the <size> element to manually set DEM size.
(1686156541 239936071) [Msg] Loading heightmap: atacama_y1a_0.0048m
(1686156541 240208158) [Msg] Loading heightmap cache data: /home/kdalal/.gazebo/paging/atacama_y1a_0.0048m/gazebo_terrain_LOD0_00000000.dat
(1686156541 844325748) [Msg] Heightmap loaded. Process took: 0.604375 seconds
(1686156541 858921494) ContactSensorPlugin: reporting only link names that match the regex pattern "regolith_\d*.*"
(1686156541 859590717) ContactSensorPlugin: Contacts will be reported on ROS topic /ow_regolith/contacts/sample_dock
```

ow_materials/include/AxisAlignedGrid.h
```diff
@@ -0,0 +1,60 @@
...
+
+  inline size_t index(size_t const i, size_t const j, size_t const k) const {
+    return i + j * m_dimensions.X() + k * m_dimensions.X() * m_dimensions.Y();
+  };
+
+  std::unique_ptr<ignition::math::AxisAlignedBox> m_domain;
+
+  ignition::math::Vector3<size_t> m_dimensions;
+
+  double const m_side_length;
```
kmdalal, line 51: I can't find any uses or initialization of this constant (and it could be initialized here if used).

kmdalal: Never mind, I found the `.tpp` file - still not used to those!

I didn't think a const could be initialized at runtime.  Even if this works, I suggest removing `const` because one typically assumes these are initialized statically.

AstroStucky: See AxisAlignedGrid.tpp:19 and the definition of that constructor. 

It is used only in the constructor, so you could argue it could just be a constructor parameter and not a member variable, but I anticipate class methods not yet defined will use it. e.g. This class still needs a method that takes a point and returns the cell or MaterialBlend of the cell that point exists in. This computation would require knowing the side length of cells.

AstroStucky:
> Here's more of the log -- see if there are problems, I can't tell...
> 
> ```
> (1686156541 125484025) MaterialDistributionPlugin: Material grid centered at (0.65, 0, 0.23) with dimensions 1.3 x 2 x 0.14 meters. 
> (1686156541 146668518) MaterialDistributionPlugin: Materials database populated with 3 materials.
> (1686156541 146815504) MaterialDistributionPlugin: Successfully loaded!
> (1686156541 148849847) [Wrn] [SystemPaths.cc:459] File or path does not exist [""] [__default__]
> (1686156541 149124961) [Wrn] [SystemPaths.cc:459] File or path does not exist [""] [__default__]
> (1686156541 173353487) [Err] [Dem.cc:236] EXCEPTION: Unable to transform terrain coordinate system to WGS84 for coordinates (0,0)
> 
> (1686156541 173515220) [Wrn] [Dem.cc:133] Failed to automatically compute DEM size. Please use the <size> element to manually set DEM size.
> (1686156541 239936071) [Msg] Loading heightmap: atacama_y1a_0.0048m
> (1686156541 240208158) [Msg] Loading heightmap cache data: /home/kdalal/.gazebo/paging/atacama_y1a_0.0048m/gazebo_terrain_LOD0_00000000.dat
> (1686156541 844325748) [Msg] Heightmap loaded. Process took: 0.604375 seconds
> (1686156541 858921494) ContactSensorPlugin: reporting only link names that match the regex pattern "regolith_\d*.*"
> (1686156541 859590717) ContactSensorPlugin: Contacts will be reported on ROS topic /ow_regolith/contacts/sample_dock
> ```

This is a pass. The other warnings and errors have been there for as long as I can remember.

ow_materials/include/AxisAlignedGrid.h
```diff
@@ -0,0 +1,60 @@
...
+  AxisAlignedGrid() = delete;
+  AxisAlignedGrid(const AxisAlignedGrid&) = delete;
+  AxisAlignedGrid& operator=(const AxisAlignedGrid&) = delete;
+
+  const ignition::math::Vector3d &getDiagonal() const;
+  const ignition::math::Vector3d &getMaxCorner() const;
+  const ignition::math::Vector3d &getMinCorner() const;
+  const ignition::math::Vector3d &getCenter() const;
+
+  inline T &getCellValue(size_t const i, size_t const j, size_t const k) {
```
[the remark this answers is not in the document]
AstroStucky: The general consensus I read is to `const` anything and everything that is truly constant. I can understand how it might seem excessive here, especially for a single line inlined function, but using `const` can also assist the compiler in making optimizations. This is also one of those low-level methods that will be called often and more than any other method defined in this PR, so I would really prefer to be as rigorous in its definition as possible.

ow_materials/include/AxisAlignedGrid.tpp
```diff
@@ -0,0 +1,80 @@
...
+static void increaseToNearestMultiple(double &out_x, double const divisor) {
+  out_x += std::remainder(out_x, divisor);
+}
+
+template <typename T>
+AxisAlignedGrid<T>::AxisAlignedGrid(
+  double const x0, double const y0, double const z0,
+  double const x1, double const y1, double const z1,
+  double const side_length, T const initial_value
+) : m_side_length(side_length)
```
kmdalal, line 19: I'll mention again that I think the superfluous `const` declarations should go.  A more important concern is to check for `side_length` not being 0, otherwise it gets passed as a divisor later on.

ow_materials/include/AxisAlignedGrid.tpp
```diff
@@ -0,0 +1,80 @@
...
+
+  // compute number of cells in each dimension
+  m_dimensions = ignition::math::Vector3<size_t>(
+    static_cast<size_t>(std::round(diagonal.X() / m_side_length)),
+    static_cast<size_t>(std::round(diagonal.Y() / m_side_length)),
+    static_cast<size_t>(std::round(diagonal.Z() / m_side_length))
+  );
+
+  // fill cells with initial value
+  // TODO: chance of int overflow?
```
kmdalal, line 49: Perhaps define a "reasonable" maximum size for the volume, or side length, and check against that.

ow_materials/include/AxisAlignedGrid.h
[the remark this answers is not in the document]
kmdalal: I strongly disagree with said consensus, in favor of code compactness and readability, but I don't want to argue this point.  Note that since this is a trivial inline function, I can't see any meaningful compiler optimization here.

ow_materials/include/AxisAlignedGrid.tpp
```diff
@@ -0,0 +1,80 @@
...
+
+template <typename T>
+const ignition::math::Vector3d &AxisAlignedGrid<T>::getMaxCorner() const {
+  return m_domain->Max();
+};
+
+template <typename T>
+const ignition::math::Vector3d &AxisAlignedGrid<T>::getMinCorner() const {
+  return m_domain->Min();
+};
```
kmdalal, line 72: Wondering why Min() and Max() are not cached in a static function variable as is Center() below.  They should all be consistent, unless there's a difference.

ow_materials/include/MaterialDatabase.h
```diff
@@ -0,0 +1,42 @@
...
+
+  MaterialDatabase(const MaterialDatabase&) = delete;
+  MaterialDatabase& operator=(const MaterialDatabase&) = delete;
+
+  void addMaterial(Material mat);
+
+  size_t size();
+
+  inline const Material &getMaterial(MaterialID id) const {
+    return m_database.at(id);
```
kmdalal, line 31: Assuming you're certain that `id` will be valid, else the `at` would throw an exception.  Minor point: `id` should be `const` per your philosophy though I certainly prefer it as is!

ow_materials/package.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>ow_materials</name>
+  <version>0.1.0</version>
+  <description>
+    <p>A Gazebo plugin that distributes a multitude of materials throughout a terrain that alter the terrain visually and terramechanically when operated on by an end-effector.</p>
```
kmdalal, line 6: Please shorten this line.

ow_materials/include/MaterialDatabase.h
```diff
@@ -0,0 +1,42 @@
...
+class MaterialDatabase
+{
+public:
+  MaterialDatabase() = default;
+  ~MaterialDatabase() = default;
+
+  MaterialDatabase(const MaterialDatabase&) = delete;
+  MaterialDatabase& operator=(const MaterialDatabase&) = delete;
+
+  void addMaterial(Material mat);
```
kmdalal, line 26: Seems that argument would be better as const&.

ow_materials/src/MaterialDatabase.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+#include <MaterialDatabase.h>
+
+using namespace ow_materials;
+
+void MaterialDatabase::addMaterial(Material mat) {
+  static MaterialID id = Material::id_min;
+  if (id == Material::id_max) {
+    // Unsure whether this should use gazebo or ROS logging, so both error
+    // message are left as TODOs for now
```
kmdalal, line 13: Minor suggestion: use one of the logging methods for now, just so the error is reported.

ow_materials/src/MaterialDatabase.cpp
```diff
@@ -0,0 +1,26 @@
...
+  if (!m_database.insert({id, mat}).second) {
+    // TODO Error: An unknown error has occurred. This should never happen.
+    return;
+  }
+  ++id;
+}
+
+size_t MaterialDatabase::size() {
+  return m_database.size();
+}
```
kmdalal, line 26: Minor suggestion: trivial enough to be inlined.

ow_materials/src/populate_materials.cpp
```diff
@@ -0,0 +1,80 @@
...
+using std::string, std::begin, std::end, std::initializer_list, std::set,
+      std::vector, std::stringstream, std::numeric_limits, std::runtime_error;
+
+using namespace ow_materials;
+
+static string join(const initializer_list<string> &args) {
+  stringstream joined;
+
+  for (auto x = begin(args); x != end(args) - 1; ++x)
+    joined << *x << "/";
```
kmdalal, line 24: Suggest bracketing for safety.  Same for one-line IF statements in rest of file.

ow_materials/include/AxisAlignedGrid.tpp
AstroStucky: Cube root of the max size_t seems reasonable enough for the max size of a single dimension, and I really doubt a single dimension will ever need to be that large anyways.

ow_materials/include/AxisAlignedGrid.tpp
AstroStucky: I did not think they took any computation, but I cannot confirm or deny that, so I went ahead and static cached them. 

This static caching technique is nice, but kinda new to me. Are there caveats I should be aware of?

kmdalal: Values should be cached only if they will never change, _and_ require non-trivial computation.  There's no risk I'm aware of.  

 I made this suggestion only because I saw Center() cached and thought Min() and Max() may be similar.   Your call on what's best.

ow_materials/include/AxisAlignedGrid.h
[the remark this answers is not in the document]
kmdalal: My main view on this matter is that arguments passed by value are _copies_, and I see no reason to force immutability on copies, and enforce an implementation detail that is invisible to callers and in principle (to me) no one should even care about.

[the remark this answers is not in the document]
AstroStucky: I have removed `const` from inline function parameters because I will concede the benefit there is minimal if not zilch to the best of my knowledge. I have also removed `const` from the prototype of `AxisAlignedGrid`'s constructor, but kept it in the implementation of the function. This way users don't have to see `const` cluttering up the header file, but the actual implementation still has the benefit of it. 

I stick to the stance that it is _always_ better to have the compiler throw an error when a variable you expect to to never mutate is being mutated. There are many reasons this may happen ranging from a typo to a bug in an API. The larger the scope of a variable, the more possibility of unwarranted mutation, and therefore the bigger the benefit of making the variable `const`.

ow_materials/include/AxisAlignedGrid.tpp
AstroStucky: I addressed this in the AxisAlignedGrid.h:32 comment.

ow_materials/include/MaterialDatabase.h
AstroStucky: I added a comment that notes this may throw an `std::out_of_range`. Presently the method is not called anywhere, but I do intend for it to only ever be called inside of a `try` block.

ow_materials/package.xml
AstroStucky: done

ow_materials/include/MaterialDatabase.h
AstroStucky: done

ow_materials/src/MaterialDatabase.cpp
AstroStucky: You're right. The redesign made it clear this is primarily a Gazebo plugin, so I will use Gazebo logging.

ow_materials/src/MaterialDatabase.cpp
AstroStucky: done

ow_materials/src/populate_materials.cpp
AstroStucky: done

ow_materials/include/AxisAlignedGrid.h
AstroStucky: I experimented more with asserts in another function, and it turns out they really aren't good to use in Gazebo plugins, and I think that makes sense since they are dynamically loaded libraries. Whenever they are triggered, gazebo will crash, but the message contained by the exception is nowhere to be found in the log file or simulation terminal. Maybe it winds up in another log file, but I think that indicates it's much better to use _only_ Gazebo's logging system when in code operates inside a plugin and to catch absolutely everything that is being thrown.

ow_materials/include/AxisAlignedGrid.h
AstroStucky: it is removed

AstroStucky:
This is now ready for another review.

kmdalal (approved):
For some reason, I see my code comments in the Conversation tab but not inline in the code files, which would be more useful.  I have resolved the conversations in the Conversation tab.  The given test passed, and I ran RM1 for regression even though I was sure nothing should have been affected.  Thanks for addressing all my comments, in particular the functionality ones -- most were on style.   Great work!

PR #262, opened 2022-08-18
Oceanwater 1003 multiple power cell support

Lumsterling:
## Linked Issues:
| Jira Ticket 🎟️ | [OCEANWATER-1003](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1003) |
| Github :octocat:  | Lumsterling |

## Summary of Changes
* Previously, a single-cell [simple prognoser](https://github.com/nasa/GSAP/tree/master/examples/simple) was used to get end-of-discharge (EoD) predictions every GSAP cycle (0.5 Hz). With the switch to a multi-cell battery pack model, the previous system will no longer work. This ticket implemented the first step of the battery pack: [asynchronous prognosers](https://github.com/nasa/GSAP/tree/master/examples/async) that allow multiple EoD predictions to be obtained in a single GSAP cycle.
  * A new class was created for this (_PowerSystemPack_) as well as a _PredictionHandler_ class based off _PredictionPrinter_ in the asynchronous prognoser example.
* Note that overall, this JIRA is non-functional: values are not yet published properly for other components of the simulator which will likely cause issues, and not all components that are used to get prediction input values (voltage/wattage/temperature) are necessarily computed correctly with the switch. This subticket of the EPIC focuses on the ability for multiple EoD predictions to be obtained in a single GSAP cycle.

## Test
* catkin build
* roslaunch ow_europa_terminator_workspace.launch
* A large amount of debug statements should continually display:
  * These statements should show the time, power, voltage, and temperature for each of the 8 nodes that are instantiated. Time should increment by 2 between cycles to sync with the 0.5 Hz rate of GSAP.
  * After waiting for all buses to return with EoD predictions (consisting of remaining useful life (RUL), state of charge (SOC), and battery temperature (TMP), they should be displayed for each node
* The timing of these displays should line up such that all 8 nodes provide EoD predictions every cycle, rather than only 1 node per cycle updating its data.
* The values of each node should be nearly identical to each other (particularly temperature) since their inputs are identical. Random variance may affect input voltage and output RUL more than the other data points.
* Note that the informational debug statements may print while the simulation is starting up, and these statements will be invalid (and to my understanding, irrelevant).
* **Additional tests can be performed by altering the `NUM_NODES` constant in `PowerSystemPack.h`, then rebuilding and re-running the simulation.** This alters the number of parallel nodes in the simulated pack, but it has a significant effect on runtime (mentioned below).

## Potential Issue(s)
* #include statements and global variables are not cleaned up; some may be irrelevant
* **The responsiveness of GSAP seems to tank significantly as more parallel nodes are added.** When `NUM_NODES` is set to 8, it takes around 2-2.5s to get all 8 EoD predictions, but when `NUM_NODES` is set to 24 (to my recollection, the number of parallel nodes Chetan said he wanted to simulate) it takes anywhere from 4-6s to get all 24 EoD predictions back. This could be a major issue when the time increment for each set of data is 2 seconds. The exact source of the lag is unknown at the moment, but it seems like GSAP's performance may be the issue; if this is the case, there isn't very much I can do about it.
* I'm not very happy with some of my methods for implementing this feature, but could not get it to work in other ways:
  * ~~Because there is a second class (_PredictionHandler_) that contains the override function when a message bus returns with data, I had to use a global variable to allow for both classes to access the information at the same time. I'm not a fan of global variables though I'm unsure if that's industry standard or otherwise.~~
    * This has been addressed in a recent commit.
  * The asynchronous prognoser does not have a public _getModel()_ function like the simple prognoser does, which is needed to get some output values like battery temperature, and I can't change this because it's part of GSAP and not ow_simulator. I worked around it by declaring a simple prognoser with the same initializing data and just pulled its model instead, but this is largely inefficient when a _getModel()_ function would address this instead.
  * ~~My hope had been to streamline initialization such that the number of battery nodes could be controlled by a global constant, but constructing _PredictionHandlers_ and _AsyncPrognosers_ does not seem scalable, resulting in code that requires the battery nodes to be 8 specifically (and the code must be updated if more/less nodes are required). I feel like there must be a way to initialize these objects into an array or other scalable data structure, but I couldn't get it to work. Attempting to do so with a vector or array caused the intended behavior to break.~~
    * This has been addressed in a recent commit.


ow_power_system/src/PowerSystemPack.cpp
```diff
@@ -0,0 +1,375 @@
...
+      std::vector<std::shared_ptr<DoubleMessage>> data_to_publish;
+
+      data_to_publish.push_back(
+        std::make_shared<DoubleMessage>(MessageId::Watts, node_names[i], timestamp, models[i][1]));
+      data_to_publish.push_back(
+        std::make_shared<DoubleMessage>(MessageId::Centigrade, node_names[i], timestamp, models[i][3]));
+      data_to_publish.push_back(
+        std::make_shared<DoubleMessage>(MessageId::Volts, node_names[i], timestamp, models[i][2]));
+
+      std::this_thread::sleep_until(timestamp);
```
AstroStucky, line 245: What is the purpose of this?

ow_power_system/src/PowerSystemPack.cpp
```diff
@@ -0,0 +1,375 @@
...
+
+const auto START_TIME       = MessageClock::now();
+//const int NUM_NODES         = 8; // NOTE: Should equal m_num_nodes in PowerSystemPack.h
+
+// The index use to access temperature information.
+// This might change to median SOC or RUL index or fixed percentile.
+//
+static constexpr int TEMPERATURE_INDEX = 1;
+
+double EoD_events[NUM_NODES][3];
```
AstroStucky, line 40: I don't like this as a global variable. There are ways to keep it internal to the classes, and that should be preferred. The first that comes to mind is to make EoD_events a data struct, for example
```C++
struct EoDEvents {
  double rul;
  double soc;
  double temperature;
}
```
Then make this struct a member of the `PowerNodeHandler` object I proposed earlier. Then pass a reference to the `PowerNodeHanlder`'s instance of `EoDEvents` into `PredictionHandler::processMessage`, so the function can directly write data to it.

AstroStucky (changes requested):
Please see my code comments. We can discuss them over a call if you would like. I have not yet ran the test, so I will return later with results.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -527,15 +548,23 @@ void PowerSystemNode::Run()
...
-  if (!psn.Initialize())
+  if (m_trigger_processing_new_power_batch)
   {
-    ROS_ERROR("Power system node failed to initialize");
-    return -1;
+    m_trigger_processing_new_power_batch = false;
+    m_processing_power_batch = true;
+    runPrognoser(m_mechanical_power_to_be_processed / m_efficiency);
+    m_processing_power_batch = false;
   }
```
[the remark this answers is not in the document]
Lumsterling: Assuming you're talking about Run() and RunOnce() being essentially the same function (minus a loop), the plan was to remove Run() as it's no longer used; I can do that now. If that's not what you're referring to, let me know.

ow_power_system/src/PowerSystemPack.cpp
```diff
@@ -0,0 +1,375 @@
...
+
+// The PredictionHandler class subscribes to the battery EoD event message and
+// prints each event as it is received.
+class PredictionHandler : public IMessageProcessor
```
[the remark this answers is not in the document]
Lumsterling: I think moving it to its own source & header file is ideal, though at the moment it uses a global variable within PowerSystemPack.cpp (EoD_events[][]) to share the EoD predictions with the PowerSystemPack class. I don't like this current solution, and I do think cleaning it up and splitting the classes into 2 files would be better, but at the moment it would interfere significantly with the implementation. Maybe I could post this change as a JIRA ticket later down the line?

ow_power_system/src/PowerSystemPack.cpp
```diff
@@ -0,0 +1,375 @@
...
+    {
+      std::cerr << "Unexpected uncertainty type for EoD prediction" << std::endl;
+      return std::exit(1);
+    }
+
+    // valid prediction
+    // Determine the median RUL.
+    auto samplesRUL = eod_time.getVec();
+    sort(samplesRUL.begin(), samplesRUL.end());
+    double eod_median = samplesRUL.at(samplesRUL.size() / 2);
```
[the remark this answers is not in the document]
Lumsterling: This code regarding the median RUL/SoC/battery temp calculations was taken from the previous writer, so I didn't modify it at first in case it caused issues with the output that I wouldn't be able to detect. It should be simple enough to update though, so I'll take a look.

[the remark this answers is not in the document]
Lumsterling: I wrote a simple function to find the median using `nth_element`, but until I get a confirmation on whether or not the original behavior was intended, I'll keep using the old code. I'll continue going through this review tomororw.

ow_power_system/src/PowerSystemNode.cpp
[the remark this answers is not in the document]
AstroStucky: Either now or in a future PR onto Epic branch.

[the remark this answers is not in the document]
Lumsterling: This comment was related to the unused Run() function, right? Now that it's removed, it shouldn't be a problem anymore.

ow_power_system/src/PowerSystemPack.cpp
```diff
@@ -0,0 +1,375 @@
...
+    // Determine the median SOC.
+    UData currentSOC = eod_event.getState()[0];
+    auto samplesSOC = currentSOC.getVec();
+    sort(samplesSOC.begin(), samplesSOC.end());
+    double soc_median = samplesSOC.at(samplesSOC.size() / 2);
+    //soc_msg.data = soc_median;
+
+    // Determine the Battery Temperature
+    auto stateSamples = eod_event.getSystemState()[0];
+    std::vector<double> state;
```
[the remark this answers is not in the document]
Lumsterling: This was code from the previous writer, but it's not hard to change a variable name. I think `temperature_state` would work fine.

ow_power_system/src/PowerSystemPack.cpp
```diff
@@ -0,0 +1,375 @@
...
+    {
+      state.push_back(sample[0]);
+    }
+
+    // HACK ALERT:
+    // The asynchronous prognoser does not have a function to get its model like
+    // the simple prognoser does. As such, one can instantiate a simple prognoser
+    // and get its model; it should be identical. However, a getter function would
+    // be much more ideal and more efficient.
+    // Construct a simple prognoser to get the battery model.
```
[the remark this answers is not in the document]
Lumsterling: You'd have to ask Chetan about that; I haven't heard anything from GSAP developers myself.

ow_power_system/src/PowerSystemPack.cpp
```diff
@@ -0,0 +1,375 @@
...
+  {
+    using namespace std::chrono;
+    // The prediction handler only ever subscribes to the BatteryEoD message
+    // id, which should always be a ProgEventMessage, so this should always
+    // succeed.
+    auto prediction_msg = dynamic_cast<ProgEventMessage*>(message.get());
+    if (prediction_msg == nullptr)
+    {
+      std::cerr << "Failed to cast prediction message to expected type" << std::endl;
+      std::exit(1);
```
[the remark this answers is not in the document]
Lumsterling: That was part of the asynchronous example I pulled from; should be a minor change.

ow_power_system/src/PowerSystemPack.cpp
```diff
@@ -0,0 +1,375 @@
...
+    // Store the newly obtained data.
+    EoD_events[node_number][0] = rul_median;
+    EoD_events[node_number][1] = soc_median;
+    EoD_events[node_number][2] = model_output[TEMPERATURE_INDEX];
+  }
+
+private:
+  MessageBus& bus;
+  std::string identifier = "Node X";
+  int node_number = -1;
```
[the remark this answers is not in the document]
Lumsterling: Mistake on my part; originally most of these variables were declared in the main function, so they didn't have the `m_` convention. When I moved them over, I forgot to change that.

ow_power_system/src/PowerSystemPack.cpp
Lumsterling: Assuming you're talking about the `sleep_until` line, it was part of the original asynchronous prognoser example. I just tested with and without the line and noticed no difference, so it's probably safe to remove. It was something I simply glossed over.

ow_power_system/src/PowerSystemPack.cpp
Lumsterling: Now that I've reached the end of the non-optional comments, I'll start tackling the overhaul involving all the optional ones. I agree that the current implementation is quite flawed.

However, to my understanding I can't modify the parameters to `PredictionHandler::processMessage`; I believe the `MessageBus` that calls it is handled by GSAP and would require a modification on their end to work with it. One of the bigger hurdles I'm considering is how to get such a reference into that function since that was the primary reason I resorted to a global variable for the first implementation anyways.

ow_power_system/src/PowerSystemPack.cpp
```diff
@@ -0,0 +1,375 @@
...
+  // Create the prognosers using the builder that will send predictions using
+  // their corresponding message bus.
+  PCOE::AsyncPrognoser prognoser_0 = builder.build(bus[0], node_names[0], "trajectory");
+  PCOE::AsyncPrognoser prognoser_1 = builder.build(bus[1], node_names[1], "trajectory");
+  PCOE::AsyncPrognoser prognoser_2 = builder.build(bus[2], node_names[2], "trajectory");
+  PCOE::AsyncPrognoser prognoser_3 = builder.build(bus[3], node_names[3], "trajectory");
+  PCOE::AsyncPrognoser prognoser_4 = builder.build(bus[4], node_names[4], "trajectory");
+  PCOE::AsyncPrognoser prognoser_5 = builder.build(bus[5], node_names[5], "trajectory");
+  PCOE::AsyncPrognoser prognoser_6 = builder.build(bus[6], node_names[6], "trajectory");
+  PCOE::AsyncPrognoser prognoser_7 = builder.build(bus[7], node_names[7], "trajectory");
```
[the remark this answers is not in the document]
Lumsterling: I talked with Chetan about this and the `PredictionHandler` declarations the other day; he said that it's alright for now, but we should return to it later once the full feature is functional. I'll work on organizing the other parallel variables into structs, but I think I'll leave this for now given it's a significantly larger change.

ow_power_system/src/PowerSystemPack.cpp
AstroStucky: I'm not sure what's the right answer is here, but I suspect it's trying to accomplish something similar to what `ros::Rate::sleep` would do. If that is so, it'd be best to use the ROS function.

Lumsterling: I think it was needed in the original example where only one prognoser was declared and the loop did not contain the `waitAll` command (which only ran after the loop terminated). However, since I'm using multiple prognosers and buses here and the `waitAll()` command is within the overall loop, this sleep function isn't actually needed anymore. I could be wrong, but given I saw no noticeable behavior change upon removing the line, that's my hunch.

ow_power_system/src/PowerSystemPack.cpp
AstroStucky: > However, to my understanding I can't modify the parameters to `PredictionHandler::processMessage`; I believe the `MessageBus` that calls it is handled by GSAP and would require a modification on their end to work with it. One of the bigger hurdles I'm considering is how to get such a reference into that function since that was the primary reason I resorted to a global variable for the first implementation anyways.

That's right, I forgot the function's prototype is API defined. But that shouldn't be a problem. You'll just have to pass a reference to `EoD_Events` into `PredictionHandler`, so it can save it as a reference member variable, which can then be modified each time `processMessage` is called.

Lumsterling: That's actually what I spent this morning trying to do, but you can't declare an uninitialized reference variable, so I wasn't able to save it outside of the constructor. I'll take another pass at it though; at least I know I was thinking along the right track.

Lumsterling: I've got this working now; the global variable is no longer needed. I can separate `PredictionHandler` into its own source file now that it no longer depends on a global var (requested in a different comment), so I'll get on that.

ow_power_system/src/PowerSystemPack.cpp
```diff
@@ -0,0 +1,375 @@
...
+
+  // Construct the prediction handlers.
+  PredictionHandler handler_0(bus[0], node_names[0]);
+  PredictionHandler handler_1(bus[1], node_names[1]);
+  PredictionHandler handler_2(bus[2], node_names[2]);
+  PredictionHandler handler_3(bus[3], node_names[3]);
+  PredictionHandler handler_4(bus[4], node_names[4]);
+  PredictionHandler handler_5(bus[5], node_names[5]);
+  PredictionHandler handler_6(bus[6], node_names[6]);
+  PredictionHandler handler_7(bus[7], node_names[7]);
```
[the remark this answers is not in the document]
Lumsterling: I finally figured out how to declare both `PredictionHandlers` and `AsyncPrognosers` within their own loops, so now the code properly supports multiple nodes and all it takes is editing the `NUM_NODES` constant. While it would still be better to add `PredictionHandler`s and `AsyncPrognoser`s to another class, I'm going to leave it as-is and it can be revisited in a future PR.

ow_power_system/src/PowerSystemPack.cpp
```diff
@@ -167,22 +54,21 @@ void PowerSystemPack::InitAndRun()
...
-  PredictionHandler handler_6(bus[6], node_names[6]);
-  PredictionHandler handler_7(bus[7], node_names[7]);
+  PredictionHandler *handlers[NUM_NODES];
+  for (int i = 0; i < NUM_NODES; i++)
+  {
+    handlers[i] = new PredictionHandler(EoD_events[i].remaining_useful_life,
+                                        EoD_events[i].state_of_charge,
+                                        EoD_events[i].battery_temperature,
+                                        m_nodes[i].bus, m_nodes[i].name, i);
+  }
```
AstroStucky, line 71: [required]
For each use of `new` you should plan to also call `delete` when the object is no longer needed, _i.e._ in the destructor. Otherwise you have a memory leak.

If the elements must be pointers, then I recommend using `std::unique_ptr`, which will delete the objects automatically when the pointer goes out of scope (no `delete` call needed). If the elements don't need to be pointers, then you could construct them in-place, although this would require a dynamically allocated container, such as `std::vector`.

There are different acceptable approaches here, but the **required** part of this comment is that we fix the memory leak. Here's just one option that both maintains the pointer-ness of the elements and the static-ness of their container. 
```suggestion
  auto handlers = std::array<std::unique_ptr<PredictionHandler>, NUM_NODES>();
  for (int i = 0; i < NUM_NODES; i++)
  {
    handlers[i] = std::make_unique<PredictionHandler>(
      EoD_events[i].remaining_useful_life,
      EoD_events[i].state_of_charge,
      EoD_events[i].battery_temperature,
      m_nodes[i].bus, m_nodes[i].name, i
    );
  }
```
`std::array` is just a modern C++ wrapper for old C++ arrays that enables the same methods other modern C++ containers have. Using either is fine here. They are essentially the same.

AstroStucky (changes requested):
The changes look good for the most part. Thank you! The memory leak should addressed, then this should be good.

AstroStucky (approved):
This is good work, well done! 

I was able to test, and everything seemed to work fine. I'll leave a closer inspection of the actual predicted values to Chetan. In addition to testing as-is, I compiled with `NUM_NODES = 24` and it looked to run alright; however, it did max out every single one of my 16 threads. 

Even before this work it bothered me that `m_profile_increment` was essentially a constant, even though I would think that slower systems may require a larger increment and GSAP may not consistently run at 0.5 Hz each and every frame. For that matter, the correct increment would not even necessarily be a whole number, which would require us to interpolate the energy profiles. I think allowing users to set the number of nodes only exacerbate this concern. However, I may be misunderstanding how the power system works in concert with GSAP.

Lumsterling:
I'll wait to merge this into OW-994 until @chetankul gives the thumbs-up regarding the output values, but I'll probably merge this branch with the other branches (OW-1033 & OW-1035) momentarily so I can continue work there.

> Even before this work it bothered me that `m_profile_increment` was essentially a constant, even though I would think that slower systems may require a larger increment and GSAP may not consistently run at 0.5 Hz each and every frame. For that matter, the correct increment would not even necessarily be a whole number, which would require us to interpolate the custom fault profiles. 

I suspect this longer delay associated with more nodes could technically be worked around by setting `m_profile_increment` to a value consistent with the delay (e.g. 4 or 5), but this would mean that fault profiles would only take every 4th or 5th indexed value whereas they take every other indexed value at the moment. It still wouldn't fix the issue with the correct increment being a decimal and not a whole number though.

I don't know enough about the accuracy of the model to make a definitive judgment call, but I feel like doing something like this would make the whole simulation less reliable. It's something I'll need to bring up with Chetan & Mike next week.

> I think allowing users to set the number of nodes only exacerbates this concern. However, I may be misunderstanding how the power system works in concert with GSAP.

The user isn't necessarily supposed to change the number of nodes themselves; the `NUM_NODES` constant simply makes it easy for testing to modify the number of parallel battery nodes. To my understanding the battery model should always have a specific number of parallel series' cells (e.g. 24) in the final production.

ow_power_system/include/PowerSystemPack.h
```diff
@@ -0,0 +1,134 @@
...
+  ros::Publisher m_remaining_useful_life_pub;  // Remaining Useful Life Publisher
+  ros::Publisher m_battery_temperature_pub;    // Battery Temperature Publisher
+  ros::Subscriber m_joint_states_sub;          // Mechanical Power Subscriber
+
+  // Create a PowerNode struct, containing all the necessary information for a
+  // node to operate, for each node.
+  PowerNode m_nodes[NUM_NODES];
+
+  // The matrix used to store EoD events.
+  EoDValues EoD_events[NUM_NODES];
```
kmdalal, line 81: Should have `m_` prefix.

ow_power_system/include/PredictionHandler.h
```diff
@@ -0,0 +1,38 @@
...
+// The PredictionHandler class subscribes to the battery EoD event message and
+// prints each event as it is received.
+class PredictionHandler : public IMessageProcessor
+{
+public:
+  PredictionHandler(double& rul, double& soc, double& temp, MessageBus& bus,
+                    const std::string& src, int node_num);
+  ~PredictionHandler();
+  void processMessage(const std::shared_ptr<Message>& message) override;
+private:
```
kmdalal, line 25: All data members should be prefixed `m_`.

ow_power_system/src/PredictionHandler.cpp
```diff
@@ -0,0 +1,137 @@
...
+static constexpr int MODEL_TEMPERATURE_INDEX = 1;
+
+/**
+ * Constructs a new prediction handler that subscribes to battery EoD
+ * predictions for the specified source and on the specified message
+ * bus.
+ **/
+PredictionHandler::PredictionHandler(double& rul, double& soc, double& temp, MessageBus& bus,
+                    const std::string& src, int node_num) : rul_ref(rul), soc_ref(soc),
+                    temp_ref(temp), bus(bus)
```
kmdalal, line 28: Could be better indented for readability.

ow_power_system/src/PredictionHandler.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+  auto model_output = model.outputEqn(now_s.count(), static_cast<PrognosticsModel::state_type>(temperature_state));
+
+  // Store the newly obtained data.
+  rul_ref = rul_median;
+  soc_ref = soc_median;
+  temp_ref = model_output[MODEL_TEMPERATURE_INDEX];
+}
+
+double PredictionHandler::findMedian(std::vector<double> samples)
```
kmdalal, line 114: `const std::vector<double>&` would be better.

ow_power_system/src/PredictionHandler.cpp
```diff
@@ -0,0 +1,137 @@
...
+{
+  std::nth_element(samples.begin(), (samples.begin() + (samples.size() / 2)),
+                    samples.end());
```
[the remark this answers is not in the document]
Lumsterling: Is there a specific guideline I should be following when it comes to indentation here? I could give each argument its own line, for example (I think I remember seeing that in the code before), but I don't recall if that's best practice or not.

I was originally just following the line character limit guidelines.

ow_power_system/src/PredictionHandler.cpp
Lumsterling: `nth_element` edits the `samples` vector used within, so to my understanding the `const` flag shouldn't be used here. Trying to do so causes a lot of errors.

Given the function edits the passed-in vector by sorting it, but we don't want the edits to be permanent or relate to the original vector, shouldn't I avoid passing in by reference?

kmdalal: You're right, never mind my comment.  I'm not that familiar with nth_element and didn't realize it is destructive.

ow_power_system/src/PredictionHandler.cpp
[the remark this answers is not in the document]
kmdalal: It was a minor comment -- the second line is a space off.  (The same in my other comment).  If that's because of tab characters, those should be replaced with spaces.  Line length is the only formal guideline, though extra lines (e.g. one per argument) can sometimes make the code more readable.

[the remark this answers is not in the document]
Lumsterling: Ah, I didn't catch that. I'll go ahead and fix that, though maybe my editor is messing up the indentation because I feel I would have caught that myself while writing the code.

kmdalal:
I haven't finished the code review, but from a test of ReferenceMission1, my overall opinion is that even with 8 nodes, the simulation slowdown is unacceptable.   Have you thought of ways to avoid the slowdown, or is it unavoidable with this general approach?

I also note that each node has practically identical values.  I guess this is expected, but I'm wondering how and if GSAP can be used to simulate differences in the nodes.

kmdalal:
Some of the slowdown might be due to all the terminal output.  Could you please remove all the debugging output and see if this makes a difference?

Lumsterling:
> I haven't finished the code review, but from a test of ReferenceMission1, my overall opinion is that even with 8 nodes, the simulation slowdown is unacceptable. Have you thought of ways to avoid the slowdown, or is it unavoidable with this general approach?

I'll do some experimentation regarding waiting for the buses. I'm wondering if it's possible to use only one MessageBus instead of one per node. If not, I'm not very optimistic about getting the output much faster.

> I also note that each node has practically identical values. I guess this is expected, but I'm wondering how and if GSAP can be used to simulate differences in the nodes.

It concerns me too, though GSAP is more or less a black box for me. I'm just sending in sets of data and receiving sets in return. I'll do some testing where I force different power usages on the different nodes and see how that affects the output.

> Some of the slowdown might be due to all the terminal output. Could you please remove all the debugging output and see if this makes a difference?

I tested this before and didn't notice a significant difference, but I'll give it another try.

Lumsterling:
> each node has practically identical values

I just finished a test where I forced each node to have a different power value and compared the results to the standard (where all nodes share the same power value).

The temperature input value for each node was identical, and the voltage values had the same extremely minor variance. However, the output battery temperatures from GSAP were different (they increased the higher the power, as expected). When the input power and temperature is identical from each node, the output battery temperature is identical from each call to GSAP.

I took a look at the function used to determine the input temperature values for each node and found it's just a simple random number generator. I assume this is why the values are identical between each node (i.e. the same seed is being used). The voltage estimate also uses a random number generator, but has other calculations that I assume account for the slight variance between each node. These functions were present before I started work on the EPIC, so I didn't pay them much mind, but it seems that they might be due for a rework? We'd need to consult with Chetan, most likely.


> slowdown

Preliminary test suggests that using only one MessageBus increases speed substantially, but I don't think I have it working properly yet so I'd take this with a grain of salt. The primary issue is that the callback function receiving the output data doesn't know which node it belongs to (and the output data doesn't specify). I'll spend some time on this today and see if I can figure out a method to get this working.

@kmdalal @chetankul Do we need to organize the battery nodes such that we can look at the data and say, for example, "Node 3 seems to have an unreasonable temperature increase", or is it good enough to be able to say "one of the nodes seems to have an unreasonable temperature increase"? Essentially, I already have an idea to store all the data from one bus, but it won't be ordered (since the order in which GSAP's bus returns data is unknown) nor will it be possible to know exactly which node it belongs to.

Lumsterling:
After reworking the code to send and receive everything from a single bus, I found it unfortunately runs into the same problem the original simple prognoser had: you only get back 1 response per cycle regardless of how many data sets you send in. Based on this, it seems like declaring multiple MessageBuses is the only way to get multiple predictions at once, and unfortunately it is significantly slow. I've discarded the changes as such since it doesn't improve the code.

I'll keep experimenting, but as of currently it seems like the only way to reduce slowdown from this end is to reduce the number of nodes simulated. I'm unsure how the modeling side of this might work for that (or if there's a way to reduce the simulated nodes without sacrificing accuracy in the first place).

kmdalal:
Thanks for exploring this thoroughly, @Lumsterling.  It looks like a significant rework may be needed for a workable solution.  I don't know what to suggest, but here are some naive ideas:
 - thread GSAP and run multiple concurrent instances, distributing the nodes (I presume nodes = cells)
 - settle for less than 8 nodes, within acceptable performance, but have each "node" somehow represent more than one cell
 - simulate the battery pack entirely outside GSAP, doing prognosis only on the aggregate combination.
Let's have a discussion with @chetankul after our Thursday tagup, or as soon as possible.  One basic thing I'd like to revisit is the relative importance of simulating individual cells of the battery.

ow_power_system/include/PredictionHandler.h
```diff
@@ -0,0 +1,38 @@
...
+#include "Messages/MessageBus.h"
+#include "Messages/ProgEventMessage.h"
+#include "Messages/ScalarMessage.h"
+#include "ModelBasedAsyncPrognoserBuilder.h"
+
+// The PredictionHandler class subscribes to the battery EoD event message and
+// prints each event as it is received.
+class PredictionHandler : public IMessageProcessor
+{
+public:
```
kmdalal, line 20: Good to be explicit about the copy constructor and assignment operator, .e.g. define them or use `= delete` or `= default`.

ow_power_system/include/PowerSystemPack.h
```diff
@@ -0,0 +1,134 @@
...
+// Struct that groups the end-of-discharge (EoD) prediction values.
+struct EoDValues {
+  double remaining_useful_life;
+  double state_of_charge;
+  double battery_temperature;
+};
+
+class PowerSystemPack
+{
+public:
```
kmdalal, line 58: Good to be explicit about the copy constructor and assignment operator, .e.g. define them or use `= delete` or `= default`.  Same for the destructor, though either define or default it, since it's required.

ow_power_system/src/PowerSystemPack.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+#include "PowerSystemPack.h"
+#include "PredictionHandler.h"
+
+using namespace PCOE;
+
+const auto START_TIME       = MessageClock::now();
+
+PowerSystemPack::PowerSystemPack()
+{ }
```
kmdalal, line 25: Could just `= default` the declaration.

ow_power_system/include/PredictionHandler.h
```diff
@@ -0,0 +1,38 @@
...
+  ~PredictionHandler();
+  void processMessage(const std::shared_ptr<Message>& message) override;
+private:
+  double findMedian(std::vector<double> samples);
+
+  // References used by other classes to process EoD predictions.
+  double& m_rul_ref;
+  double& m_soc_ref;
+  double& m_temp_ref;
+  MessageBus& m_bus;
```
kmdalal, line 32: This idiom (aggregation instead of composition, or having a copy of the objects instead of references) can be risky.  I need to look at the overall usage, but don't have time at the moment.

Lumsterling: I more or less had to implement it this way, because PredictionHandler needs to get its results into the variables of the PowerSystemPack object. PowerSystemPack is the object that handles and stores these variables, which also controls PredictionHandlers, so I think it's alright. I could be misusing it though still.

PR #263, opened 2022-08-25
Oceanwater 1008 add joint limit to all arm joints

anjan1984:
## Linked Issues:
| EPIC ⚡ | [OCEANWATER-XXX](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-XXX) |
| Jira Ticket 🎟️ | [OCEANWATER-XXX](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1008) |
| Github :octocat:  | # |

## Summary of Changes
* 
Added joint limits for all the joints 
## Test
* Run reference mission 1 to see there is no issue with joint limits 
* Can move all the joints to tests limits using 
* python3 arm_move_joints_client.py false 0.1 0.1 0.1 0.1 0.1 0.1


mogumbo (changes requested):

ow_lander/urdf/lander.xacro
```diff
@@ -920,7 +920,7 @@
     <axis
       xyz="0 0 -1" />
     <dynamics damping="50.0" />
-    <limit effort="100.0" velocity="0.2" />
+    <limit effort="100.0" velocity="0.2" lower="-3.14" upper="3.14"/>
```
mogumbo, line 923: Can the tilt limits be tightened up to prevent collisions with the rover. We might not be able to prevent all collisions this way, but something like -pi/2 to pi/2 would probably be better.

ow_lander/urdf/lander.xacro
```diff
@@ -844,7 +844,7 @@
     <axis
       xyz="0 0 -1" />
     <dynamics damping="50.0" />
-    <limit effort="80.0" velocity="0.2" />
+    <limit effort="80.0" velocity="0.2" lower="-3.14" upper="3.14" />
```
mogumbo, line 847: We discussed pan a long time ago and settled on something greater than pi (I think it was 3.2), so if there's any kind of oscillation in the controller the joint won't need to wrap all the way around.

3.14 is actually a little less than pi and would leave a small gap through which you cannot pan. The same consideration should be given to other joints where these limits are used.

mogumbo (changes requested):

AstroStucky (changes requested):
This branch is missing the ArmMoveJoints.action file, and it prevents me from compiling it from scratch. I think `noetic-devel` needs to be merged into it. Please do this and let me know.

AstroStucky:
I merged it locally on my branch, so I can test, but stopped short of pushing to origin. 

Can you explain what scenarios cause this error to print? 
```
[INFO] [1661540608.704885, 1916524835.695000]: Arm move joint started
[ERROR] [1661540608.971473868, 1916524835.962500095]: arm/arm: Unable to sample any valid states for goal tree
[ WARN] [1661540608.971794097, 1916524835.962500095]: Fail: ABORTED: No motion plan found. No execution attempted.
```
Is it a joint angle being out-of-range, or the planner not finding a solution that does not self-collide? It seems like this error will sometimes print for some manual angle movements dependent on the current configuration of the arm, which makes me think it's the latter.

ow_lander/urdf/lander.xacro
kmdalal: The tilt range we decided on a while back is in fact [-pi/2, pi/2] .

ow_lander/urdf/lander.xacro
kmdalal: The pan range we decided on a while back is in fact [-3.2, 3.2] .

ow_lander/urdf/lander.xacro
anjan1984: Done!

ow_lander/urdf/lander.xacro
anjan1984: Done!

anjan1984:
> I merged it locally on my branch, so I can test, but stopped short of pushing to origin.
> 
> Can you explain what scenarios cause this error to print?
> 
> ```
> [INFO] [1661540608.704885, 1916524835.695000]: Arm move joint started
> [ERROR] [1661540608.971473868, 1916524835.962500095]: arm/arm: Unable to sample any valid states for goal tree
> [ WARN] [1661540608.971794097, 1916524835.962500095]: Fail: ABORTED: No motion plan found. No execution attempted.
> ```
> 
> Is it a joint angle being out-of-range, or the planner not finding a solution that does not self-collide? It seems like this error will sometimes print for some manual angle movements dependent on the current configuration of the arm, which makes me think it's the latter.

I merged the noetic devel again. As you mentioned, motion plan can fail for both the reasons.

ow_lander/urdf/lander.xacro
mogumbo: This should also be done to other joints that currently use [-3.14, 3.14]. Although, j_prox_pitch could probably be tightened up to use a smaller range.

kmdalal (approved):
My review was light.  The tests pass and the pan/tilt ranges look correct.  I believe we need a more thorough test for all the other joints.

AstroStucky:
> It seems the antenna tilt will try to move through its deadzone. If you have it at -1.57, then command 1.57, it'll move down toward the lander.
For that matter it doesn't obey its limits at all. I command a tilt of 1.8, and it face planted against the lander deck.

AstroStucky:
What is intended behavior for the antenna pan? Should it be able to full rotate around 360 degrees? It appears to be able to, but then why do we bother giving it limits?

mogumbo:
> What is intended behavior for the antenna pan? Should it be able to full rotate around 360 degrees? It appears to be able to, but then why do we bother giving it limits?

We're trying to simulate the hardware that engineers would probably choose to send on such a mission. Building a (physical, not simulated) pivot joint is more complex if it has no limits, so it probably wouldn't fly on a real mission.

kmdalal:
Enforcing the pan/tilt limits in the action server is the work of OW-890, which I started but did not finish for release 10.  I don't actually know how the ranges in the .xacro file are used.  Should they be automatically enforced?

AstroStucky:
Sounds like another ticket exists to implement enforcement of antenna joint limits, so I'll leave the efficacy of their new limits out of my review.

ow_lander/urdf/lander.xacro
anjan1984: Do we want the range to be -3.2 to 3.2 for all the joints that are free to rotate ?

mogumbo: Seems like a reasonable guess. It's impossible to know how much wrap-around would be designed into a joint for a real mission.

AstroStucky (commented):
Results of my thorough limit testing so far. All tests used **arm_move_joint_client.py**
| Joint | Results |
| ---     | ---          |
| **j_shou_yaw** | Limits work ✅ |
| **j_shou_pitch** | Limits work ✅ |
| **j_prox_pitch** | Skipped. Awaiting requested change in limits |
| **j_dist_pitch**  | Lower limit did not work. I was able to command the joint as far as -2.2 before further commanding would result in an abort <br> Upper limit did not work. I could command the joint as far as 3.65 before seeing an abort. |
| **j_hand_yaw** | Skipped. Awaiting requested change in limits |
| **j_scoop_yaw** | Skipped. Awaiting requested change in limits |

anjan1984:
> 
great feedback! Should I change the limits for j_dist_pitch -2.2 and 3.65. I could only go till -1.9 to 3.5 but it depends on the other parts of the arm as well.

ow_lander/urdf/lander.xacro
anjan1984: ok changed the defaults to 3.2

AstroStucky:
> > 
> 
> great feedback! I changed the limits for j_dist_pitch -2.2 and 3.65. I could only go till -1.9 to 3.5 but it depends on the other parts of the arm as well. thanks

Are the limits enforced by anything more than the planner checking for self-collision?

anjan1984:
> > > 
> > 
> > 
> > great feedback! I changed the limits for j_dist_pitch -2.2 and 3.65. I could only go till -1.9 to 3.5 but it depends on the other parts of the arm as well. thanks
> 
> Are the limits enforced by anything more than the planner checking for self-collision?

No. the idea was you should not be able to command a joint outside the range

mogumbo:
I think it is really weird that the joint limits aren't enforced by the physics engine; there *must* be a simpler way to enforce them built into ROS. Though I am not sure how much it matters to us as long as they are strictly enforced. In an actual lander, hardware limits would surely be programmed into the software and the software would not push servos to try to exceed the limits. Does anyone have any thoughts on this?

I also think it's weird that we have joint position limits in the lander.xacro file and joint velocity and acceleration limits in joint_limits.yaml. If there is a way to put all limits in one place, that would be preferable. This could be done as a separate ticket, though.

mogumbo:
I tried starting the sim and then doing:
```
./arm_move_joint_client.py False 0 0
./arm_move_joint_client.py False 2 -3
```
The second movement fails telling you `Lower and upper limit of the joint is -3.2 3.2`, but I was asking for a position within those limits. Perhaps the system detects that there will be a collision and does not move, but the error message is misleading?

AstroStucky:
> The second movement fails telling you Lower and upper limit of the joint is -3.2 3.2, but I was asking for a position within those limits.

What causes this output? I've only ever seen the following error result in the simulation log from calling the **arm_move_joint_client.py** script. 
```
[ERROR] [1662076127.023447766, 1916525080.092499971]: arm/arm: Unable to sample any valid states for goal tree
[ WARN] [1662076127.023710580, 1916525080.092499971]: Fail: ABORTED: No motion plan found. No execution attempted.
```

AstroStucky (approved):
ReferenceMission1.plx works, and the limits look reasonable.

PR #264, opened 2022-08-25
return scoop tip position

snoyes:
The interface was looking for the position of the scoop rather than the scoop tip.


mogumbo (approved):
Good catch. This looks correct to me. I tried a variety of tests but don't see any change in behavior.

anjan1984 (approved):

kmdalal:
In Dev Day this morning we decided to merge this for Release 10.  If it introduces unexpected issues in release testing that we don't want to fix for Release 10, we'll back it out.

kmdalal:
Thank you, @snoyes for the contribution!  Hope OceanWATERS is working out well for you.

PR #269, opened 2022-08-31
Updated battery values

kmdalal:
## Summary of Changes
* Changes max temperature threshold of battery from 50 C to 70 C, per advice from @chetankul 
* Changes base voltage from 3.2 to 2.8, per advice from @chetankul 
* 
## Test
* catkin build
* start any simulator world
* start ow_plexil
* run the plan Continuous.plx
* In a new terminal, `rostopic echo /power_system_node/battery_temperature`
* In a new terminal, `rostopic echo /faults/power_faults_status`
* Inject a high power draw of ~30 watts
* Watch the battery temperature and power fault status.  The former should start around 20 and the latter should be 0.  The temperature should steadily increase, and within a few minutes reach 70.  When it does, the power fault status should be 1.   (Before this fix, 50 was the threshold temperature).

NOTE: there is no test for the base voltage change.

### DIFF
--- a/ow_faults_detection/include/ow_faults_detection/FaultDetector.h
+++ b/ow_faults_detection/include/ow_faults_detection/FaultDetector.h
@@ -66,7 +66,7 @@ public:
   static constexpr std::bitset<3> isCapLossError{    0b010 };
   static constexpr std::bitset<3> isThermalError{    0b100 };
 
-  static constexpr float THERMAL_MAX = 50;
+  static constexpr float THERMAL_MAX = 70;
   static constexpr float SOC_MIN = 0.1;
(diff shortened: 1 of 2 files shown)


Lumsterling (approved):
Looks like it works as intended. Given it's a one-line change, I don't really have any other feedback, but this worked well as a basic runthrough of how reviews work. It'll likely help in the upcoming month. Thanks!

kmdalal:
@Lumsterling I just updated the base voltage based on @chetankul's email.  I don't think the voltage will affect this test, and I don't know how to specifically test the impact of the new base voltage -- will ask @chetankul and then ask you to re-test.

PR #270, opened 2022-09-01
Oceanwater 1033 & 1035: value publishing & multi-cell fault injection

Lumsterling:
## Linked Issues:
| Jira Ticket 🎟️ | [OCEANWATER-1033](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1033) & [OCEANWATER-1035](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1035) |
| Github :octocat:  | Lumsterling |

NOTE: This PR merges the changes from both OW-1033 and OW-1035. OW-1033 is relatively small in scope and should be simple to test, so a full PR for it is less necessary.

## Summary of Changes
* OW-1033
  * The switch to multiple cells in the power model required a rework of value publishing. The power system should now publish end-of-discharge (EoD) predictions identically to how it was done before the change.
  * Remaining useful life (RUL), state of charge (SoC) and battery temperature are published by the PowerSystemPack object now, rather than the multiple PowerSystemNode objects.
* OW-1035
  * The switch to multiple cells in the power model also required a rework of fault injection. When a high power draw is applied, the power system now spreads the additional consumption evenly across all nodes rather than the full strength on each node. This required removing the fault injection mechanism from PowerSystemNode and adding it to PowerSystemPack instead.
  * As part of this change, it was brought to my attention that the voltage and temperature columns in the fault profiles are non-functional. They have been removed to reduce user confusion. The only method of fault injection is through additional power draw.
* While this JIRA should overall work, there are many undesired components that need to be addressed in future JIRAs:
  * The amount of real time it takes to get multiple EoD predictions back is unreasonably long.
  * Many sections of code are commented out but not fully deleted yet.

## Test
* Open up 4 terminals 
* In the 1st terminal, run the following in the oceanwaters directory
  * `catkin build`
  * `roslaunch ow_europa_terminator_workspace.launch`
* In the other three terminals, run one of the following commands
    * `rostopic echo /power_system_node/remaining_useful_life`
    * `rostopic echo /power_system_node/state_of_charge`
    * `rostopic echo /power_system_node/battery_temperature`
* In the primary, a large number of debug statements should continually display:
  * These statements show the status of each of the 8 nodes, as well as the EoD predictions resulting from passing those statuses into GSAP.
  * Of particular note is the line detailing `min_rul`, `min_soc`, and `max_tmp`. These should correspond to the minimum or maximum of all node EoD predictions for that cycle.
* Check the other 3 terminals and ensure their reported values line up with `min_rul`, `min_soc`, and `max_tmp`. The timing of these published values should line up with the main terminal's outputs.
* In the 'Dynamic Reconfigure' tab of the RQT window, navigate to the `power_faults` subsection of `faults` and activate `high_power_draw`.
* Monitor the debug statement outputs and confirm the additional power draw injected is distributed evenly over all nodes and adds up to the value set in RQT. This should also update for the next cycle if the value is modified.
* Create a new fault profile within the `profiles` directory of `ow_power_system`. Reference `example_fault.csv` to understand how the file should be formatted. Note the new 2-column format.
  * Inject the modified fault profile using the `custom_fault_profile` section in RQT.
  * Monitor the debug statement outputs and confirm the additional power draw injected is distributed evenly over all nodes and adds up to the value set in the current index of the fault profile.

## Potential Issue(s)
* This was discovered inadvertently while writing this PR: if the simulation runs long enough such that the SoC of the battery approaches 0, the behavior starts falling apart. The minimum SoC appeared to sit at 0.0230556 for many cycles, but the RUL remained over 2000. Eventually the process crashed due to a `Matrix is not positive definite` error. I will test this on the `noetic-devel` branch to see if similar behavior occurs at very low SoC, or if a modification I made seemingly caused this, and update this PR accordingly.
  * Testing on `noetic-devel` revealed that the EoD predictions start outputting 0.0 values as the battery status degrades. The times at which this occurs initially seemed to be tied to battery temperature (there was a sharp increase in temperature to ~96C before the predictions flatlined), but I attempted to bring SoC to 0 without triggering any other faults by deactivating `high_power_draw` and allowing the temperature to decrease before reactivating it. In this case, I was able to bring SoC to around 30% before the predictions flatlined (the sharp increase in battery temperature still occurred before the flatline, but this time it was only at ~60C).
  * My initial guess is that since the output temperature is defined as the maximum temp of all nodes, but faults now spread their draw across all nodes evenly (causing the individual temperature to rise slowly and not trigger a fault), that SoC is reaching 0 before a fault can be detected via temperature changes. I was unable to get SoC to 0 in `noetic-devel` before a fault was triggered. This is a problem with the fault injection & detection given the change to the battery model.
  * Chetan stated the published battery temperature should be the maximum of any one node temperature, but if power draw is split across all nodes, this leads to much lower temperature increases than previously. This may need to be discussed.
* Note that this erratic behavior could be due to the single-cell model still being used in GSAP. The series' cell model has not been implemented yet. Will do further testing to try and narrow down the source of the error.


Lumsterling:
Further testing is revealing some serious issues when injecting power draw. The last run I had, I injected 230W of power (split into 29.75W on every node) and after about 20 seconds, the output started flying forward despite GSAP not returning anything yet. Reached timestamp 334 in the span of about 5 seconds, then crashed.

Given this, I'm converting this PR into a draft because it isn't ready for a full merge yet (even just into OW-994).

Lumsterling:
I'm closing this PR, as a newer version has been created with more organized commits and updated changes: https://github.com/nasa/ow_simulator/pull/273

PR #273, opened 2022-09-20
Oceanwater 1035/1033/1040: published values, fault injection, & code cleanup

Lumsterling:
## Linked Issues:
| Jira Ticket 🎟️ | [OCEANWATER-1033](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1033), [OCEANWATER-1035](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1035), & [OCEANWATER-1040](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1040) |
| Github :octocat:  | # |

NOTE: This is a rework of the original pull request, which contained (as a result of several branch merges) many commits in an unorganized manner. Hopefully this PR does not have that same issue. Additional fixes have also been made such that this branch is hopefully better suited to a PR than the previous one, though a full merge into _noetic-devel_ is not suitable until future changes (both potentially inside and outside OceanWATERs) have been completed.

This PR merges the changes from OW-1033, OW-1035, and OW-1040. OW-1033 and OW-1040 are relatively small in scope and should be simple to test, so a full PR for them is less necessary.

## Summary of Changes (updates)
* GSAP uses the very first set of data sent in as its initial starting values for predictions, and was not set properly in the initial PR. This has since been addressed.
* There is a new constant in PowerSystemPack.h, ``NUM_SAMPLES``. This affects the number of samples generated by GSAP when predicting, and has a large effect on performance. The default is ``100``, but personal testing found similar performance to the original simple prognoser with a value of ``25``. This value may differ from computer to computer. Further testing is needed to determine changes in prediction accuracy as ``NUM_SAMPLES`` changes.
* The asynchronous prognoser now uses a moving average for future load estimations rather than a constant value of ``8``. This led to inaccurate predictions, and was present even in the original simple prognoser, implying its predictions were off as well. The current moving average window is set to ``1``, but this can easily be modified in ``async_prognoser.cfg``. It will likely need to be increased in the future.
* The behavior when the 'RUL goes to 0' bug occurs still results in wild predictions and eventually a crash, and until that is addressed it is unlikely the EPIC ticket can be fully resolved. Testing in recent weeks seems to suggest the root of the problem is outside these changes, however. It seems that the majority of issues surrounding fault injection stem from this one problem, though I could be wrong in that assessment.
* All 3 primary files and 3 header files in _ow_power_system_ have been cleaned up. Unnecessary comments and unused variables have been removed, while additional comments have been added for clearer documentation.

## Summary of Changes (original)
* OW-1033
  * The switch to multiple cells in the power model required a rework of value publishing. The power system should not publish end-of-discharge (EoD) predictions identically to how it was done before the change.
  * Remaining useful life (RUL), state of charge (SoC), and battery temperature are published by the PowerSystemPack object now, rather than multiple PowerSystemNode objects.
* OW-1035
  * The switch to multiple cells in the power model also required a rework of fault injection. When a high power draw is applied, the power system now spreads the additional consumption evenly across all nodes rather than the full strength on each node. This required removing the fault injection mechanism from PowerSystemNode and adding it to PowerSystemPack instead.
  * As part of this change, it was brought to my attention that the voltage and temperature columns in the fault profiles are non-functional. They have been removed to reduce user confusion. The only method of fault injection is through additional power draw.
* While this JIRA should overall work, there are many undesired components that need to be addressed in future JIRAs:
  * ~~The amount of real time it takes to get multiple EoD predictions back is unreasonably long.~~
    * This can be modified by adjusting the prognoser's sample count and horizon, though the overall accuracy of the simulation needs to be tested with this change.
  * ~~Many sections of code are commented out but not fully deleted yet.~~
    * This has been addressed now that OW-1040 has been integrated into this PR.

## Test
* OW-1033 (published values)
  * Open up 4 terminals
  * In the 1st terminal, run the following in the oceanwaters directory:
    * ``catkin build``
    * ``roslaunch ow_europa_terminator_workspace.launch``
  * In the other three terminals, run one of the following commands in each:
    * ``rostopic echo /power_system_node/remaining_useful_life``
    * ``rostopic echo /power_system_node/state_of_charge``
    * ``rostopic echo /power_system_node/battery_temperature``
  * In the primary terminal, a large number of debug statements should continually display:
    * These statements show the status of each of the nodes, as well as the EoD predictions resulting from passing those statuses into GSAP.
    * Of particular note is the line detailing ``min_rul``, ``min_soc``, and ``max_tmp``. These should correspond to the minimum or maximum of all node EoD predictions for that cycle.
  * Check the other 3 terminals and ensure their reported values line up with ``min_rul``, ``min_soc``, and ``max_tmp``. The timing of these published values should line up with the main terminal's outputs.
* OW-1035 (fault injection)
  * In the 'Dynamic Reconfigure` tab of the RQT window, navigate to the ``power_faults`` subsection of ``faults`` and activate ``high_power_draw``. Adjust the value of high power draw as needed (note the new maximum of 720W is to allow for testing of up to 30W injection to 24 nodes at once).
  * Monitor the debug statement outputs and confirm the additional power draw injected is distributed evenly across all nodes and adds up to the value set in RQT. This should also update for the next cycle if the value was modified.
    * **NOTE:** It is likely that the simulation will start outputting negative RUL values (and eventually crash) if additional power draw is left on for long enough. This is a longstanding issue and is not addressed in this PR. More info below.
  * Create a new fault profile within the ``profiles`` directory of ``ow_power_system``. Reference ``example_fault.csv`` to understand how the file should be formatted. Note the new 2-column format.
    * Inject the modified fault profile using the ``custom_fault_profile`` section in RQT.
    * Monitor the debug statement outputs and confirm the additional power draw injected is evenly distributed across all nodes and adds up to the value set in the current index of the fault profile.

## Potential Issue(s)
* When a power draw is applied to the simulation for long enough, eventually the predictions from GSAP appear to go off course. The RUL dips into negative values and eventually the process crashes. The timing of when this occurs varies, but appears linked to the severity of the power draw; the higher the draw, the faster the bug occurs. It appears to be the same issue that causes RUL to suddenly drop to 0 in ``noetic-devel``, as it appears under extremely similar behavior with similar results (though the process crashes rather than all future predictions flatlining at ``0.0``). This bug needs to be addressed before a full merge into ``noetic-devel``, but it is unclear if the root of the problem is within OceanWATERs or GSAP at this time.
* While this PR renders OW-994 technically feature-complete, there are still significant roadblocks preventing it from being merged into ``noetic-devel``.
  * The RUL bug described above needs to be addressed.
    * ***This requires an update to GSAP's ``develop`` branch using the tag ``v1.0.2b-OW``.***
  * The single-cell model within GSAP needs to be updated to a series cell model to output accurate predictions with the new battery pack implementation.
  * Performance is still an issue. More testing with the value of ``NUM_SAMPLES`` needs to be done to determine if the accuracy loss from significantly reducing the number of samples per prediction is viable.


kmdalal:
Results of testing:

- I ran the simulator for about 30 minutes under varying power loads.  Behavior was as expected and consistent with the test description.
- After experimenting with different high power loads, I let the sim run its course under a normal power load (no fault injection, Continuous.plx plan running).  The RUL gradually reduced to 0 (non-monotically) and then started going more and more negative.  It hit 0 when SOC was 0.22 : this is clearly a bug in itself, aside from the erratic/negative values.
- Performance was acceptable in the current configuration.  I have not yet tried increasing NUM_SAMPLES.

kmdalal:
The types PrognoserVector and PrognoserMap are duplicated in PowerSystemNode.h and PowerSystemPack.h.  There is also a duplicated member of the former type, m_custom_power_fault_sequence.  This needs some refactoring.  One solution may be to create a new header file of definitions used by both classes.

kmdalal:
The many lines commented "NOTED FOR DELETION" would better just be removed, unless you think they may need to be restored soon.

ow_power_system/include/PowerSystemPack.h
```diff
@@ -29,6 +29,22 @@ using PrognoserVector = std::vector<PrognoserMap>;
...
+// Monte Carlo prediction process. Lower values mean faster performance, but lower
+// accuracy (needs testing for confirmation).
+// The default value is 100, but this is very slow and infeasible for the simulation.
+// My computer reached similar speeds to the original simple prognoser at a sample
+// value of 25, but this may vary from computer to computer. ~Liam
+const int NUM_SAMPLES = 100;
+
+const std::string FAULT_NAME_HPD           = "high_power_draw";
+const std::string FAULT_NAME_HPD_ACTIVATE  = "activate_high_power_draw";
+const int CUSTOM_FILE_EXPECTED_COLS           = 2;
```
kmdalal, line 42: I think we agreed verbally that this should go back to 4, unless @chetankul also feels that voltage and temperature should be excluded from fault injection.  (Values placed here never seemed to have any effect, and were never explained).

kmdalal:
I'm confused by how fault injection is both split and duplicated (in various ways) between PowerSystemNode and PowerSystemPack.  It seems some refactoring is needed.

kmdalal:
PowerSystemNode.h and PowerSystemPack.h would each benefit from a short comment at the top explaining what the class is for.  Is the latter intended to replace the former?

ow_power_system/include/PowerSystemPack.h
Lumsterling: That's one of the things I have to do (and I'll request a re-review once I'm done with them):
- I need to revert the custom CSV formats to include voltage and temperature again. Still not totally sure if they do anything, but it seems like they are supposed to affect the estimation step of the GSAP prediction. They just don't directly apply the values in the file to the current status of the battery.

Lumsterling:
> The many lines commented "NOTED FOR DELETION" would better just be removed, unless you think they may need to be restored soon, or if it's a helpful note to yourself. (But they should be deleted in the final version).

That's the plan for the final Jira (OW-1040) following this PR; I'll be going through all the code and removing parts that are no longer necessary. I just haven't done it yet in case.

Lumsterling:
@kmdalal Given the remaining time in my internship, let me know if I should just start implementing OW-1040 (code cleanup and comment removal) within this PR rather than making a new one following this. Maybe that's the better option now?

The hope is to leave the OW-994 EPIC branch in a state that is as close to ready-to-merge as possible, along with a list of any remaining potential issues that would still be preventing the merge.

kmdalal:
@Lumsterling yes doing the OW-1040 cleanup work in this PR is a good idea, for the reason you mentioned.

kmdalal:
I retested after updating GSAP to the prescribed version and getting the latest commit of this branch.  I do one thing differently: add a constant power load by running the non-terminating PLEXIL plan Continuous.plx, which repeatedly stows/unstows the arm and pans/tilts the antenna.

I kept alternating between no faults, and injections of 40, 60, and 80 watts.  Temperature and SOC behaved intuitively, but not RUL:

- An RUL of infinity (which is 0 as telemetry, a separate issue) doesn't seem to make sense for a baseline power draw of ~3 watts per cell.  
- With injections of 60+ watts, the RUL drops to 0 within several cycles and then goes negative, with fluctuating numbers.  (Known issue).
- When the power fault is removed, the RUL sometimes returns to a value much _higher_ than it was before the fault was injected.  This doesn't seem right.  I didn't record specifics.
- When the power fault is removed, the baseline power usage returns to different values.  E.g. it started around 3 watts, and later returned to 5 watts.  This  problem seems outside fault injection altogether.

I realize that the above problems are out of scope of this PR.  Are there more code changes or GSAP updates coming in the next couple days?  If so I'll re-review, otherwise I can approve this now with the above caveats noted.

Lumsterling:
> * An RUL of infinity (which is 0 as telemetry, a separate issue) doesn't seem to make sense for a baseline power draw of ~3 watts per cell.
If RUL outputs a value of infinity, it's because the prediction went beyond the 'horizon' value, which is 10000 by default. Infinity in this case simply means the RUL is above 10000. This value actually needs to probably be reduced, as the time it takes for a prediction to reach an RUL of 10000+ is infeasible for a given cycle (the higher the RUL, the longer the prediction step takes).
A possible future step is to modify the published RUL such that any 'infinity' value is simply published as the horizon value.

> * With injections of 60+ watts, the RUL drops to 0 within several cycles and then goes negative, with fluctuating numbers.  (Known issue).
According to Chris, this is expected behavior given a single cell can and will degrade extremely rapidly with a power draw of that magnitude. Another future step is to modify the published RUL/SoC values such that any negative value is instead published as 0, and the battery is indicated to be in a fail state. Negative predictions from GSAP simply indicate battery failure, to my understanding.

> * When the power fault is removed, the RUL sometimes returns to a value much _higher_ than it was before the fault was injected.  This doesn't seem right.  I didn't record specifics.
If RUL is returning to a value higher than before the fault was injected, that doesn't seem right, but it makes sense for RUL to return to a high value if the fault is removed. I'll see if I'm getting that behavior on my end.

> * When the power fault is removed, the baseline power usage returns to different values.  E.g. it started around 3 watts, and later returned to 5 watts.  This  problem seems outside fault injection altogether.
I've never seen that behavior on my end. If you've got more specifics on this, let me know.

> I realize that the above problems are out of scope of this PR. Are there more code changes or GSAP updates coming in the next couple days? If so I'll re-review, otherwise I can approve this now with the above caveats noted.
I don't know of any upcoming GSAP updates at the moment, you'd have to reach out to Chris. I mentioned the issue regarding SoC still being out of sync with RUL a few days ago, and he said he'd try to recreate it on his end.

@kmdalal I'm about to push a comment overhaul to the files within ow_power_system; a quick review of that should be sufficient to merge this PR into OW-994, after which I can make additional changes directly to OW-994 before the end of the week.

kmdalal:
I am currently testing the latest commit.  Suggested improvements to power node printout:
- make this toggle-able, e.g. with ROS parameter or config file setting
- put on one line instead of two.  Headings could be shortened, e.g. V for volts.

And a question: what's the meaning of TMP and "temp", assuming at least one them is temperature?

kmdalal:
Thanks for the earlier comments.  A horizon of 10000 seems too small for RUL.  That's under 3 hours, when this lander is to operate for around 20 days!  I'll add this to the list of issues for @chetankul and Chris.

Lumsterling:
> Suggested improvements to power node printout:
> * make this toggle-able, e.g. with ROS parameter or config file setting

I'll see about putting this in system.cfg instead of a const within the code. Shouldn't be too difficult.

> * put on one line instead of two.  Headings could be shortened, e.g. V for volts.

On my end, the printouts are one line each. My console window might be stretched out larger than yours? Regardless, I'll look to trim the printouts a bit; if they're too long for your window then they're probably too long for others as well.

> And a question: what's the meaning of TMP and "temp", assuming at least one them is temperature?

Both actually mean temperature, but while trying to shorten printouts I was using TMP instead of temp since it's one character shorter. I'll update most 'temp'-related code to 'tmp' to maintain consistency, plus 'temp' could mean other things than temperature after all.

kmdalal:
The printouts look separated by hard returns for me.  Widening the window made no difference.
```
[ INFO]: Node 0 RUL: 3144.71. SOC: 0.855539. TMP: 20.8893
[ INFO]: Node 1 RUL: 3142.71. SOC: 0.855539. TMP: 20.8893
[ INFO]: Node 2 RUL: 3152.71. SOC: 0.855539. TMP: 20.8893
[ INFO]: Node 3 RUL: 3146.71. SOC: 0.855539. TMP: 20.8893
[ INFO]: Node 4 RUL: 3139.71. SOC: 0.855539. TMP: 20.8893
[ INFO]: Node 5 RUL: 3144.71. SOC: 0.855539. TMP: 20.8893
[ INFO]: Node 6 RUL: 3165.71. SOC: 0.855539. TMP: 20.8893
[ INFO]: Node 7 RUL: 3161.71. SOC: 0.855539. TMP: 20.8893

[ INFO]: min_rul: 3139, min_soc: 0.855539, max_tmp: 20.889308
[ INFO]: Node 0  time: 388. power: 5.74987.  volts: 3.28359.  temp: 20.5916
[ INFO]: Node 1  time: 388. power: 5.74987.  volts: 3.28362.  temp: 20.5916
[ INFO]: Node 2  time: 388. power: 5.74987.  volts: 3.28362.  temp: 20.5916
[ INFO]: Node 3  time: 388. power: 5.74987.  volts: 3.28362.  temp: 20.5916
[ INFO]: Node 4  time: 388. power: 5.74987.  volts: 3.28362.  temp: 20.5916
[ INFO]: Node 5  time: 388. power: 5.74987.  volts: 3.28362.  temp: 20.5916
[ INFO]: Node 6  time: 388. power: 5.74987.  volts: 3.28362.  temp: 20.5916
[ INFO]: Node 7  time: 388. power: 5.74987.  volts: 3.28362.  temp: 20.5916
```

kmdalal:
Regarding my previous comment, I see that the two groups of printouts are done in different places.  I haven't looked into why, but to rephrase my request, one group of printouts (a line for each node) per cycle would be easier to read.

Lumsterling:
> The printouts look separated by hard returns for me. Widening the window made no difference.

Oh, I think I get it now: you're asking to put all information from a single node per cycle onto one line (i.e. input time/power/voltage/temperature and output RUL/SoC/TMP all on one line instead of the separated inputs and outputs as they are right now). I'll see if I can make the changes, but I'll have to make it clear that the input temperature and output temperature values are different.

kmdalal:
> > The printouts look separated by hard returns for me. Widening the window made no difference.
> 
> Oh, I think I get it now: you're asking to put all information from a single node per cycle onto one line (i.e. input time/power/voltage/temperature and output RUL/SoC/TMP all on one line instead of the separated inputs and outputs as they are right now). I'll see if I can make the changes, but I'll have to make it clear that the input temperature and output temperature values are different.

Yes, but it wasn't until this explanation that I see that the group separate input and output!  An easier change that may do the trick is have it print "Node N input ..." and "Node N output...".   

A one-line solution would be nice but harder to design to be readable and is low priority.

Lumsterling:
> Yes, but it wasn't until this explanation that I see that the group separate input and output! An easier change that may do the trick is have it print "Node N input ..." and "Node N output...".

Yeah, I think I'll just make it more clear which parts are input and output, as well as when debug data from a new cycle begins. That shouldn't be too difficult.

ow_power_system/src/PredictionHandler.cpp
```diff
@@ -14,7 +16,7 @@ using namespace PCOE;
 
 // The indices use to access temperature information.
 // This might change to median SOC or RUL index or fixed percentile.
-//
+// (predecessor comment to SU22, unsure exactly what it means)
```
kmdalal, line 19: I don't know what the preceding comments mean, either, but what does SU22 mean?

ow_power_system/include/PowerSystemPack.h
```diff
@@ -29,6 +35,33 @@ using PrognoserVector = std::vector<PrognoserMap>;
...
+const int NUM_SAMPLES = 100;
+
+// Change this value to modify the maximum RUL estimation output from the Monte
+// Carlo prediction process. Lower values mean faster performance in the event
+// a RUL prediction would exceed the horizon value, but it also means the prognoser
+// can't return RUL values higher than this.
+// The default value is 10000, but this is slow in situations where the prognoser
+// predicts a value that high (or higher than that). Testing and discussion
+// is needed to determine the ideal value for this.
+const int MAX_HORIZON = 10000;
```
kmdalal, line 55: Should specify this is seconds, or better yet include it in the name, e.g. MAX_HORIZON_SECS.

Outside the scope of this PR, but I don't see how we can do useful prognosis with a value even this small (less than 3 hours) given the timeline of the mission.

ow_power_system/src/PredictionHandler.cpp
Lumsterling: 'SU22' stands for summer 2022. I figured it was best to date comments like that for the future. I'll go back and remove the abbreviation tomorrow morning though, it's not particularly necessary here.

ow_power_system/include/PowerSystemPack.h
Lumsterling: The horizon isn't the only value that affects GSAP performance, at least; sample count is significantly more important. The problem is that we don't really know how big of a dip in accuracy results from much lower sample counts, but if we need RUL predictions in the realm of ~20 hours (72000 seconds) I wouldn't be surprised if we have to reduce sample count to 10 or even lower.

The other option could be increasing the profile/timestamp increment to something longer, like 5 or even 10, so instead of requiring GSAP to return 24 predictions in 2 seconds it could be every 5 or 10 seconds. This would require a lot of updating across the power system and possibly other components, though.

kmdalal (approved):
I tested this a lot, under both nominal and fault conditions, and the behavior has been well summarized in this PR.  The problems all seem all out of scope, but aside from that the approach and implementation seem reasonable.  We've learned a lot about GSAP and power modeling.  Great work!

PR #276, opened 2022-10-07
Oceanwater-994: Battery pack model

Lumsterling:
## Linked Issues:
| Jira Ticket 🎟️ | [OCEANWATER-994](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-994) |
| Github :octocat:  | [OCEANWATER-994-Implement_Battery_Pack_Model](https://github.com/nasa/ow_simulator/tree/OCEANWATER-994_Implement_battery_pack_model) |

***(As of 5/31/23, this draft was fully rewritten and shifted into an active PR)***

**UPDATE (6/7/23)**: This is a large PR for merging a full battery model rework into _noetic-devel_. To my understanding, the power system is now in a testable state. This PR is the product of months of work, however, and the CPU load of the power system is now infeasibly high. This is being worked on ([OW-1163](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1163)), though the tests within this PR are still valid.

**UPDATE (6/8/23)**: Created a new branch ``OW-1163`` (based off ``noetic-devel``) for the JIRA ticket for improving the performance of the pack model, and updated this PR to merge into that ticket instead. This way, this PR can continue and potentially merge into a copy of ``noetic-devel`` while the performance work is underway.

**UPDATE (6/23/23)**: Updated wording to reflect name changes of the classes. Note that the main file remains ``PowerSystemNode``, where 'node' in this context refers to the ROS node for the power system. There are also battery nodes represented by the ``PowerModelHandler`` class, which are different.

Note that GSAP's ``v1.0.3-OW`` version is necessary for ideal behavior involving these changes. _noetic-devel_ already updated its installation instructions to mention the new tag.

## Summary of Changes
* Currently, a single-cell [simple prognoser](https://github.com/nasa/GSAP/tree/master/examples/simple) is used to generate end-of-discharge (EoD) predictions at a rate of 0.5Hz. This does not scale into a multi-cell implementation as calls to simple prognosers are sequential and must complete their previous prediction before being allowed to start on another one. Instead, **the battery model now uses [asynchronous prognosers](https://github.com/nasa/GSAP/tree/master/examples/async)** that allow multiple EoD predictions to process at once.
  * Each asynchronous prognoser represents one parallel node in the battery pack; with an expected 24 parallel nodes, there are now 24 asynchronous prognosers running at once.
  * While these individual prognosers typically have a lower prediction frequency than 0.5Hz, the power system is able to continue looping separately and simply updates its published data as prognosers return predictions on their own time.
* ``PowerSystemNode`` was reworked to utilize multiple ``PowerModelHandler`` classes as well as ``PredictionHandler`` classes (adapted from ``PredictionPrinter`` in the asynchronous prognoser example).
  * ``PowerModelHandlers`` generate inputs while ``PowerSystemNode`` handles sending and receiving inputs and outputs to and from the power system (e.g. publishing RUL, SoC, and battery temperature). **The inputs and publications of the power system from an external perspective should be identical**, even though they may be handled differently now. For example, the published RUL value is the minimum of all predicted RUL values from the asynchronous prognosers rather than just the output of one, but one RUL value is still published at the same 0.5Hz rate as before.
* **Fault injection and mechanical power also required reworking** to handle multiple prognosers running at once. Any additional power draw on the battery, be it from a fault injection or an arm movement that requires mechanical power, is now spread evenly across all models rather than the full strength on one (or all).
* (*as of 6/13/23*) The custom fault function was changed to read every line from a custom fault .CSV instead of every other line, as per [OW-1007](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1163). The change was done here rather than in ``noetic-devel`` because the restructuring of the power system in this PR means that it would have to be changed again anyways once this is merged.

## Tests Overview
* For all tests, check ``ow_power_system/config/system.cfg`` and ensure the ``print_debug`` flag is set to ``true`` so that battery stats are printed to the terminal during runtime.
  * The specific battery stats to be printed are filtered by the five other flags below ``print_debug``, and which ones should be ``true`` will be specified in the individual tests.
* It is possible to do all 3 tests in one run of the simulation, though not recommended due to flooding the main terminal with printout statements if all debug flags are set to ``true`` at once.

**UPDATE (6/1/23):**
* This PR likely will induce a much higher CPU load, simply due to 24 prognosers running at once compared to only one before. Take note of ``max_horizon`` and ``num_samples`` within ``system.cfg``, and modify them for performance if desired:
  * ``max_horizon`` (default ``10000``) is the highest value a GSAP prediction can return for RUL. If the prediction tries to go beyond this value, it ends prematurely and returns this value instead. This parameter only impacts performance on predictions that would exceed its value, but realistically the default may be too low for battery predictions. After all, 10000 seconds is less than 3 hours, and a full charge should probably last longer than this.
  * ``num_samples`` (default ``100``) is the number of times a GSAP prognoser re-runs a prediction for accuracy. It is possible the default is far too high for the simulation, however, with 24 prognosers each running 100 samples per cycle. If the performance is tanking, try reducing this value, though discussion is needed to determine how much of a reduction is viable with accuracy in mind.
* The number of concurrent models is defined in ``PowerSystemNode.h``, and is set to ``24`` as per Chetan's specifications. It is possible that this many models puts too much of a performance strain on the simulation however. Modify this value if desired, but because it is a compile-time constant, any change will require rebuilding the simulation.

## Test 1 (Published values)
* This test confirms the power system publishes RUL/SoC/temperature identically to the single-cell implementation.
* *Relevant ``print_debug`` flag(s): ``timestamp_print_debug``, ``outputs_print_debug``, ``topics_print_debug``*
* Open up 4 terminals
  * In the 1st terminal, build and run the simulator in the OceanWATERS directory. The world shouldn't matter, though I used the ``atacama`` world (``roslaunch ow atacama_y1a.launch``)
* In the other three terminals, run one of the following commands in each:
  * ``rostopic echo /battery_remaining_useful_life``
  * ``rostopic echo /battery_state_of_charge``
  * ``rostopic echo /battery_temperature``
* The 1st terminal should display the current loop's timestamp, the currently stored prediction values of each individual model, as well as the ``min_rul``, ``min_soc``, and ``max_tmp`` values which correspond to the min or max of each respective value.
* **These statements should output to the terminal at roughly a 0.5Hz rate**. The timestamps associated with each printout should assist in confirming a ~2s interval between batches of statements.
* Note that some models will update their values sooner than others. This is expected because the prognosers are unlikely to return predictions at the same time.
* **Check the other 3 terminals and ensure their reported values line up with ``min_rul``, ``min_soc``, and ``max_tmp``**. The timing of these published values should line up with the main terminal's outputs.

## Test 2 (Fault Application)
* This test confirms any additional power draw from fault injection is evenly applied across all running models.
* *Relevant ``print_debug`` flag(s): ``timestamp_print_debug``, ``inputs_print_debug``*
* Open up 1 terminal
  * Build and run the simulator in the OceanWATERS directory
* Once the simulation has started, navigate to the ``Dynamic Reconfigure`` tab on the bottom of the RQT window, and from there, to the ``faults`` section. Scroll down until you see the section governing high power draw and custom fault profiles.
  * Adjust the value of ``high_power_draw`` to some significan nonzero value, and activate it by checking ``activate_high_power_draw``. Note the new maximum of 500W for testing.
* **Monitor the debug statement outputs and confirm the additional power draw injected is distributed evenly across all models and adds up to the value set in RQT**. If the ``high_power_draw`` value is changed mid-injection, it should update for the next cycle if the value was modified.
* Create a new fault profile within the ``profiles`` directory of ``ow_power_system``. Reference ``example_fault.csv`` to understand how the file should be formatted.
  * The exact values within the custom fault do not matter, as long as they are valid; the chief takeaway from this test is that the additional wattage from the custom fault is distributed evenly. For example, a custom fault with ``120`` in the ``wattage`` column for the first 10 indices and ``240`` for the next 15 should result in a 5W increase to each model for the 10 cycles following injection (assuming 24 simulated nodes) and a 10W increase for the 15 cycles afterwards.
  * Inject the modified fault profile using the ``custom_fault_profile`` section in RQT.
  * Monitor the debug statement outputs and **confirm the additional power draw injected is evenly distributed across all models and adds up to the value set at the relevant index of the fault profile. Additionally, the profile should apply values from each line sequentially instead of skipping every other line** (so if there is a ``120W`` draw for the first 10 indices, the 10 cycles following injection should show the relevant increase. Previously, it would only apply for the 5 cycles following injection).

## Test 3 (Arm Movements)
* This test confirms mechanical power created from arm movements is properly applied across all battery nodes.
* *Relevant ``print_debug`` flag(s): ``inputs_print_debug``, ``mech_power_print_debug``*
* Open up 2 terminals
  * In the 1st terminal, build and run the simulator in the OceanWATERS directory
  * In the 2nd terminal, run PLEXIL (``roslaunch ow_plexil ow_exec.launch``).
* Run some PLEXIL plans in the 2nd terminal. Some examples are ``ReferenceMission1.plx``, ``ReferenceMission2.plx``, and ``TestActions.plx``. Run at least two of these.
  * While a plan is running, the main terminal should display an increase in the raw/applied mechanical power each cycle. **The applied power is simply a moving average of the last 25 raw mechanical power values, and is the value applied to the battery**.
  * **The distributed power should be the applied power divided by the number of nodes, and the inputs to each model for the cycle should be around 1W plus the distributed power**. E.g. if the applied power is ``48W`` for the cycle, each model in the 24-node pack should display an input of ``1 + (48 / 24) = 3W`` draw. There may be slight variations in the displayed power draw, but they should be minor.
  * **If there are any errors in the plan executions, confirm that they are distinct from the errors seen on the ``noetic-devel`` branch**. This PR should not have changed anything that affects plan execution, and as of writing this PR (5/31/23) there may be issues with ``ReferenceMission1`` that have not been addressed yet in ``noetic-devel``.


AstroStucky:
### Test 1 Results
Worked as described, but I thought it was strange that the RUL stayed fixed squarely at 10000 seconds. I even ran Continuous.plp, which did lower SoC and raise the temperature a tiny bit, but RUL still stayed at 10000 seconds. 

### Test 2 Results
Test worked as described.

Why is there 1 W applied on each node? Does 1 W represent the idle power draw of the lander? Shouldn't the 1 W also be divided across all nodes just like other power draws?

### Test 3 Results
Worked as described.

### Comments on Performance
I'm concerned about how much compute power the simulator will demand by default with these changes. I have one of the faster computers on the team, and just idling the simulation ate up 100% of all 16 of its cores. The simulation still managed to mostly run in realtime, but barely. If I were to increase the update rate to 1200, as I do very commonly for testing, the most the realtime factor would get up to is only 1.2x. I'm used to running the sim at 3x with no problems. This doesn't even get into the concern for older, less powerful machine, which I strongly suspect will not even be able to run the simulation in realtime. 

It comes as a surprise to me that GSAP would be such a CPU hog and I am wondering if this can improved by tweaking its simulation parameters. Alternatively, I think we should consider **A)** using significantly fewer power nodes by default, maybe 6 or 8, and **B)** making the number of power nodes a configurable parameter for users that want to run more or less at once. 

### Comments on Mechanical Power
This may be out of scope for this PR, but it's relevant to my tests.

I understand the need for a moving average window on mechanical power, but the width being used is far too large making the value meaningless in terms of what the arm is doing. If the arm begins moving from a stationary position, the window average power will remain nearly zero for a long time. If the arm stops moving it will remain at a non-zero but slowly declining value for a long time. If the arm encounters a sudden high torque situation the average power will not change much at all. 

I think the solution is to use a much smaller window width. Something between 0.1 to 1 seconds. This may need to be tweaked to avoid spikes, but we should also consider whether or not the spikes in power (effort * velocity) are physical or not.

Lumsterling:
> ### Test 1 Results
> Worked as described, but I thought it was strange that the RUL stayed fixed squarely at 10000 seconds. I even ran Continuous.plp, which did lower SoC and raise the temperature a tiny bit, but RUL still stayed at 10000 seconds.

Ah, this and the performance comments made me realize I forgot to list the parameters that affect performance: ``num_samples`` and ``max_horizon``. They're both located in ``system.cfg``. I'll update the PR with explanations of each, but essentially ``max_horizon`` is the RUL cap; if the prognoser prediction would exceed ``max_horizon``, it stops prematurely and caps the RUL value there. Default is ``10000``, but the RUL of a full or near-full battery should exceed 10000 (after all, 10000 is a little under 3 hours, and a full charge should probably last a lot longer than that). That said, I'm not knowledgeable in accurate ranges of RUL, we'd probably have to consult Chetan on that. Bottom line is the behavior you're seeing is probably expected, and in fact ``max_horizon`` might need to be greatly increased.

``num_samples``, on the other hand, is the number of times each GSAP prognoser re-runs predictions for accuracy (to my konwledge). ``100`` is the default, but honestly that may be far too many on a per-cycle basis (especially when there are 24 prognosers running concurrently), and this very likely ties into the performance bottleneck you're worried about. I don't know what an acceptable number for this is either though--it was never discussed--but I can do some simple tests on Monday.

If you're able to, could you try tweaking ``num_samples`` within ``system.cfg`` on your end as well? I'm curious if much lower values like ``10`` or ``20`` result in a significant performance boost.

> Why is there 1 W applied on each node? Does 1 W represent the idle power draw of the lander? Shouldn't the 1 W also be divided across all nodes just like other power draws?

I had the same question and asked Chetan about it earlier this week; he said each parallel node in the pack should have a baseline 1W draw (which would indeed add up to a 24W total idle draw for the pack). 

> It comes as a surprise to me that GSAP would be such a CPU hog and I am wondering if this can improved by tweaking its simulation parameters. Alternatively, I think we should consider **A)** using significantly fewer power nodes by default, maybe 6 or 8, and **B)** making the number of power nodes a configurable parameter for users that want to run more or less at once.

Chetan wanted the pack model to contain 24 nodes, but it's possible this is too much to process at once. I'll have to talk with him regarding this, but I'll also see on Monday if heavily reducing ``num_samples`` improves performance significantly. Maybe a parameter change like that could make a significant difference.

> I understand the need for a moving average window on mechanical power, but the width being used is far too large making the value meaningless in terms of what the arm is doing. If the arm begins moving from a stationary position, the window average power will remain nearly zero for a long time. If the arm stops moving it will remain at a non-zero but slowly declining value for a long time. If the arm encounters a sudden high torque situation the average power will not change much at all.
> 
> I think the solution is to use a much smaller window width. Something between 0.1 to 1 seconds. This may need to be tweaked to avoid spikes, but we should also consider whether or not the spikes in power (effort * velocity) are physical or not.

Yeah, I think we talked about the 25-cycle moving average in the tag-up earlier today. ``25`` is just the size of the window that was present before I started work on this project, and I could easily tweak it. Bear in mind, though, that the main loop for the power system runs at 0.5Hz, meaning it only ever adds a new value to the moving average every 2 seconds at most. I don't think it could allow for a 0.1-1s window unless we simply made the moving average window size 1, and even then it would be a 2s window instead.

That said, I wonder how viable it is to increase the frequency of the power system's main loop. The reason it is 0.5Hz is simply because that was originally the response time of GSAP's simple prognoser for the single-cell model, so it lined up. My changes mean the loop no longer has to line up with GSAP's response time though, so maybe it's possible I could tweak the loop frequency to 1Hz or potentially even higher. The change would probably be far-reaching and require a fair bit more work, but it sounds like that might be ideal.

Lumsterling:
Having done some performance tests regarding ``num_samples``, I've found so far that drastically reducing that number is helpful, though I need to talk with Chetan (and/or Chris) to determine how reasonable a change is with regards to prediction accuracy.

On my setup, reducing ``num_samples`` to ``10`` kept my CPU usage of the power system at a consistent 70%. 
Reducing it to ``5`` caused CPU usage to bounce between 50-70%. (*This is because all nodes seem to complete their predictions within the 0.5Hz cycle window at this value of ``num_samples`` or lower, so CPU usage jumps back and forth between the nodes waiting for the next cycle and generating predictions*).
Reducing it to ``2`` caused CPU usage to bounce between 30-50%.
Reducing it to the minimum ``1`` caused CPU usage to drop to a consistent 30%.

However, the original single-cell model's CPU usage (which had a ``num_samples`` value of ``100``) was only ever around 6%. I think it's to be expected that the new pack model is significantly heavier on the performance side of things, but I'm surprised reducing ``num_samples`` to the minimum possible value still results in a pretty high performance load.

Reducing the number of nodes, on the other hand, had a very strong impact on performance as expected. I'm not sure if that is feasible, though, with regards to the original power model goal (simulating a 24-parallel-node battery pack). I'll need to discuss with Chetan.

Lumsterling:
Set this PR to 'on hold' while working on reducing the CPU load of the battery pack.

Lumsterling:
Changed the base branch to a current copy of ``noetic-devel`` so work & testing on this PR can continue while progress on pack performance continues. As such, I also removed the 'on hold' tag.

ow_faults_injection/cfg/Faults.cfg
```diff
@@ -100,7 +100,7 @@ ft_sensor_faults.add("signal_noise_failure", double_t, 0,
 
 power_faults = gen.add_group("power_faults")
 power_faults.add("high_power_draw",   double_t, 0,
-    "High power draw increment", 1.0, 1.0, 20.0)
+    "High power draw increment", 1.0, 1.0, 400.0)
```
kmdalal, line 103: I'm guessing that this value should be 20 times the number of cells.  So for 24 cells, shouldn't it be 480?  And we'd also need to document that cell count and this max needs to be kept in sync.

Lumsterling: The exact value is probably something we should briefly discuss with Chetan, though yes, 20 times the number of cells is probably a good benchmark. I just set a basic higher value for general testing purposes.

That said, if we wanted the max wattage per cell to be 20 then we'd probably also want to take into account the 1W baseline draw each cell has, so the actual max would be something like 456W instead. Such a value seems kind of arbitrary? Again, it's something to discuss.

AstroStucky: I don't think this needs to be done for this PR, but I think it makes the most sense to make this value 20, and track it per cell, ideally issuing only a single warning when the value is breached for any cell instead of 24 warnings for each cell.

Lumsterling: For what it's worth, I updated the mechanism for that specific warning. It'll now only output one warning message that lists every node that went over the 20W cap.

That said, because power is distributed evenly across all nodes, if one node is over the cap then every other node is also over the cap, so all nodes are listed any time the limit is exceeded. It also might not make sense to individually list each node in such a warning anyways; the point is that the pack acts as a single larger battery, right?

kmdalal: I agree with the previous comment, that warning seems sensible if the battery as a whole exceeds a power threshold.

The warning may make sense for individual cells if somehow one can hit the threshold without the whole battery reaching its threshold.   This doesn't seem possible with our current battery,  but maybe it would become possible if we start modeling intra-battery faults and failures.  

Would be good for @chetankul to review these comments.

kmdalal:
In Test 2, what should the custom fault profile be?

Lumsterling:
> In Test 2, what should the custom fault profile be?

The exact profile doesn't matter, as long as whatever values you enter for the 'wattage' column are shown to distribute evenly amongst all nodes in the debug printouts. An example .CSV could contain 120W for the first 10 indices (which should show an increase of 5W to each individual node for the following 10 cycles).

I'll update the PR to suggest this.

kmdalal:
> The exact profile doesn't matter, as long as whatever values you enter for the 'wattage' column are shown to distribute evenly amongst all nodes in the debug printouts.

This strikes me as a test for the custom power profile, but the power distribution across nodes seems adequately tested with the high power draw fault.

Lumsterling:
> > The exact profile doesn't matter, as long as whatever values you enter for the 'wattage' column are shown to distribute evenly amongst all nodes in the debug printouts.
> 
> This strikes me as a test for the custom power profile, but the power distribution across nodes seems adequately tested with the high power draw fault.

You're right, though now that the minor change in OW-1007 is part of this PR, Test 2 can also serve as a way to confirm the change from reading every other line in a custom profile to each line. That wasn't the original intention, but it might be useful as such.

Otherwise, yeah, the high power draw fault works for this too.

AstroStucky (changes requested):
Going off of my tests from last week, it mostly passes, but I am a still confused about the 10,000 max RUL value. Why does it cap at 10000 if we expect our battery to last much longer than 3 hours? 

My high-level review of the code is that I would like to see more constrained scope of certain variables and more cohesive structs/classes. In one case variables get defined in one class, but only ever need to be modified in another. The `PowerNode` struct contains a `PowerSystemNode` and several other variables closely related to `PowerSystemNode` that could probably simply be contained by `PowerSystemNode`, and the struct would not be needed. 

The above paragraph makes reference to certain code comments I have made, and will make more sense after reading those. Please don't hesitate to reach out for clarification on anything.

AstroStucky:
One more small comment. I have heard battery cells referred to as nodes, which is also the name of a process registered in ROS, and I get the impression this has caused some confusion in the naming of variables and classes. 

For instance, I seem to recall that `PowerSytemNode` was originally named that way because it was the ROS node for the power system. That name is now reused for a class the defines a power cell within the battery. I wonder if it would be better to not name power cells "nodes" at all since it collides with ROS terminology. Would naming them "cells" be more appropriate? I do not know battery terminology, and would like to hear what others think.

Lumsterling:
> Going off of my tests from last week, it mostly passes, but I am a still confused about the 10,000 max RUL value. Why does it cap at 10000 if we expect our battery to last much longer than 3 hours?

To my understanding, the single-cell model currently implemented in ``noetic-devel`` predicts values far lower than ``10000`` even at startup (I believe it tends to start around ``3000``). I don't necessarily know the exact reason for this, but I imagine it's due to having only 1 cell represent the entire battery, which probably wouldn't last that long even on minimal usage. As such, the ``max_horizon`` cap of ``10000`` was reasonable for the current build, but I don't think it will be once these changes go through (at which point the RUL starts out upwards of ``27000``). It's something to confirm with Chetan first.

I'll be looking through the suggestions from your review and working on them through the rest of the day.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -15,52 +17,31 @@ using namespace std;
...
-  if (!initPrognoser()) {
-    ROS_ERROR("Failed to initialize power system prognoser.");
-    return false;
-  }
-
-  if (!initTopics()) {
-    ROS_ERROR("Failed to initialize power system topics.");
-    return false;
-  }
+  m_init_time = system_clock::now(); // Taken from initPrognoser()
```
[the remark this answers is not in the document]
Lumsterling: ``generateVoltageEstimate`` is called every cycle to update the voltage value sent into GSAP's prognosers. To my understanding it's supposed to mimic a sensor while using a random distribution. The minimum and maximum values are based off of the base voltage (``m_base_voltage``), the expected battery lifetime (``m_battery_lifetime``), and how long the simulation has been running (``timelapse``). But as you said, ``timelapse`` is based on ``system_clock::now()`` instead of ``ros::Time::now()``. I wouldn't know why, except for the simple explanation that it might just be a mistake and should be corrected.

I could easily change it, but should probably confirm with Chetan first to be safe.

ow_power_system/include/PowerSystemNode.h
```diff
@@ -12,76 +22,42 @@
...
 
-  double m_initial_power = 0.0;         // This is probably always zero
-  double m_initial_temperature = 20.0;  // 20.0 deg. C
-  double m_initial_voltage = 4.1;       // Volts
   double m_min_temperature = 17.5;      // minimum temp = 17.5 deg. C
   double m_max_temperature = 21.5;      // maximum temp = 21.5 deg. C
   double m_battery_lifetime = 2738.0;   // Estimate of battery lifetime (seconds)
   double m_base_voltage = 3.2;          // [V] estimate
   double m_voltage_range = 0.1;         // [V]
   double m_efficiency = 0.9;            // default 90% efficiency
```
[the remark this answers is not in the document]
Lumsterling: A lot of the header variables essentially contain filler values as you mentioned. That was simply how they were set up before I started work on this. I'll go ahead and remove the values, since we wouldn't exactly want to continue if ``system.cfg`` failed to initialize a variable anyways.

kmdalal:
My understanding is that the max horizon of 10,000 seconds (settable in a config file) was chosen as a compromise between performance and practicality.  The battery nominally has a one month life at start of mission, but the longer the horizon, the longer the prognoses will take.

If a prognoser is run onboard in a real mission, its performance is critical.

kmdalal (approved):
The tests all passed, and I did additional tests of my own that further stress-tested the system and tried other plans.

@AstroStucky has done a pretty thorough code review and I agree with his suggestions; added only a few minor ones myself.

Overall I'm happy with this and appreciate a few things in particular: the detail of explanation in this PR (will transfer nicely to user docs) and the number of configurable parameters.

As we all agree, the computing performance is unacceptable (I got a real-time factor of 0.15 in Gazebo), but this will be addressed in the target branch, and there are several methods already on the table.

Great work!

ow_power_system/src/PowerSystemPack.cpp
```diff
@@ -0,0 +1,784 @@
...
+            msgIndex++;
+            break;
+          case 3:
+            highDrawMsg += ", " + header;
+            msgIndex = 0;
+            break;
+          default:
+            break;
+        }
+      }
```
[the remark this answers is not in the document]
Lumsterling: I think, in hindsight, the complexity is unnecessary. It stems from trying to list every node that goes over the power draw limit, as well as its calculated power draw before the limit, in the same warning message, but the reality is that (with the current implementation) if one node goes over the limit, every other node will too. I'll probably scrap this logic and just make a simple one-line warning.

ow_power_system/src/PowerSystemPack.cpp
```diff
@@ -0,0 +1,784 @@
...
+        
+        // Compile a vector<shared_ptr<DoubleMessage>> and then individually
+        // publish each component.
+        std::vector<std::shared_ptr<DoubleMessage>> data_to_publish;
+        double input_power, input_tmp, input_voltage;
+
+        // HACK ALERT (May 2023):
+        // The first batch of data (sent in during startup) seems to be lost to 
+        // the void. This is a bandaid fix to send the initial data in twice
+        // for each node on the next loop.
```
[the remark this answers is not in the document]
Lumsterling: GSAP treats the initial data sent into its prognosers as the starting conditions of the battery, which is a different and specific state than the randomly distributed wattage/voltage/temperature generated during main cycles. If you don't send in this initial data and instead jump straight to the generated values, the starting RUL/SoC values are invalid (RUL starts anywhere from 1-3k and SoC starts at 7-15%).

I found previously that unless this initial data was sent in twice, GSAP seemingly wouldn't receive it and the prognosers would output invalid predictions. However, I did some extensive testing this morning and it I don't observe the problem behavior anymore with sending in the starting data once to each prognoser. I'm not sure what changed to address the issue (it may very well have been an issue with my own past implementation), but I'll go ahead and remove this stipulation since I can't replicate this issue anymore.

ow_power_system/src/PowerSystemPack.cpp
```diff
@@ -0,0 +1,784 @@
...
+        }
+
+        // DEBUG PRINT
+        if (m_inputs_print_debug && (m_nodes[i].model.timestamp > 0))
+        {
+          std::string header = "N";
+          if (i < 10)
+          {
+            header += "0";
+          }
```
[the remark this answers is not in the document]
Lumsterling: I simplified both cases to a single 'if' statement, and I'm not sure if that is worth making a separate function for it. However, it might be worth making standalone functions for each of the debug printouts rather than having them defined within the loop (since it's already pretty large). I'll go ahead and implement that.

ow_power_system/include/PredictionHandler.h
```diff
@@ -0,0 +1,49 @@
...
+  PredictionHandler& operator=(const PredictionHandler&) = delete;
+  void processMessage(const std::shared_ptr<Message>& message) override;
+private:
+  double findMedian(std::vector<double> samples);
+
+  // References used by other classes to process EoD predictions.
+  double& m_rul_ref;
+  double& m_soc_ref;
+  double& m_temp_ref;
+  MessageBus& m_bus;
```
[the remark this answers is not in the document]
Lumsterling: Maybe I'm misunderstanding, but ``PowerSystemPack`` creates prognosers using the buses, and then publishes input data each main cycle to the corresponding bus (though ``PredictionHandler`` is indeed the only class that subscribes to the bus).

I could maybe rework the loop so that the ``Pack`` has to send input data to ``PredictionHandler`` which then publishes it to the now-private bus, though I'm not as sure about this change.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -71,154 +52,14 @@ bool PowerSystemNode::loadSystemConfig()
...
-      accumulate(begin(m_power_values), end(m_power_values), 0.0) / m_power_values.size();
-
-  Float64 mechanical_power_raw_msg, mechanical_power_avg_msg;
-  mechanical_power_raw_msg.data = power_watts;
-  mechanical_power_avg_msg.data = mean_mechanical_power;
-  m_mechanical_power_raw_pub.publish(mechanical_power_raw_msg);
-  m_mechanical_power_avg_pub.publish(mechanical_power_avg_msg);
-
-  m_unprocessed_mechanical_power = mean_mechanical_power;
+  m_unprocessed_mechanical_power = mechanical_power;
```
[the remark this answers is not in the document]
Lumsterling: The naming convention (and comment a few lines below) imply accumulation, but it's been a value assignment for as long as I can remember. I'll do some tests and see if changing it to ``+=`` affects any behavior, though I don't think it will because I remember checking this block of code some time ago and finding that the ``if`` statement always ran, so it never had to accumulate in the first place.

Essentially, the ``if`` statement was there to withhold mechanical power from being updated if the loop was currently processing the previous mechanical power value in ``runPrognoser`` (since it's a callback function it could potentially interrupt). Because the assignment allows for one interrupt with no issue, there would only be an observable problem of losing unprocessed power if two callback interrupts occurred before ``runPrognoser`` finished once. This never happens because of how long it takes for GSAP to return predictions.

That said, I'll probably still change this to a ``+=`` since at worst it won't affect anything and it will make more sense with the naming of the variables and the comments.

ow_power_system/src/PowerSystemPack.cpp
```diff
@@ -0,0 +1,784 @@
...
+  owl_msgs::BatteryTemperature tmp_msg;
+
+  // Calculate the min_rul, min_soc, and max_tmp from the new data.
+  for (int i = 0; i < NUM_NODES; i++)
+  {
+    // If RUL is infinity, set it to the maximum horizon value instead.
+    if (isinf(m_EoD_events[i].remaining_useful_life))
+    {
+      m_EoD_events[i].remaining_useful_life = m_max_horizon_secs;
+    }
```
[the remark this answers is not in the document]
Lumsterling: EoD events return an infinite RUL whenever the prediction would generate a RUL value beyond the limit set by ``max_horizon``. As such, with the current max being ``10000``, this happens nearly every loop in every node at the start when idling (increasing ``max_horizon`` might be ideal as a result).

Regardless, it's not an error or warning-relevant behavior, just part of how ``max_horizon`` affects predictions.

Lumsterling:
I've now completed changes to address most of the PR feedback; with regards to comments involving ``system_clock``, those will be looked at in a future ticket that will be created following the completion of this PR. Additionally, the moving average window with regards to mechanical power was also not changed; I think this is better suited to another ticket and further discussion.

Because there were so many changes, especially in the final commit which changed class names and required renaming of many different variables, I'm re-requesting a review (from Mike; to my understanding Thomas will be out of the office for a while) to help ensure nothing new broke as a result. I re-ran the tests on my end and everything seems to work as before, but you never know.

AstroStucky:
Actually I am in the office today and next week. You may be thinking of Terry.

I agree with leaving clock changes and any kind of an investigation/changes around the moving average window for future tickets.

Lumsterling:
> Actually I am in the office today and next week. You may be thinking of Terry.

Oops; in that case, I'll re-request a review from you as well if you've got the time.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -4,194 +4,807 @@
...
+      // do not publish the data to prevent crashes.
+      if (m_power_models[i].previous_time != m_power_models[i].model_info.timestamp)
+      {
+        auto timestamp = START_TIME + std::chrono::milliseconds(
+          static_cast<unsigned>(m_power_models[i].model_info.timestamp * 1000)
+        );
+        
+        // Compile a vector<shared_ptr<DoubleMessage>> and then individually
+        // publish each component.
+        std::vector<std::shared_ptr<DoubleMessage>> data_to_publish;
```
AstroStucky, line 205:
```suggestion
```
This variable looks to be no longer in use.

ow_power_system/include/PredictionHandler.h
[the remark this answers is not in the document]
AstroStucky: I see what you mean, MessageBus is actually used almost entirely by `PowerSystemNode`. I guess I should've argued for the opposite if anything, that MessageBus should remain exterior to PredictionHandler, but the `unsubscribe` call in its destructor would make that tricky. 

I think it's fine how it is.

ow_power_system/src/PowerSystemNode.cpp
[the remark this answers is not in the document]
AstroStucky: I won't take the time to look at it in depth right now, but I think it will be worth investigating these values and how they work in that ticket about the mechanical power moving average window. For now, if the if statement is always hit, I think your change is fine, but there is something about how this moving average works that bothers me. I may also just be misunderstanding something.

ow_power_system/src/PowerSystemPack.cpp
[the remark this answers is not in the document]
AstroStucky: Now that I understand that this is basically a simulate-until limit, this makes more sense to me. I agree, now warning or error is needed.

AstroStucky:
Looked over my requested changes, and responded to some comments from the author. I don't have time for a test or a full re-review, but you can expect one from me early next week. 

I like what I see so far. Thanks for incorporating all my feedback. Superb work! 😃

AstroStucky (approved):
Re-ran tests and they all pass. Great work!

ow_power_system/include/PowerModelHandler.h
```diff
@@ -0,0 +1,113 @@
+// The Notices and Disclaimers for Ocean Worlds Autonomy Testbed for Exploration
+// Research and Simulation can be found in README.md in the root directory of
+// this repository.
+
+// This is the header file for the PowerModelHandler class, which handles
+// the simulation of a single model within a PowerSystemNode. It stores the
```
kmdalal, line 6: I've not heard of a "power model" as a singular entity.  Is this explained somewhere?  A diagram with the relations would be helpful.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -2,305 +2,354 @@
...
+    {
+      double temp_draw = m_power_models[i].model.RunOnce();
+      // Signal the warning message to be displayed after the loop.
+      if (temp_draw > 0)
+      {
+        // Current implementation means if one input exceeds the draw limit,
+        // all inputs will exceed at the same amount, so it doesn't matter which
+        // cell input we take.
+        excessive_draw = temp_draw;
+      }
```
kmdalal, line 191: This block is a bit confusing.   Would help to explain what the return value of runOnce() means.  I think the logic could also be accomplished with just one of `temp_draw` and `excessive_draw`.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -2,305 +2,354 @@
...
-  m_mechanical_power_avg_pub.publish(mechanical_power_avg_msg);
+    // Print a warning message if any models computed an excessive power draw.
+    if (excessive_draw > 0)
+    {
+      ROS_WARN_STREAM("Power system computed excessive power input of "
+                      << excessive_draw << "W for one or more GSAP "
+                      << "prognosers.\nInput was capped at "
+                      << m_max_gsap_input_watts
+                      << "W for each model over the limit.");
+    }
```
kmdalal, line 248: I find it odd that `excessive_draw` is not compared to `m_max_gsap_input_watts`.  In fact the latter is used only for a message.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -387,161 +435,373 @@ void PowerSystemNode::injectCustomFault(bool& fault_activated,
...
-  return PrognoserMap {
-    { MessageId::Watts, Datum<double>{ power } },
-    { MessageId::Volts, Datum<double>{ voltage } },
-    { MessageId::Centigrade, Datum<double>{ temperature } }
-  };
+  double power_watts = 0.0;  // This includes the arm + antenna
+  for (int i = 0; i < ow_lander::NUM_JOINTS; ++i)
+    power_watts += fabs(msg->velocity[i] * msg->effort[i]);
+
+  m_power_values[++m_power_values_index % m_power_values.size()] = power_watts;   // [W]
```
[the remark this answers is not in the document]
Lumsterling: I'll go ahead and address it now, since it's a quick and easy change.

ow_power_system/include/PowerModelHandler.h
Lumsterling: Maybe a better name would be ``BatteryModelHandler``? The purpose of the class is to handle the inputs going into whatever battery model GSAP's prognosers are simulating (based on the parameters set in ``prognoser.cfg`` & ``async_prognoser.cfg``). I didn't want to call it a ``PowerSystemCell`` because of the developments with Chetan involving using a different model that represents more than one cell at a time.

Changing the name again shouldn't be too difficult compared to before, I imagine most comments and variables would remain the same from ``PowerModelHandler`` to ``BatteryModelHandler``. That said, if you've got an idea for a different name for the class, I'm all ears.

ow_power_system/src/PowerSystemNode.cpp
Lumsterling: The ``temp_draw`` functionality is mostly a holdover at this point from previously when I had a more involved warning message when a node had excessive power draw applied to it. I think I'll rework this back to what it was before and see about implementing a simpler check for it excessive power draw was calculated.

ow_power_system/src/PowerSystemNode.cpp
Lumsterling: They are compared, just in ``PowerModelHandler`` instead of ``PowerSystemNode``.

ow_power_system/include/PowerModelHandler.h
kmdalal: I can't think of a better name without seeing the bigger picture, which would be nicely conveyed in a UML/entity relation diagram like the one found in https://github.com/nasa/ow_simulator/wiki/Autonomy.  Such a diagram would help understand the power system and be a good addition to the user guide.  

Could you create one?  If you don't have an easy diagramming tool, don't spend a lot of time on it.  You could even sketch one on paper and photograph it.

I use the (non-free) Mac tool Omnigraffle and can create diagrams quickly.  It's much faster than Powerpoint or Confluence's Gliffy.

ow_power_system/src/PowerSystemNode.cpp
kmdalal: Is this issue resolved?

Lumsterling: Yup, I removed it. Forgot to resolve this conversation as a result, whoops.

ow_power_system/include/PowerModelHandler.h
Lumsterling: Sure thing, I'll get to work on that. I was looking at OW-768 (Diagram the Power System) while waiting on the PR reviews a few weeks back, and it got put on the backburner once they came through.

I didn't find Gliffy to be too cumbersome last time I used it, though I was sticking to the basics. I should be able to draft something up today.

kmdalal (approved):
Tests pass, and new code edits look good -- my comments were minor and most have been resolved.  If you want I can retest after the impending class name change.  Great work!

Lumsterling:
> If you want I can retest after the impending class name change.

I think another retest is probably unnecessary at this point; I ran it again myself after the change and it works fine. Looks like this PR is finally complete (sort of)! Still have to deal with the performance issue but at least this works.

PR #277, opened 2022-10-12
Oceanwater 1036 & 1052 - Improved Balovnev model parameters

AstroStucky:
## Linked Issues:
| EPIC ⚡ | [OCEANWATER-1069](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1069) |
| Jira Ticket 🎟️ | [OCEANWATER-1036](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1036) <br /> [OCEANWATER-1052](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1052) |
| Github :octocat:  | # |

## Summary of Changes
* Model bucket and blade parameters made more true to the OceanWATERS scoop model
* Soil parameters made more true to high temperature ice particulates
* A moving max filter is applied the depth input for a more consistent and accurate estimate

## Test
1. Launch atacama_y1a with rqt_gui
`roslaunch ow atacama_y1a.launch use_rviz:=false`
2. In the rqt_gui menu bar select Perspectives > balovnev_analysis. You will now see 4 live plot windows. **Top left** plot will show forces acting on the scoop (both isolated model forces and forces detected by the F/T sensor), **top right** will show joint efforts, **bottom left** will show the value of depth fed into the model (nothing published yet), and **bottom right** will show SoC of the battery. No need to adjust the viewing limits as they should already be well adjusted for the values seen in this test. 
3. Dig a trench by running the following command, which will execute each action client in sequence.
`./unstow_action_client.py && ./grind_action_client.py && ./dig_linear_action_client.py 1.46 0.0 0.01 0.1 -0.1`
Watch all 4 plots. If you need, tick "autoscroll" off when the dig completes in all four plots. In the **top left** plot the forces experienced should lie anywhere between 20 to 100 N, and you will notice that the magnitude of the F/T sensor climbs with the magnitude of the balovnev forces. In the **top right** plot the effort[5] (j_prox_pitch) and effort[7] (j_shou_pitch) will both have climbed with the Balovnev forces to almost 50 N*m. In the **bottom left** plot the depth will have climbed, plateaued around 0.06 m, then dropped to zero. Then finally in the **bottom right** plot the SoC should show a steep decline during the dig operation. 
Screenshot_20221011_202948
4. Finally, command the scoop to dig 10cm deeper. We expect to see unintended behavior here.
`./dig_linear_action_client.py 1.46 0.0 0.01 0.1 -0.2`
This will stall the scoop; however, due to [OCEANWATER-1067](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1067), a stall will result in non-physical behavior whereby the scoop is forced backwards. This will result in a feedback loop that causes the scoop to dig even deeper, which causes the force estimate to climb higher, and the scoop to get pushed back even more. This step is simply to demonstrate the need for the next PR in this Epic. There is no need to inspect the plots.


AstroStucky:
FYI **balovnev_analysis.perspective** does not need to be reviewed. The perspective is only here to assist in tests and I plan to remove it from the epic branch before merging into noetic-devel, along with anything marked `// DEBUG CODE`.

kmdalal:
I'm not seeing a `Perspectives > balovnev_analysis` selection in RQT.

ow_gazebo_plugins/src/BalovnevModelPlugin/MovingMaxFilter.h
```diff
@@ -0,0 +1,54 @@
...
+  void clear() {
+    static constexpr auto fill = -std::numeric_limits<double>::infinity();
+    m_data = std::deque<double>(m_width, fill);
+    m_max = fill;
+  };
+
+private:
+  double m_max;
+  std::deque<double> m_data;
+  size_t const m_width;
```
kmdalal, line 51: Minor: it is written that ordering data members from smallest to largest (roughly) improves memory efficiency.  In this case m_data would be last.

AstroStucky:
> I'm not seeing a `Perspectives > balovnev_analysis` selection in RQT.

Try Perspectives > Import, then select `balovnev_analysis.perspective` in `ow/config/`

ow_gazebo_plugins/src/BalovnevModelPlugin/MovingMaxFilter.h
```diff
@@ -0,0 +1,54 @@
...
+    double popped = m_data.front();
+    m_data.pop_front();
+    m_data.push_back(datum);
+    // update max value if needed
+    if (datum > m_max) { // datum replaces max if it's greater than
+      m_max = datum;
+    } else if (popped == m_max) { // find new max if popped value was the max
+      m_max = *std::max_element(std::begin(m_data), std::end(m_data));
+    }
+  };
```
kmdalal, line 36: Suggest un-inlining this (and clear()) unless there's a good reason to define them here.  My general practice is to only inline simple one-line functions or when speed is more important than program memory.  I _don't_ inline to avoid creating a .cpp file.

kmdalal:
Importing the Balovnev perspective worked.  Nice!  I didn't know such perspectives (plots) were possible.  It may be good to leave this in, and not consider it debug code.

kmdalal (approved):
The test seemed to work -- my plotted values were about as described.  My code comments were minor, no change needed.

ow_gazebo_plugins/src/BalovnevModelPlugin/BalovnevModelPlugin.cpp
```diff
@@ -111,13 +118,10 @@ void BalovnevModelPlugin::onUpdate()
...
 {  
-  if (x <= 0.5 * asin(sin(efa)/sin(ifa)) - efa)
-  {
+  if (x <= 0.5 * asin(sin(efa)/sin(ifa)) - efa) {
     double a = (1 - sin(ifa) * cos(2 * x)) / (1 - sin(ifa));
     return a;
-  } 
-  else 
-  {
+  } else {
```
anthonykoutroulis, line 124: ROS style wants braces on their own lines

ow_gazebo_plugins/src/BalovnevModelPlugin/BalovnevModelPlugin.cpp
```diff
@@ -159,29 +161,42 @@ void BalovnevModelPlugin::computeForces(double vertical_cut_depth)
     + d*a3 * (2*s + 4*ls*tan(delta))*(d*g*sd/2 + c*(1/tan(phi))
     + g*q + BURIED * (d-ls*sin(beta)) * g*sd * (1-sin(phi))/(1+sin(phi)));
 
+  // NOTE: With accurate depths being fed to the equation, the forces are now
+  //       far too large, and immediately deflect the scoop off course. Until we
+  //       find the parameter that's causing this, we have added a fudge factor
+  //       that lowers both forces by a factor of 1/100.
+  constexpr double FUDGE_FACTOR = 1.0; // 0.05
```
anthonykoutroulis, line 168: What is the `0.05` inline comment?

ow_gazebo_plugins/src/BalovnevModelPlugin/BalovnevModelPlugin.cpp
kmdalal: Good catch, though we have been using "cuddled" braces since the start of the project, and "uncuddled" ones as well.   I've added a proposed change to the coding guidelines that we can review next week.

ow_gazebo_plugins/src/BalovnevModelPlugin/BalovnevModelPlugin.cpp
```diff
@@ -213,23 +233,27 @@ void BalovnevModelPlugin::onModDiffVisualMsg(
...
-    }
-  }
-  float depth = sum / (rows * cols);
+  // get max pixel value
+  // NOTE: pixel values should always be negative, so grab the minimum
+  double min_pixel;
+  cv::minMaxLoc(image_handle->image, &min_pixel);
+  m_moving_max_depth->addDatum(static_cast<float>(-min_pixel));
+
+  float depth = m_moving_max_depth->evaluate();
```
anthonykoutroulis, line 242: Are these supposed to be cast/declared float? It looks like all of MovingMaxFilter members are doubles, so this would lose precision.

anthonykoutroulis (approved):
Test worked and produced output as described. I only added a couple minor code questions, but otherwise looks good on my end.

ow_gazebo_plugins/src/BalovnevModelPlugin/BalovnevModelPlugin.cpp
AstroStucky: Right you are. Although the project has not kept to this convention, which might be why it has been purposed to loosen the requirement in our own [internal code style](https://babelfish.arc.nasa.gov/confluence/display/OCEANWATERS/Coding+Style+and+Practices). I'll change this section, since the original code had it on separate lines. I am fine changing other uses where you think it'll improve readability.

AstroStucky: > Good catch, though we have been using "cuddled" braces since the start of the project, and "uncuddled" ones as well. I've added a proposed change to the coding guidelines that we can review next week.

Ah! I did not see this until after my own comment.

ow_gazebo_plugins/src/BalovnevModelPlugin/BalovnevModelPlugin.cpp
AstroStucky: Removed along with the long-winded explanation. FUDGE_FACTOR is enough explanation on its own. I'll keep this variable for now, but hope to remove it by the time the epic branch is merged.

ow_gazebo_plugins/src/BalovnevModelPlugin/BalovnevModelPlugin.cpp
AstroStucky: Good catch! It should definitely just remain double. This may be a hold over from some previous iteration that used floats.

AstroStucky: I will investigate making the `depth` double as well tomorrow. I cannot recall why it's not right now, but it's the end of the day.

AstroStucky:
Thank you for the tests and code comments! I committed some of the suggestions, but plan to commit the rest tomorrow before I merge.

AstroStucky:
Final round of suggestions incorporated. They weren't significant enough to justify retesting. Thanks for the reviews!

PR #279, opened 2022-10-25
Oceanwater 959 camera capture action

mogumbo:
## Linked Issues:
| Jira Ticket 🎟️ | [OCEANWATER-959](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-959) |
| Github :octocat:  | # |

## Summary of Changes
* Added CameraCapture action with client and server.

## Test
* In another terminal:
* cd src/ow_simulator/ow_lander/scripts
* ./antenna_pan_tilt_action_client.py 1 0.5
* ./camera_capture_action_client.py
* ./camera_capture_action_client.py 0.1  (and other exposures)
* view images in rqt

### DIFF
--- a/ow_lander/action/CameraCapture.action
+++ b/ow_lander/action/CameraCapture.action
@@ -1,5 +1,5 @@
 # goal
-float32 exposure
+float64 exposure  # (seconds) Must be greater than 0 or previous exposure will be used
 ---
 # result
 bool point_cloud_created
--- a/ow_lander/action/CameraCapture.action
+++ b/ow_lander/action/CameraCapture.action
@@ -1,0 +1,7 @@
+# goal
+float32 exposure
+---
+# result
+bool point_cloud_created
+---
+# feedback
--- a/ow_lander/action/CameraCapture.action
+++ b/ow_lander/action/CameraCapture.action
@@ -2,6 +2,5 @@
 float64 exposure  # (seconds) Must be greater than 0 or previous exposure will be used
 ---
 # result
-bool point_cloud_created
 ---
 # feedback
--- a/ow_lander/launch/spawn.launch
+++ b/ow_lander/launch/spawn.launch
@@ -100,13 +100,18 @@
     output="screen"/>
   <node pkg="ow_lander" name="AntennaPanTiltAction"
     type="antenna_pan_tilt_action_server.py"
-    launch-prefix="bash -c 'sleep $(arg node_start_delay); $0 $@' "/>
+    launch-prefix="bash -c 'sleep $(arg node_start_delay); $0 $@' "
+    output="screen"/>
   <node pkg="ow_lander" name="DockIngestSampleAction"
     type="dock_ingest_sample_action_server.py"
     launch-prefix="bash -c 'sleep $(arg node_start_delay); $0 $@' "
     output="screen"/>
   <node pkg="ow_lander" name="LightSetIntensity"
     type="light_set_intensity_action_server.py"
+    launch-prefix="bash -c 'sleep $(arg node_start_delay); $0 $@' "
+    output="screen"/>
+  <node pkg="ow_lander" name="CameraCapture"
+    type="camera_capture_action_server.py"
     launch-prefix="bash -c 'sleep $(arg node_start_delay); $0 $@' "
     output="screen"/>
 </launch>
--- a/ow_lander/scripts/camera_capture_action_client.py
+++ b/ow_lander/scripts/camera_capture_action_client.py
@@ -1,0 +1,45 @@
+#!/usr/bin/env python3
+
+# The Notices and Disclaimers for Ocean Worlds Autonomy Testbed for Exploration
+# Research and Simulation can be found in README.md in the root directory of
+# this repository.
+
+import rospy
+import actionlib
+import argparse
+import ow_lander.msg
+from guarded_move_action_client import print_arguments
+
+
+def camera_capture_client():
+  parser = argparse.ArgumentParser(
+    formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+  parser.add_argument(
+    'exposure', type=float, help='Camera exposure in seconds', nargs='?', default=-1, const=0)
+  args = parser.parse_args()
+  print_arguments(args)
+
+  client = actionlib.SimpleActionClient('CameraCapture', ow_lander.msg.CameraCaptureAction)
+  client.wait_for_server()
+
+  goal = ow_lander.msg.CameraCaptureGoal(exposure=args.exposure)
+
+  # Sends the goal to the action server.
+  client.send_goal(goal)
+
+  # Waits for the server to finish performing the action.
+  client.wait_for_result()
+
+  # Prints out the result of executing the action
+  return client.get_result()
+
+
+if __name__ == '__main__':
+  try:
+    # Initializes a rospy node
+    rospy.init_node('camera_capture_client_py')
+    result = camera_capture_client()
+    rospy.loginfo(f'Result: {result}')
+  except rospy.ROSInterruptException:
+    rospy.logerror('program interrupted before completion')
+
--- a/ow_lander/scripts/camera_capture_action_client.py
+++ b/ow_lander/scripts/camera_capture_action_client.py
@@ -12,10 +12,10 @@ from guarded_move_action_client import print_arguments
 
 
 def camera_capture_client():
-  parser = argparse.ArgumentParser(
-    formatter_class=argparse.ArgumentDefaultsHelpFormatter)
-  parser.add_argument(
-    'exposure', type=float, help='Camera exposure in seconds', nargs='?', default=-1, const=0)
+  parser = argparse.ArgumentParser(formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+  parser.add_argument('exposure', type=float,
+                      help='Camera exposure in seconds. Must be > 0 or current exposure will be used.',
+                      nargs='?', default=-1)
   args = parser.parse_args()
   print_arguments(args)
 
--- a/ow_lander/scripts/camera_capture_action_client.py
+++ b/ow_lander/scripts/camera_capture_action_client.py
@@ -30,16 +30,12 @@ def camera_capture_client():
   # Waits for the server to finish performing the action.
   client.wait_for_result()
 
-  # Prints out the result of executing the action
-  return client.get_result()
-
 
 if __name__ == '__main__':
   try:
     # Initializes a rospy node
     rospy.init_node('camera_capture_client_py')
-    result = camera_capture_client()
-    rospy.loginfo(f'Result: {result}')
+    camera_capture_client()
   except rospy.ROSInterruptException:
     rospy.logerror('program interrupted before completion')
 
--- a/ow_lander/scripts/camera_capture_action_server.py
+++ b/ow_lander/scripts/camera_capture_action_server.py
@@ -28,8 +28,6 @@ class CameraCaptureActionServer(object):
     self._server.start()
 
   def execute_cb(self, goal):
-    # helper variables
-    r = rospy.Rate(1)
     self.point_cloud_created = False
 
     # Set exposure if it is specified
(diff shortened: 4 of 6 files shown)


AstroStucky (commented):
Passing a zero or negative exposure results in misleading output. The simulation output says `CameraCapture: Succeeded` and the action client returns the following
```
$ ./camera_capture_action_client.py 0.0
[INFO] [1666735926.833607, 0.000000]: Requested exposure value: 0.0
[INFO] [1666735927.906582, 1916525886.082500]: Result: point_cloud_created: True
```
Additionally the image in RQT does not update, which leads me to think zero/negative exposure values are simply rejected by the camera system, but the user has no way of knowing that is what has happened. 

I made some suggestions relating to this in the code, and I think we should verify the Action's Goal-Result-Feedback format works for JPL.

ow_lander/action/CameraCapture.action
```diff
@@ -0,0 +1,7 @@
+# goal
+float32 exposure
```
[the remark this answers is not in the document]
kmdalal: Yes, we settled on Float64 for all real numbers.

ow_lander/launch/spawn.launch
```diff
@@ -100,7 +100,8 @@
     output="screen"/>
   <node pkg="ow_lander" name="AntennaPanTiltAction"
     type="antenna_pan_tilt_action_server.py"
-    launch-prefix="bash -c 'sleep $(arg node_start_delay); $0 $@' "/>
+    launch-prefix="bash -c 'sleep $(arg node_start_delay); $0 $@' "
+    output="screen"/>
```
kmdalal, line 104: Good addition, it may explain why I wasn't seeing INFO messages in my current antenna work.

ow_lander/scripts/camera_capture_action_server.py
```diff
@@ -0,0 +1,74 @@
...
+
+  def _handle_point_cloud(self, points):
+    """
+    :type points: sensor_msgs.msg.PointCloud2
+    """
+    
+    # CameraCapture was successful if a point cloud is received. However, there
+    # does not appear to be a way to associate this point cloud with the
+    # original trigger message. A trigger could have been sent using without
+    # using the CameraCapture action client.
```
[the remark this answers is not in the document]
kmdalal: Minor: extra "using" in code comment.  @AstroStucky I don't follow your second sentence.

ow_lander/scripts/camera_capture_action_client.py
```diff
@@ -0,0 +1,45 @@
...
+import argparse
+import ow_lander.msg
+from guarded_move_action_client import print_arguments
+
+
+def camera_capture_client():
+  parser = argparse.ArgumentParser(
+    formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+  parser.add_argument(
+    'exposure', type=float, help='Camera exposure in seconds', nargs='?', default=-1, const=0)
```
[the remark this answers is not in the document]
kmdalal: This is the first time I've known that exposure is even settable, and it's interesting to see the result of different values.  But I too don't know what 0 and negative values mean, or what the default is or should be.  This does need some discussion.  Good topic for the JPL tagup tomorrow.

ow_lander/scripts/camera_capture_action_server.py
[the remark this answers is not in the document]
AstroStucky: > Minor: extra "using" in code comment. @AstroStucky I don't follow your second sentence.

We can disregard this comment. I did some digging and what I said does not make sense. I thought the `image_trigger` topic calls a function that we created, which is what handles the function calls that make photography happen, but it looks as though that is in fact set up by a thirdparty ROS Package, and `image_trigger` is just the interface it provides.

ow_lander/scripts/camera_capture_action_server.py
```diff
@@ -0,0 +1,74 @@
...
+from sensor_msgs.msg import PointCloud2
+
+class CameraCaptureActionServer(object):
+  #_result = std_msgs.msg.Bool()
+
+  def __init__(self, name):
+    self._action_name = name
+    self._result = ow_lander.msg.CameraCaptureResult()
+    self._exposure_pub = rospy.Publisher(
+      '/gazebo/plugins/camera_sim/exposure', Float64, queue_size=10)
```
kmdalal, line 22: Just curious: RQT also exposes `/gazebo/plugins/camera_sim/gui/exposure` and `/gazebo/plugins/camera_sim/gui/gamma`.  What are these for, and would they be of interest in this action?

ow_lander/scripts/camera_capture_action_server.py
```diff
@@ -0,0 +1,74 @@
...
+    self._trigger_pub = rospy.Publisher(
+      '/StereoCamera/left/image_trigger', Empty, queue_size=10)
+    self._point_cloud_sub = rospy.Subscriber('/StereoCamera/points2', PointCloud2, self._handle_point_cloud)
```
kmdalal, line 25: This line could be nicely shortened.

ow_lander/scripts/camera_capture_action_server.py
```diff
@@ -0,0 +1,74 @@
...
+import ow_lander.msg
+from std_msgs.msg import Empty, Float64
+from sensor_msgs.msg import PointCloud2
+
+class CameraCaptureActionServer(object):
+  #_result = std_msgs.msg.Bool()
+
+  def __init__(self, name):
+    self._action_name = name
+    self._result = ow_lander.msg.CameraCaptureResult()
```
kmdalal, line 20: Is the result message actually a good initializer, as opposed to None?  Might it be stale data? (I don't know the details of action message management).

ow_lander/scripts/camera_capture_action_server.py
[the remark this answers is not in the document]
kmdalal: I think this is not an important issue.   Actions are the official interface to autonomy; the camera should be operated only though actions.

ow_lander/scripts/camera_capture_action_client.py
```diff
@@ -0,0 +1,45 @@
...
+  args = parser.parse_args()
+  print_arguments(args)
+
+  client = actionlib.SimpleActionClient('CameraCapture', ow_lander.msg.CameraCaptureAction)
+  client.wait_for_server()
+
+  goal = ow_lander.msg.CameraCaptureGoal(exposure=args.exposure)
```
kmdalal, line 25: If we keep exposure, would it make sense to define a valid range?  I found that values over 1 seemed unusable and indistinguishable.  If there's a valid range, it should be checked.

ow_lander/scripts/camera_capture_action_server.py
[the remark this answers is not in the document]
mogumbo: To elaborate on my original comment, it might be possible to set exposure with a service and block until it is complete. This would require significant work on IRGCameraSimPlugin. (ROS2 allows you to set a parameter on a specific node and get a callback after it is set, which would be a good solution if we were using ROS2.)

ow_lander/scripts/camera_capture_action_server.py
mogumbo: That camera doesn't apply here. The "gui" camera is the one used to view the scene in gzclient. It renders frames constantly and is not triggerable.

ow_lander/scripts/camera_capture_action_server.py
mogumbo: I don't understand the question. http://wiki.ros.org/actionlib_tutorials/Tutorials/Writing%20a%20Simple%20Action%20Server%20using%20the%20Execute%20Callback%20%28Python%29#CA-3786eb3a74acabd7ea4ba08b8b23ad9583d5bd79_12 shows _result initialized at the class level, outside the __init__ method. With my limited Python experience, I don't know what the difference is between these two styles.

ow_lander/scripts/camera_capture_action_client.py
mogumbo: Valid range is anything > 0. If you were taking pictures at night, you might want a very high exposure. (Of course, a camera on a real mission would probably have a range set by hardware and software constraints.)

ow_lander/action/CameraCapture.action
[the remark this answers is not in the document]
mogumbo: Added comment

ow_lander/scripts/camera_capture_action_server.py
```diff
@@ -0,0 +1,74 @@
...
+import rospy
+
+import actionlib
+
+import ow_lander.msg
+from std_msgs.msg import Empty, Float64
+from sensor_msgs.msg import PointCloud2
+
+class CameraCaptureActionServer(object):
+  #_result = std_msgs.msg.Bool()
```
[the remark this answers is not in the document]
mogumbo: done

ow_lander/action/CameraCapture.action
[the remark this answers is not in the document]
mogumbo: Changed to Float64

ow_lander/scripts/camera_capture_action_server.py
mogumbo: done

ow_lander/scripts/camera_capture_action_server.py
AstroStucky: Class level defined variables are static data members, where as those defined in __init__ or any other member method are non-static data members specific to the instance of the class (`self`). 

Given how there are never more than 1 instance of any of our action servers, the difference between one or the other is meaningless.

ow_lander/scripts/camera_capture_action_server.py
```diff
@@ -0,0 +1,71 @@
...
+    # Set exposure if it is specified
+    if goal.exposure > 0:
+      self._exposure_pub.publish(goal.exposure)
+      # There is no guarantee that exposure is set before the image is triggered.
+      # Pause to make it highly likely that exposure is received first.
+      r.sleep()
+
+    self._trigger_pub.publish()
+
+    for i in range(0, 5):
```
kmdalal, line 44: Minor/optional: could use a named timeout constant instead of '5' .

kmdalal (approved):
I retested, and add only one more minor code comment.  Looks good!

ow_lander/scripts/camera_capture_action_server.py
kmdalal: I don't know Python or ROS well enough to make a substantial claim here, so you can ignore my first comment :)

mogumbo: Actually, that was bugging me a little while making this: How should an action server handle multiple simultaneous requests? I didn't see any good examples of it, but that's probably a problem for another day.

AstroStucky:
Just re-tested. Why do I see a reddish area in high exposure images? 

**At 0.4 s exposure**
exposure_0 4

**At 0.8 s exposure**
exposure_0 8

mogumbo:
> Just re-tested. Why do I see a reddish area in high exposure images?

No idea. I don't think it has anything to do with this ticket, so we should make a separate ticket to investigate.

ow_lander/scripts/camera_capture_action_server.py
mogumbo: Done. Also running at 10 Hz instead of 1 Hz now.

AstroStucky (approved):
Simulation output still says "succeeded" when zero/negative exposure is not used. Since a point cloud is apparently still being created, ableit likely a useless one, I am torn on whether zero/negative exposure should make the server abort or not. Everything else looks great, so I'll approve this. I just wanted to note that.

mogumbo:
> Simulation output still says "succeeded" when zero/negative exposure is not used. Since a point cloud is apparently still being created, ableit likely a useless one, I am torn on whether zero/negative exposure should make the server abort or not. Everything else looks great, so I'll approve this. I just wanted to note that.

The help message states: "Must be > 0 or current exposure will be used." If you use any value <= 0 you still get a valid point cloud using the exposure that is currently set on the camera. This is a successful result.

PR #287, opened 2022-11-11
Revised a few logging calls in pan/tilt action server.

kmdalal:
No Jira.  This is a simple change to a few ROS log calls in the antenna pan/tilt action server.  It quiets the main message while making some others louder if they occur.

No test is needed - it's mainly a matter of agreeing with these preferences.


kmdalal:
Closing this PR with these decisions made today in DevDay:

- The logging levels should be consistent in all action servers.
- Logging levels will be handled in the action server refactoring that @AstroStucky is working on.
- The log level for preempted and success should be INFO, and ERROR for failure.

PR #288, opened 2022-11-15
Oceanwater 1039 Regolith less likely to be flung out of sample dock

AstroStucky:
## Linked Issues:
| EPIC ⚡ | n/a |
| Jira Ticket 🎟️ | [OCEANWATER-1039](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1039) |

This change reduces the likelihood, but I suspect does not entirely solve the problem because there was one instance during testing when I saw the scoop rapidly change orientation for only a single frame when no sample was in it. It's not clear why it does this since it is on a fixed mount attached to the lander. 

## Summary of Changes
* Sample dock's implementation in the lander.xacro simplified
* `expectedFailure` decorator removed from some unit tests in **sample collection**
* Greatly reduced likelihood of sample getting flung out of sample dock by
  * Switching the collision model of regolith to that of a cube
  * Switching the collision model of the sample dock from a mesh to a collection of boxes
_These two changes together were required to reduce sample flinging. They were also tested individually, but those tests did not reduce the likelihood of sample being flung._

## Test
Run the **sample collection** test a minimum of 3 times. 
```
rostest ow_sim_tests sample_collection.test
``` 
It is fine to increase speed during these tests in the GUI or from a second terminal
```
gz physics -u 1200
```
Any failures you see should only arise from the deliver step, when all of the regolith fails to fall into the dock, which may cause either **Test 7** or **Test 10** to fail. This problem is out of scope for this PR.

## Caveat
The box collision model allows the visual sphere mesh of regolith to poke through the bottom of the scoop.
default_gzclient_camera(1)-2022-11-14T17_23_26 717565

### DIFF
--- a/ow_lander/urdf/lander.xacro
+++ b/ow_lander/urdf/lander.xacro
@@ -21,7 +21,7 @@
 
   <xacro:include filename="$(find ow_lander)/urdf/lander_sample_dock.xacro" />
   <xacro:lander_sample_dock lander_sample_dock_parent_link="base_link"
-                     off_x="0.5226" off_y="-0.35" off_z="0.32" off_R="-${pi/2.0}" off_P="0.0" off_Y="-${pi/2.0}"
+                     off_x="0.5476" off_y="-0.35" off_z="0.32" off_R="-${pi/2.0}" off_P="0.0" off_Y="-${pi/2.0}"
                      vis_mask="${vis_mask}" />
 
   <xacro:arg name="freeze_base_link" default="false"/>
--- a/ow_lander/urdf/lander_sample_dock.xacro
+++ b/ow_lander/urdf/lander_sample_dock.xacro
@@ -30,14 +30,15 @@
 
   <xacro:macro name="lander_sample_dock_link" params="suffix roll pitch y">
     <link name="lander_sample_dock${suffix}">
+
       <collision>
-        <origin rpy="0 0 0" xyz="0 0 0.025"/>
+        <origin rpy="0 0 0" xyz="0 0 0"/>
         <geometry>
           <mesh filename="package://ow_lander/meshes/sample_dock.stl" />
         </geometry>
       </collision>
       <visual>
-        <origin rpy="0 0 0" xyz="0 0 0.025"/>
+        <origin rpy="0 0 0" xyz="0 0 0.0"/>
         <geometry>
           <mesh filename="package://ow_lander/meshes/sample_dock.stl" />
         </geometry>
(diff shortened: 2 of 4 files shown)


kmdalal (approved):
I ran the rostest 5 times:

1. Unstow and DigLinear failed out of tolerance, and the regolith particles "popped" out of the bin many seconds after they had settled.
2. Unstow and Deliver failed -- I emailed you a full transcript.
3. Success
4. Success
5. Deliver failed - I emailed you a transcript of the failure.

Assuming that the failures above are either expected occasionally or out of scope, and that the intended goal is accomplished.

AstroStucky:
@kmdalal Was there only one of those 4 tests that saw regolith pop out of the sample dock?

In the end that is really what this PR seeks to reduce the occurrence of. The other things ought to be fixed, but not in this PR.

kmdalal:
> @kmdalal Was there only one of those 4 tests that saw regolith pop out of the sample dock?
> 
> In the end that is really what this PR seeks to reduce the occurrence of. The other things ought to be fixed, but not in this PR.

I only saw it in test 1.  Of the rest, there was just one (I can't remember which) that I didn't watch carefully.

AstroStucky:
> I only saw it in test 1. Of the rest, there was just one (I can't remember which) that I didn't watch carefully.

It's not ideal, but I think it is acceptable. I just ran the test 5 or 6 more times and I still have yet see regolith flung out of the sample dock. That said, I eagerly await @mogumbo's test.

mogumbo (changes requested):
You can aggravate the problem by multiplying the pixel_area on line 223 of RegolithSpawner.cpp by 5. This results in more regolith spheres being created.

Though the problem may have decreased a bit with these changes, it isn't really solved. I think we should decline this PR and look for other solutions.

I think the underlying problem comes from one sphere getting wedged between a sample collector wall and another sphere. Sometimes this results in the physics being solved violently. Perhaps we can modify physics parameters for the regolith spheres and maybe the sample collector so collisions are resolved less aggressively.

AstroStucky:
> You can aggravate the problem by multiplying the pixel_area on line 223 of RegolithSpawner.cpp by 5. This results in more regolith spheres being created.

Great find! You can also do the same thing without having to recompile by modifying the `spawn_volume_threshold` on line 93 in **ow/launch/common.launch**. The relation for this value is indirect, so to get 5x the number of regolith particles you would divide the existing value by 5. 

> Though the problem may have decreased a bit with these changes, it isn't really solved. I think we should decline this PR and look for other solutions.

I am all for this suggestion. I would like to solve the problem at its source.

PR #292, opened 2022-11-30
Remove obsolete light transform bug workaround

mogumbo:
## Linked Issues:
| Jira Ticket 🎟️ | [OCEANWATER-1089](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1089) |

## Test
* Launch OceanWATERS without this branch and observe that the wireframe light frusta are not connected to the light model cylinders on the antenna.
* Check out this branch and observe that the light frusta are no in the right place.
* Tilt the light and see that the light beams are projected in the right place: `rostopic pub -1 /ant_tilt_position_controller/command std_msgs/Float64 1.4`

### DIFF
--- a/ow_lander/urdf/lander_lights.xacro
+++ b/ow_lander/urdf/lander_lights.xacro
@@ -73,8 +73,6 @@
         </plugin>
       </visual>
       <light name="lander_light_light${suffix}" type="spot">
-        <!-- Light is not transformed by the joint above, so we transform it here. -->
-        <pose>0 ${y} 0 ${roll} ${pitch} 0</pose>
         <diffuse>1 1 1 1</diffuse>
         <attenuation>
           <range>100</range>


anthonykoutroulis:
Recapping a discussion that @mogumbo and I had while I was testing this:

The behavior described in the PR (the previous bug and the update to remove the deprecated fix) produces the expected results in the latest Gazebo (ver: 11.12.0). However, removing the previous hotfix re-introduces the lighting bug in Gazebo 11.11.0.

Since we currently require Gazebo v9+, we probably hold off on this until we discuss updating our required version.

AstroStucky:
> Recapping a discussion that @mogumbo and I had while I was testing this:
> 
> The behavior described in the PR (the previous bug and the update to remove the deprecated fix) produces the expected results in the latest Gazebo (ver: 11.12.0). However, removing the previous hotfix re-introduces the lighting bug in Gazebo 11.11.0.
> 
> Since we currently require Gazebo v9+, we probably hold off on this until we discuss updating our required version.

Looks like I'm running 11.12.0. I have been wondering why our light placement looked off to me 🤔 

Should I go ahead and a find a way to downgrade?

AstroStucky (approved):
I am on Gazebo 11.12.0 and this fixes the bug for me, so it gets an approval. 

Whether we want to merge it or not is another discussion (one I see no reason we can't have here).

anthonykoutroulis:
Okay this looks good to merge.

PR #295, opened 2022-12-13
Oceanwater 1016 arm joint accelerations

mogumbo:
## Linked Issues:
| EPIC ⚡ | [OCEANWATER-934](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-934) |
| Jira Ticket 🎟️ | [OCEANWATER-1016](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1016) |

## Summary of Changes
* Publish arm joint accelerations from state_publisher.py.

## Test
* rostopic echo /arm_joint_accelerations

### DIFF
--- a/ow_lander/scripts/state_publisher.py
+++ b/ow_lander/scripts/state_publisher.py
@@ -6,11 +6,12 @@
 
 import rospy
 from sensor_msgs.msg import JointState
-from owl_msgs.msg import ArmJointPositions, ArmJointTorques, ArmJointVelocities, PanTiltPosition
+from owl_msgs.msg import ArmJointAccelerations, ArmJointPositions, ArmJointTorques, ArmJointVelocities, PanTiltPosition
 
 class StatePublisher(object):
 
   def __init__(self, name):
+    self._joint_acc = ArmJointAccelerations()
     self._joint_pos = ArmJointPositions()
     self._joint_tor = ArmJointTorques()
     self._joint_vel = ArmJointVelocities()
@@ -18,6 +19,8 @@ class StatePublisher(object):
 
     self._subscriber = rospy.Subscriber(
             '/joint_states', JointState, self._handle_joint_states)
+    self._arm_joint_accelerations_pub = rospy.Publisher(
+            '/arm_joint_accelerations', ArmJointAccelerations, queue_size=10)
     self._arm_joint_positions_pub = rospy.Publisher(
             '/arm_joint_positions', ArmJointPositions, queue_size=10)
     self._arm_joint_torques_pub = rospy.Publisher(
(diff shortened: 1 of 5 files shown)


ow_lander/scripts/state_publisher.py
```diff
@@ -36,6 +39,15 @@ def _handle_joint_states(self, data):
       rospy.logerr_throttle(1, 'StatePublisher: Array size mismatch')
       return
 
+    dur = data.header.stamp - self._joint_acc.header.stamp
+    dur_float = dur.to_sec() + dur.to_nsec() / 1000000000
```
AstroStucky, line 43: I think you can get away with just `to_sec`
```suggestion
    dur_float = dur.to_sec()
```
_Source_ http://wiki.ros.org/rospy/Overview/Time#Converting_Time_and_Duration_instances

ow_lander/scripts/state_publisher.py
```diff
@@ -36,6 +39,15 @@ def _handle_joint_states(self, data):
...
+    # Only publish accelerations if time is moving forward (this should always happen)
+    if dur_float > 0 :
+      self._joint_acc.header = data.header
+      for i in range(0, len(self._joint_acc.value)):
+        self._joint_acc.value[i] = (data.velocity[i+2] - self._joint_vel.value[i]) / dur_float
```
AstroStucky, line 48: Is the `+2` in the index to skip the first two joints, which I believe are antenna pan and tilt? I see 2's elsewhere in the file, so can we make it a constant?

AstroStucky (approved):
Test worked fine for me, and it publishes at the 50 Hz I would expect. Values looked moved away from zero when I performed an unstow, as expected. I did not look too closely too check their accuracy. 

My code suggestions can be considered optional.

ow_lander/scripts/state_publisher.py
mogumbo: Thanks. You found my bug. I saw other examples that made me think I had to sum these two values. Now durations are 0.02 as expected.

ow_lander/scripts/state_publisher.py
mogumbo: Changed to `OFFSET`

ow_lander/scripts/state_publisher.py
mogumbo: Fixed

AstroStucky (approved):

kmdalal (approved):
The test passes, and I also tried it with the arm in motion and the result seemed fairly intuitive.  I have no comments on the code, which looks good after the fixes.

PR #298, opened 2022-12-20
Oceanwater 1021 implement arm pose

mogumbo:
## Linked Issues:
| EPIC ⚡ | [OCEANWATER-934](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-934) |
| Jira Ticket 🎟️ | [OCEANWATER-1021](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1021) |

## Summary of Changes
* Added /arm_pose topic

## Test
* rostopic echo /arm_pose
* ./guarded_move_action_client.py

Note that the final position of guarded_move is similar to the position of /arm_pose because guarded_move shows the position of l_scoop_tip while /arm_pose shows the position of l_scoop. Both positions are relative to base_link.

### DIFF
--- a/ow_lander/scripts/state_publisher.py
+++ b/ow_lander/scripts/state_publisher.py
@@ -5,8 +5,10 @@
 # this repository.
 
 import rospy
+import tf2_ros
 from sensor_msgs.msg import JointState
-from owl_msgs.msg import ArmJointAccelerations, ArmJointPositions, ArmJointTorques, ArmJointVelocities, PanTiltPosition
+from owl_msgs.msg import ArmJointAccelerations, ArmJointPositions, ArmJointTorques, \
+                         ArmJointVelocities, ArmPose, PanTiltPosition
 
 class StatePublisher(object):
 
@@ -16,9 +18,11 @@ class StatePublisher(object):
     self._joint_tor = ArmJointTorques()
     self._joint_vel = ArmJointVelocities()
     self._pan_tilt = PanTiltPosition()
+    self._arm_pose = ArmPose()
 
     self._subscriber = rospy.Subscriber(
             '/joint_states', JointState, self._handle_joint_states)
+
     self._arm_joint_accelerations_pub = rospy.Publisher(
             '/arm_joint_accelerations', ArmJointAccelerations, queue_size=10)
     self._arm_joint_positions_pub = rospy.Publisher(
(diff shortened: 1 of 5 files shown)


kmdalal:
The given test passes, though when I tried an Unstow, its final X position was about .74 and the /arm_pose X position was about 1.74.   I'd like to understand why -- not sure how to glean it from the Unstow server code.

mogumbo:
> The given test passes, though when I tried an Unstow, its final X position was about .74 and the /arm_pose X position was about 1.74. I'd like to understand why -- not sure how to glean it from the Unstow server code.

The final unstow position is coming from `LinkStateSubscriber` in `LanderInterface.py`. `LinkStateSubscriber` is poorly named because all it is doing is grabbing the position of "lander::l_scoop_tip" from `/gazebo/link_states`. That topic's positions are relative to the gazebo world frame, while guarded move and `/arm_pose` show positions relative to "base_link." We should do a separate ticket to make all these things consistent and rename `LinkStateSubscriber` to something reasonable.

AstroStucky:
> `LinkStateSubscriber` is poorly named because all it is doing is grabbing the position of "lander::l_scoop_tip" from `/gazebo/link_states`. 

In the refactor branch it performs the same function, but is called `LinkPositionSubscriber` and the name of the link it queries position on is parameter.

kmdalal (approved):

ow_lander/scripts/state_publisher.py
```diff
@@ -67,6 +78,24 @@ def _handle_joint_states(self, data):
...
+    # Use /joint_states as the clock for publishing /arm_pose. Alternatively,
+    # this could be refactored put publish /arm_pose at a different rate.
+    self._publish_arm_pose()
+
+
+  def _publish_arm_pose(self):
+    try:
+      trans = self._tf_buffer.lookup_transform('base_link', 'l_scoop',  rospy.Time(0))
+    except (tf2_ros.LookupException, tf2_ros.ConnectivityException, tf2_ros.ExtrapolationException) as e:
+      rospy.logwarn("StatePublisher: tf2 raised an exception: %s" % e)
```
AstroStucky, line 90: I'd prefer this logged as an error

owl_msgs/msg/ArmPose.msg
```diff
@@ -0,0 +1,2 @@
+Header header
```
AstroStucky, line 1: Do we require the NASA license header in these files? Will owl_msgs be under the NASA license?

AstroStucky (approved):
Test works great. Only a few minor code comments.

owl_msgs/msg/ArmPose.msg
mogumbo: We haven't been putting the header in our other message files. I don't know how important it is, but it should be a separate issue if we decide to do it.

ow_lander/scripts/state_publisher.py
mogumbo: done

owl_msgs/msg/ArmPose.msg
kmdalal: NASA license info, no.   We've never put those in message files.  

The ROS Header, good question.  We're not consistent here.  I like the brevity of messages without the header, but I wonder if they are useful enough to add.  Could discuss in Dev Day.

PR #301, opened 2023-01-04
Oceanwater-944 Arm Move Cartesian

AstroStucky:
# Linked Issues:
| Jira Ticket 🎟️ | [OCEANWATER-944](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-944) |

# Summary of Changes
* Added ArmMoveCartesian
* Minor improvements to **mixin.py** and **trajectory_planer.py**

# Test
Launch the simulator in any world and perform the following rosrun calls. Your results should match the image, and the simulator output should show
```
ArmMoveCartesian action started
ArmMoveCartesian: Succeded - ArmMoveCartesian trajectory succeeded
ArmMoveCartesian action complete
```
for each command.

## Test 1 - Help Message
The output of `rosrun ow_lander arm_move_cartesian.py --help` should be
```
usage: arm_move_cartesian.py [-h] [--frame {0,1}] [--relative] [-x X] [-y Y] [-z Z]
                             [--euler X Y Z | --quaternion X Y Z W]

Move end-effector by Cartesian coordinates/translation.

optional arguments:
  -h, --help            show this help message and exit
  --frame {0,1}, -f {0,1}
                        The frame index corresponding to the frame that orientation and position are relative
                        to. {0: 'base_link', 1: 'l_scoop_tip'} (default: 0)
  --relative, -r        If True, pose will be interpreted as relative to the current pose. (default: False)
  -x X                  X-coordinate or the change in the X-coordinate (default: 0)
  -y Y                  Y-coordinate or the change in the Y-coordinate (default: 0)
  -z Z                  Z-coordinate or the change in the Z-coordinate (default: 0)
  --euler X Y Z, -e X Y Z
                        Orientation in Euler angle form. Takes 3 radian values. (default: [0, 0, 0])
  --quaternion X Y Z W, -q X Y Z W
                        Orientation in quaternion form. Takes 4 scalars, the first 3 are the imaginary parts,
                        and the 4th is the real part. (default: [0, 0, 0, 1])
```
You'll noticed I made an attempt to improve our approach to action client arguments. You can now provided only the arguments you wish to provide and leave the rest as their defaults.

## Test 2 - Absolute Poses
1. `rosrun ow_lander arm_move_cartesian.py -x 1.7 -y -0.2 -z 0.6`
Scoop will be upside down facing away from the lander
default_gzclient_camera(1)-2023-01-04T13_57_42 685770

2. `rosrun ow_lander arm_move_cartesian.py -x 1.7 -y -0.2 -z 0.6 -q 1 0 0 0`
Now right side up facing away from the lander
default_gzclient_camera(1)-2023-01-04T13_58_01 566416

3. `rosrun ow_lander arm_move_cartesian.py -x 1.7 -y -0.2 -z 0.6 -q 0 1 0 0`
Right side up facing toward the lander
default_gzclient_camera(1)-2023-01-04T13_58_20 717144

4. `rosrun ow_lander arm_move_cartesian.py -x 1.7 -y -0.2 -z 0.6 -q 0 0 1 0`
Upside down facing toward the lander
default_gzclient_camera(1)-2023-01-04T14_06_25 519300

5. `rosrun ow_lander arm_move_cartesian.py -x 1.7 -y -0.2 -z 0.6 -q 0 0 0 1`
Scoop should now be in the same orientation as **Step 1**, because the default orientation argument is the unit quaternion that we just provided.

6. `rosrun ow_lander arm_move_cartesian.py -x 1.7 -y -0.2 -z 0.6 -q 0.7071068 0.7071068 0 0`
Scoop will face perpendicular to the arm and level to the terrain.
default_gzclient_camera(1)-2023-01-04T14_20_03 182814

## Test 3 - Relative Poses
Relative poses can be achieved using either `--frame 1` or `--relative`. They both communicate the same thing to the action, which is to acquire the provided pose relative to the tool's frame. Do the following command using either of these options or both. If you do the following commands from the position in **Step 2.6** (above) then the following images should match.

1. `rosrun ow_lander arm_move_cartesian.py -r -x 0.2`
The scoop should move in the scooping direction by 20cm.
default_gzclient_camera(1)-2023-01-04T14_36_02 728442

2. `rosrun ow_lander arm_move_cartesian.py -r -z 0.2`
The scoop should move in the direction of its bottom by 20cm.
default_gzclient_camera(1)-2023-01-04T14_36_44 297096

3. `rosrun ow_lander arm_move_cartesian.py -r -y 0.2`
The scoop should move perpendicular to the two previous direction by 20cm. 
default_gzclient_camera(1)-2023-01-04T14_36_59 554611

4. `rosrun ow_lander arm_move_cartesian.py -r -q .5 .5 .5 .5`
The scoop should now be sideways and facing away from the lander.
default_gzclient_camera(1)-2023-01-04T14_37_24 380014

## Test 4 - Freestyle
Try out any combination of commands. Normalized quaternions are required, or it will either fail planning or use the identity quaternion instead of the one you provided.

These two links can help think in terms of quaternions
https://quaternions.online/ -- This is the only visualization I was able to find, and unfortunately some quaternions will not be printed with enough significant figures to be normalized. 
https://www.andre-gaschler.com/rotationconverter/ -- Alternatively, this will calculate your quaternion from Euler angles without visuals and will provide significant figures to be normalized.

## Test 5 - Euler Angle input
Do a few relative and non-relative commands with Euler angles. 

1. Command an absolute pose
`rosrun ow_lander arm_move_cartesian.py -x 1.7 -y -0.2 -z 0.6 -e 3.14 0 -1.5`
Scoop will face to lander's right.
default_gzclient_camera(1)-2023-01-06T16_24_21 846729

2. Spin the scoop 180 deg along its yaw joint (it will also changes position slightly to keep the tip at the same position in space).
`rosrun ow_lander arm_move_cartesian.py -r -e 0 0 3.14`
default_gzclient_camera(1)-2023-01-06T16_28_09 790409

3. Move downward and spin scoop so that it faces the ground.
`rosrun ow_lander arm_move_cartesian.py -r -z 0.5 -e 0 -1.57 1.57`
default_gzclient_camera(1)-2023-01-06T16_33_54 260810

## Test 6 -- Regression Test
Do all of **Test 5** prior to the following steps. Use the same simulation instance. 
1. rosrun ow_lander guarded_move_action_client.py
2. rosrun ow_lander grind_action_client.py
3. rosrun ow_lander dig_linear_action_client.py
4. rosrun ow_lander discard_action_client.py
5. rosrun ow_lander dig_circular_action_client.py
6. rosrun ow_lander deliver_action_client.py
7. Rerun **Test 5** in this same simulation instance.


ow_lander/src/ow_lander/actions.py
```diff
@@ -189,6 +197,86 @@ def plan_trajectory(self, _goal):
...
+  action_type   = owl_msgs.msg.ArmMoveCartesianAction
+  goal_type     = owl_msgs.msg.ArmMoveCartesianGoal
+  feedback_type = owl_msgs.msg.ArmMoveCartesianFeedback
+  result_type   = owl_msgs.msg.ArmMoveCartesianResult
+
+  def publish_feedback_cb(self):
+    self._publish_feedback(pose=self._arm_tip_monitor.get_link_pose())
+
+  def execute_action(self, goal):
+    ARM_END_EFFECTOR = 'l_scoop_tip'
```
mogumbo, line 212: The end effector is really the scoop and the scoop tip is just a point on the scoop.  It there a good reason to use one over the other?

ow_lander/src/ow_lander/transforms.py
```diff
@@ -0,0 +1,51 @@
...
+import tf2_ros
+
+from ow_lander.common import Singleton
+
+# NOTE: These imports are not directly used, but importing them works around a
+#       bug discussed here
+# https://answers.ros.org/question/249433/tf2_ros-buffer-transform-pointstamped/
+from tf2_geometry_msgs import *
+
+class FrameTransformer(metaclass = Singleton):
```
mogumbo, line 15: Since this is the only class in the file I think the file should be named FrameTransformer.py. That's the convention we use for C++, but I don't see a similar convention written down in our python style guide. Do we have rules for naming python scripts?

ow_lander/src/ow_lander/actions.py
AstroStucky: If `l_scoop` was used then the origin of the scoop would be the point that gets posed, which I think is set to its center-of-mass. I don't do a ton of work with robotic arms, but I feel like it makes the most sense to instead pose the tip or the furthest out point of the end-effector. That is certainly what you would want to use for ArmFindSurface, otherwise the surface you find would appear to be inside the scoop.

ow_lander/src/ow_lander/transforms.py
AstroStucky: Neither our own or ROS's Python style guide outline guidelines for module file names, so I think it defer's to Python's [PEP8 style guide](https://peps.python.org/pep-0008/#naming-conventions).

kmdalal:
In the help message, for `frame`, it may be helpful to mention tool/base.

ow_lander/src/ow_lander/trajectory_planner.py
```diff
@@ -139,7 +169,10 @@ def go_to_XYZ_coordinate(self, cs, goal_pose, x_start, y_start, z_start, approxi
         """
         :param approximate: use an approximate solution. default True
         :type move_group: class 'moveit_commander.move_group.MoveGroupCommander'
-        :type x_start: float
+        :type x_start: floa
```
kmdalal, line 172: typo: floa

kmdalal (approved):
Nice set of tests and descriptions -- they all passed.  

I tried some additional ones, and saw what I'd expect.  I like the errors/warnings when I requested an invalid position.  I didn't observe any very strange arm movements.

I agree that quaternions are hard and I didn't try to learn them.  If Euler angles can be easily substituted, this seems a good idea; else it can be future work.

My code review was light, since I'm not strong in Python, and not good for reviewing the math.  My comments were minor.

Nice work!

ow_lander/src/ow_lander/transforms.py
AstroStucky: The name of the module is now `frame_transformer.py`

ow_lander/src/ow_lander/trajectory_planner.py
AstroStucky: fixed

AstroStucky:
> In the help message, for `frame`, it may be helpful to mention tool/base.

Done

AstroStucky:
The action client now accepts Euler angles as a substitute for quaternions. Providing both arguments will result in an error. Euler angles are read as radians.

AstroStucky:
So long as we're improving user-friendliness, I do think it would be cool to use a flag parameter that causes angle inputs to be parsed as degrees. But I don't want to expand the scope of this PR any further, so I will consider this for future work.

kmdalal:
> So long as we're improving user-friendliness, I do think it would be cool to use a flag parameter that causes angle inputs to be parsed as degrees. But I don't want to expand the scope of this PR any further, so I will consider this for future work.

I agree, not now.

kmdalal:
Please update the test instructions, and in the help message, add the one-letter options for Euler and quaternion arguments.

AstroStucky:
> Please update the test instructions, and in the help message, add the one-letter options for Euler and quaternion arguments.

Done. 

Reviewers, I think it is sufficient to just complete the newly added Test 5.

kmdalal:
After running the new Test 5 (which worked), I tried a default GuardedMove, which failed (see error below).  I did an Unstow, and then the default GuardedMove again, and it failed the same way.

Could Cartesian moves to arbitrary positions break the path planning in subsequent existing operations?  It may be essential to test this.

```
[INFO]: GuardedMove action started
[ WARN]: More than 80% of the sampled goal states fail to satisfy the constraints imposed on the goal sampler. Is the constrained sampler working correctly?
[ WARN]: More than one constraint is set. If your move_group does not have multiple end effectors/arms, this is unusual. Are you using a move_group_interface and forgetting to call clearPoseTargets() or equivalent?
[ERROR]: GuardedMove: Aborted - Trajectory planning failed
[INFO]: GuardedMove action complete
```

ow_lander/src/ow_lander/transforms.py
mogumbo: Thanks. This is one thing I think C++ style does a lot better than Python, but I'll go with it for now.

AstroStucky:
Good find, @kmdalal. I have replicated the error, and I'm investigating the cause. So far it appears to only affect GuardedMove.

AstroStucky:
So it turns out a call to `set_pose_target` should generally be followed by either `clear_pose_target` or `clear_pose_targets`, otherwise other plans may fail. The problem actually touched multiple arm actions, namely: GuardedMove, Grind, DigLinear, DigCircular, Deliver, and Discard.  I believe the refactor is partially responsible for the problem because it reuses the same move group instances, `_move_arm` and `_move_grinder`. I plan to explore whether this is the correct approach in a future refactor of ArmTrajectoryPlanner, but hopefully this is a suitable work around for now.

Please perform **Test 5**, then in the same simulation run the above mentioned actions in the following order
1. `rosrun ow_lander guarded_move_action_client.py`
2. `rosrun ow_lander grind_action_client.py`
3. `rosrun ow_lander dig_linear_action_client.py`
4. `rosrun ow_lander discard_action_client.py`
5. `rosrun ow_lander dig_circular_action_client.py`
6. `rosrun ow_lander deliver_action_client.py`

Then redo **Test 5** in the same simulation. If that performs nominally and you do not see the error mentioned by Mike, then this PR should be good.

PR #303, opened 2023-01-08
Oceanwater 952 Dummy commands

kmdalal:
## Linked Issues:
| EPIC ⚡ | [OCEANWATER-952](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-952) |
| Jira Ticket 🎟️ | [OCEANWATER-952](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-952) |
| Github :octocat:  | #303 |

**This is a work in progress.**

### Summary of Changes
* New dummy (no-op) commands to unify with OWLAT:
  * ArmSetTool

### Test
*


kmdalal:
The client script `arm_set_tool.py` just hangs...

ow_lander/src/ow_lander/actions.py
```diff
@@ -207,6 +208,18 @@ def modify_joint_positions(self, goal):
...
+  name          = 'ArmSetTool'
+  action_type   = owl_msgs.msg.ArmSetToolAction
+  goal_type     = owl_msgs.msg.ArmSetToolGoal
+  feedback_type = owl_msgs.msg.ArmSetToolFeedback
+  result_type   = owl_msgs.msg.ArmSetToolResult
+
+  def execute_action(self, goal):
+    msg = f"{self.name} is not yet supported in OceanWATERS."
+    rospy.logwarn(msg)
+    self._set_succeeded(msg, success=True, message=msg)
```
AstroStucky, line 222: The keyword args for this function should correspond to the action's Result parameters. The action Result does not have success or message, it has previous_tool.

The positional argument, which `msg` is passed to, is good. This is just whatever is logged by ActionServer when the action completes and will also be put into the Result's text parameter (this is a parameter that is internal to actionlib and exists for all actions Results).

AstroStucky:
I am not sure why the action client hangs. It could be that wrong Result parameters were being passed, but that should also cause an error message. 

I will investigate more later. For now try correcting the **_set_succeded** call and see if that helps.

ow_lander/src/ow_lander/actions.py
kmdalal: Thanks for pointing this out -- I wasn't familiar with the parameter scheme and just copied another call without seeing they can differ.

This did not solve the client script hang, which stumps me...

ow_lander/src/ow_lander/actions.py
```diff
@@ -207,6 +208,18 @@ def modify_joint_positions(self, goal):
...
     return pos
 
+class ArmSetToolServer(ActionServerBase):
+
+  name          = 'ArmSetTool'
+  action_type   = owl_msgs.msg.ArmSetToolAction
+  goal_type     = owl_msgs.msg.ArmSetToolGoal
+  feedback_type = owl_msgs.msg.ArmSetToolFeedback
+  result_type   = owl_msgs.msg.ArmSetToolResult
+
```
AstroStucky, line 218: The action client blocks and does nothing because the action server is never started using `_start_server`. Since this is a dummy action, and does not use the arm interface, it does make sense that it does not inherit from `ArmActionMixin`, which would only needlessly initialize an arm and planning interface, but you still require a call to `_start_server` in this class's `__init__`. See other non-arm actions such as `LightSetIntensity` for an example. 

```suggestion
class ArmSetToolServer(ActionServerBase):

  name          = 'ArmSetTool'
  action_type   = owl_msgs.msg.ArmSetToolAction
  goal_type     = owl_msgs.msg.ArmSetToolGoal
  feedback_type = owl_msgs.msg.ArmSetToolFeedback
  result_type   = owl_msgs.msg.ArmSetToolResult
  
  def __init__(self, *args, **kwargs):
    super(ArmSetToolServer, self).__init__()
    self._start_server()

```

kmdalal: Thanks, that was it!  I'll let you know when this PR is ready for review - or when I have another implementation question.

kmdalal:
Per group discussion today, decided against this for Release 11.

PR #308, opened 2023-01-19
Oceanwater 1015 arm end effector force torque telemetry

anthonykoutroulis:
## Linked Issues:
| EPIC ⚡ | [OCEANWATER-1015](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1015) |
| Jira Ticket 🎟️ | [OCEANWATER-1015](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1015) |
| Github :octocat:  | # |

## Summary of Changes
Update Force/Torque sensor telemetry to unified standard
* update topic `/ft_sensor_dist_pitch` topic to `/arm_end_effector_force_torque`
* add `ArmEndEffectorForceTorque.msg` to the owl_msgs package with updated semantics
* update code to use float array of new msg semantics over previous `geometry_msg/WrenchStamped` msg
* update fault injection to map gazebo sensor topic to new topic name
* change variable names to be consistent with new naming scheme

## Test
*Ensure the ow_autonomy repo is on the [934 epic branch](https://github.com/nasa/ow_autonomy/tree/OCEANWATER-934_telemetry_unification_epic)*

Test that the sensor information is published on the correct topic
* Open two sourced terminals, in one launch the sim `roslaunch ow atacamy_y1a.launch`
* In the second terminal, monitor the newly renamed topic `rostopic echo /arm_end_effector_force_torque`
* Observe the force torque values displaying rapidly

Test that fault injection redirects the topic and behaves correctly
* Add the FT sensor faults for zero_signal_failure and enable. Observe the values zero out.
* Uncheck the zero_signal_failure fault and move the signal_bias slider to the right and observe the values changing by ~1.0
* Reset the signal_bias slider and now move the signal_noise_slider to the right and observe an increase in noise in the values.

~~Test the modified ros_test~~
* ~~run `rostest ow_sim_tests ft_sensor.test`~~
* ~~the test should pass, but there is a failure on closure that is present in noetic-devel~~

*Note: The previous test is not working regardless of changes in this branch. However, datatype changes were made to be consistent with msg changes for future work fixing the ros_test.*


owl_msgs/msg/ArmEndEffectorForceTorque.msg
```diff
@@ -0,0 +1,4 @@
+# Arm force-torque sensor force and torque
+
+Header header
+float64[6] value # force(N) [0-2], torque(N-m) [3-5], end effector frame
```
kmdalal, line 4: Minor: I think this would be a little easier to read with the comment on the preceding line.

kmdalal:
The instructions should state that `ow_autonomy` must be on the same branch.  Also, this branch should be brought up to date with the target epic branch.

kmdalal:
All the tests except the `rostest` pass for me as well, and I get the same failure and output as reported by @AstroStucky .

kmdalal:
The code changes look fine.  However, I'm having second thoughts on the value of this change for OceanWATERS:
 - Is the new topic name deceptive, i.e. is it accurate to call it end-effector FT when the sensor is on the distal pitch joint?
 - Even though we decided on it, the downgrade from Wrench to Float[6] doesn't look great :(
 
We can discuss these in Dev Day.

ow_faults_injection/src/FaultInjector.cpp
```diff
@@ -75,33 +75,40 @@ void FaultInjector::jointStateCb(const sensor_msgs::JointStateConstPtr& msg)
...
   }
 
-  auto out_msg = msg;
+  owl_msgs::ArmEndEffectorForceTorque out_msg;
+  out_msg.header = msg.header;
 
   if (m_faults.groups.ft_sensor_faults.zero_signal_failure) {
-    out_msg.wrench.force = geometry_msgs::Vector3();
-    out_msg.wrench.torque = geometry_msgs::Vector3();
+    out_msg.value = {0.0};
```
AstroStucky, line 89: I find this a little bit misleading. When I saw this I thought it was an array fill method that I had never seen before. But in fact what is being done here is explicitly assigning the first value to zero, then zero-initializing all the remaining values. In the end it gets you the result that you want, but I feel it misleads the reader. 
Some alternatives I would propose
Option 1: all 
```suggestion
    out_msg.value = {0.0, 0.0, 0.0, 0.0, 0.0, 0.0};
```
or Option 2: Nothing (haven't tested this but I think it works)
```suggestion
    out_msg.value = {};
```
Option 3: Use a method which would require an additional std include of **algortihm**
```suggestion
std:fill(out_msg.value.begin(), out_msg.value.end(), 0.0);
```

AstroStucky:
> * Even though we decided on it, the downgrade from Wrench to Float[6] doesn't look great :(

I have to agree. It still feels like a step backwards in readability of our telemetry. I feel less strongly about the name change, so long as we document the location of the FT sensor somewhere. This would be a good topic of discussion for the dev day tomorrow.

AstroStucky (commented):
I looked into the test that was failing, and I believe it would have failed before this PR because it is attempting to call the stow and unstow ROS Services, which don't even run anymore. Fixing this test would require substantially more work, and I don't think it's worth doing in this PR.

ow_faults_injection/src/FaultInjector.cpp
kmdalal: I agree that @AstroStucky 's suggestions would be clearer.  I like the first and third best.

anthonykoutroulis:
Good code change suggestions @AstroStucky , I'll incorporate those.
> I looked into the test that was failing, and I believe it would have failed before this PR because it is attempting to call the stow and unstow ROS Services, which don't even run anymore. Fixing this test would require substantially more work, and I don't think it's worth doing in this PR.
Sounds right, I was getting the timeout without the changes when I tested.
> The code changes look fine. However, I'm having second thoughts on the value of this change for OceanWATERS:
I also agree that there is readability loss with these unifications. I'll revisit during DevDay.

kmdalal:
> > * Even though we decided on it, the downgrade from Wrench to Float[6] doesn't look great :(
> 
> I have to agree. It still feels like a step backwards in readability of our telemetry. I feel less strongly about the name change, so long as we document the location of the FT sensor somewhere.

Per our discussion today, we decided to stick with Wrench.

kmdalal (approved):
Looks good on my end!

AstroStucky (approved):
No changes look good and test still works.

kmdalal:
The Jira for this is closed, and this PR was fully approved last week -- is it waiting anything for merge?

PR #311, opened 2023-01-24
Oceanwater 948 ArmMoveJointsGuarded added

AstroStucky:
## Linked Issues:
| Jira Ticket 🎟️ | [OCEANWATER-948](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-948) |

## Summary of Changes
* **ArmMoveJointsGuarded** action added
* minor reworking of **ModifyJointValuesMixin**

### Help Message
Make sure language and the way in which values are accepted make sense for the action client script. 
```
rosrun ow_lander arm_move_joints_guarded.py --help
```

### Test action
Perform the following test in atacama_y1a for consistent results

#### 1. Out stretch the arm by passing zeroes for all joint values
```
rosrun ow_lander arm_move_joints_guarded.py 0 0 0 0 0 0
```
You'll see the following
```
ArmMoveJointsGuarded action started
ArmMoveJointsGuarded: Succeeded - ArmMoveJointsGuarded trajectory completed without breaching force or torque thresholds
ArmMoveJointsGuarded action complete
```
default_gzclient_camera(1)-2023-01-24T13_34_03 358856

#### 2. Pitch shoulder down to breach the force threshold
```
rosrun ow_lander arm_move_joints_guarded.py 0 -0.2 0 0 0 0
```
You'll see 
```
ArmMoveJointsGuarded action started
ArmMoveJointsGuarded: Succeeded - ArmMoveJointsGuarded trajectory stopped by a force of 621.50 N
ArmMoveJointsGuarded action complete
```
default_gzclient_camera(1)-2023-01-24T13_37_11 979438

#### 3. Reset to zero positions and trigger the torque threshold by yawing the shoulder joint, and colliding the scoop with a nearby mound. Here we also lower the torque value to ensure it triggers. 
```
rosrun ow_lander arm_move_joints_guarded.py 0 0 0 0 0 0
rosrun ow_lander arm_move_joints_guarded.py 0.4 0 0 0 0 0 --torque 50
```
You'll see
```
ArmMoveJointsGuarded action started
ArmMoveJointsGuarded: Succeeded - ArmMoveJointsGuarded trajectory stopped by a torque of 103.58 Nm
ArmMoveJointsGuarded action complete
```
default_gzclient_camera(1)-2023-01-24T13_39_47 433192

### DIFF
--- a/ow_lander/scripts/arm_move_joints_guarded.py
+++ b/ow_lander/scripts/arm_move_joints_guarded.py
@@ -11,20 +11,22 @@ import argparse
 
 parser = argparse.ArgumentParser(
   formatter_class=argparse.ArgumentDefaultsHelpFormatter,
-  description="Move all arm joints to either absolute or relative positions "
-              "and stop movement if resistance is encountered.")
+  description="Move all arm joints to either absolute or relative positions. "
+              "Movement will cease if the distal pitch force/torque sensor "
+              "detects a force/torque larger than the provided values. Torque "
+              "of individual arm joints will not cause movement to stop.")
 parser.add_argument('shoulder_yaw', type=float,
-  help="Shoulder yaw joint position/translation in radians")
+  help="Shoulder yaw joint position/rotation in radians")
 parser.add_argument('shoulder_pitch', type=float,
-  help="Shoulder pitch joint position/translation in radians")
+  help="Shoulder pitch joint position/rotation in radians")
 parser.add_argument('proximal_pitch', type=float,
-  help="Proximal pitch joint position/translation in radians")
+  help="Proximal pitch joint position/rotation in radians")
 parser.add_argument('distal_pitch', type=float,
-  help="Distal pitch joint position/translation in radians")
+  help="Distal pitch joint position/rotation in radians")
 parser.add_argument('hand_yaw', type=float,
-  help="Hand yaw joint position/translation in radians")
+  help="Hand yaw joint position/rotation in radians")
 parser.add_argument('scoop_yaw', type=float,
-  help="Scoop yaw joint position/translation in radians")
+  help="Scoop yaw joint position/rotation in radians")
 parser.add_argument('--relative', '-r', action='store_true', default=False,
   help="If true, joints will be moved to their current position plus the "
        "value provided. If false, joints will move to absolute positions "
--- a/ow_lander/scripts/arm_move_joints_guarded.py
+++ b/ow_lander/scripts/arm_move_joints_guarded.py
@@ -1,0 +1,51 @@
+#!/usr/bin/env python3
+
+# The Notices and Disclaimers for Ocean Worlds Autonomy Testbed for Exploration
+# Research and Simulation can be found in README.md in the root directory of
+# this repository.
+
+from ow_lander import actions
+from ow_lander import node_helper
+
+import argparse
+
+parser = argparse.ArgumentParser(
+  formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+  description="Move all arm joints to either absolute or relative positions "
+              "and stop movement if resistance is encountered.")
+parser.add_argument('shoulder_yaw', type=float,
+  help="Shoulder yaw joint position/translation in radians")
+parser.add_argument('shoulder_pitch', type=float,
+  help="Shoulder pitch joint position/translation in radians")
+parser.add_argument('proximal_pitch', type=float,
+  help="Proximal pitch joint position/translation in radians")
+parser.add_argument('distal_pitch', type=float,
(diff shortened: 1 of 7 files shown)


ow_lander/scripts/arm_move_joints_guarded.py
```diff
@@ -0,0 +1,51 @@
...
+from ow_lander import node_helper
+
+import argparse
+
+parser = argparse.ArgumentParser(
+  formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+  description="Move all arm joints to either absolute or relative positions "
+              "and stop movement if resistance is encountered.")
+parser.add_argument('shoulder_yaw', type=float,
+  help="Shoulder yaw joint position/translation in radians")
```
mogumbo, line 17: Should say "rotation" or "position/rotation." These definitely aren't translation joints. ROS refers to all joint displacements as "position" whether you're using a rotation or translation joint, but most robot joints tend to be rotational.

owl_msgs/action/ArmMoveJointsGuarded.action
```diff
@@ -0,0 +1,15 @@
+# goal definition
+bool relative
+float64[] angles
```
mogumbo, line 3: To be ROS-like and future-proof, you could call this "positions." And it appears the word "angle" is propagated through the code.

mogumbo (changes requested):
There should be a comment in the script or maybe a caveat that prints in the help message that explains the limitation of the force/torque sensor being attached to j_dist_pitch.

Now we have scripts arm_move_joint_client.py, arm_move_joints_client.py, and arm_move_joints_guarded.py and I think the differences between scripts would be unclear to many users. Should we have another ticket to remove or rename any of these, or maybe add more descriptive help messages and comments?

Does arm_move_joints_client.py behave the same as this new script with force and torque set to very high values? Do we even need both scripts?

owl_msgs/action/ArmMoveJointsGuarded.action
AstroStucky: Good to know. I am unsure whether a change was requested here?

AstroStucky:
> There should be a comment in the script or maybe a caveat that prints in the help message that explains the limitation of the force/torque sensor being attached to j_dist_pitch.

Done. I added the same language to `arm_move_cartesian_guarded.py` as well, which has the same limitation.

> Now we have scripts arm_move_joint_client.py, arm_move_joints_client.py, and arm_move_joints_guarded.py and I think the differences between scripts would be unclear to many users. Should we have another ticket to remove or rename any of these, or maybe add more descriptive help messages and comments?

I have made a ticket for this. Could you clarify what differences you are referring to? Is it that `arm_move_joints_guarded.py` takes command line arguments differently than `arm_move_joints_client.py`? Feel free to clarify the ticket I made yourself. https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1110

> Does arm_move_joints_client.py behave the same as this new script with force and torque set to very high values? Do we even need both scripts?

The outputs would be different. If very high values are set for the **ArmMoveJointsGuarded** action and it fails to reached requested joint positions, it will suggest to the user that force/torque values may be set too high. If the **ArmMoveJoints** action does not reach its joint values, it will also fail and return a message that does not mention force/torque values. 

Successful runs of the action also result in different messages, but really the difference would only come down to different log messages. I suppose we could make these the same action, or at the very least let them share more code. 

On a slightly related note, while the action server refactor did a lot to minimize code repetition, implementing these new actions has shown me how it could improved to better allow actions to share operations with a much smaller scale refactor of the action server interface. I do not want to take this on before our next release, but I do want to mention the possibility as another way to minimize code repetition between these two actions.

ow_lander/scripts/arm_move_joints_guarded.py
AstroStucky: Done.

owl_msgs/action/ArmMoveJointsGuarded.action
mogumbo: It's up to you. Right now we're dealing with only angles. If we ever added a different type of joint that didn't rotate, then some refactoring would be in order for the code to make sense.

mogumbo (approved):

owl_msgs/action/ArmMoveJointsGuarded.action
AstroStucky: I do like positions better than angles, but I will opt to stick with the unification doc. It's a good note for the future.

kmdalal: I agree it was best to stick with the unified doc on the name.  Good thing to point out as a comment in that doc.

PR #315, opened 2023-01-31
Oceanwater 1106 minor message edits

anthonykoutroulis:
This is a minor fix that addresses a missing publishing of the camera fault and a simple renaming of the battery fault messages.
The testing is mostly regression testing.

Before testing, build with respective branch on [ow_autonomly](https://github.com/nasa/ow_autonomy/pull/110).
---
Testing: 

- Launch the atacama sim and the ow_plexil node in separate sourced terminals.
- Test the battery faults and camera faults
- Observe that the camera fault does get set to NO_IMAGE and is exonerated both in plexil and the rostopics
- Observe that the battery faults work under the rename.


ow_faults_detection/src/FaultDetector.cpp
```diff
@@ -74,8 +74,10 @@ void FaultDetector::cameraPublishFaultMessages(bool is_fault)
   owl_msgs::CameraFaultsStatus camera_faults_msg;
```
kmdalal, line 74: The header doesn't seem to be set, e.g. with `setFaultsMessageHeader`

anthonykoutroulis: Fixed!

kmdalal (approved):
Looks good!  I tried several different tests, including regression using ReferenceMission1.

PR #317, opened 2023-02-03
Oceanwater 1100 refactor fault publishing

anthonykoutroulis:
This is strictly a code cleanup of the fault detector module. A lot of redundant publishing functions were introduced during telemetry unification, but the components essentially all do the same thing: set flags based on fault logic matching the component's fault message (from owl_msgs), similarly update system faults message, and publish the messages with values that directly correspond to the fault flags.

This PR restructures the code around that design and uses a single generic `publishFaultMessage` function and removes individual publishing functions

## Summary of Changes
* Update component listeners to set fault flags with a standardized format
* Introduce a generic `publishFaultsMessage` function which is called whenever a listener updates fault flags
* Remove specific publishing messages and state variables

## Test
* Build with the [OceanWATERS-934_telemetry_unification_epic](https://github.com/nasa/ow_autonomy/tree/OCEANWATER-934_telemetry_unification_epic) branch
* launch the Atacama sim
* Open five additional sourced terminals, each subscribing to one of the faults topics
  * `rostopic echo /system_faults_status`  
  * `rostopic echo /arm_faults_status`
  * `rostopic echo /camera_faults_status`
  * `rostopic echo /pan_tilt_faults_status`
  * `rostopic echo /power_faults_status`
 * Exercise each fault with the appropriate fault injections
 * Observe fault topics change and exonerate values appropriately

### DIFF
--- a/ow_faults_detection/include/ow_faults_detection/FaultDetector.h
+++ b/ow_faults_detection/include/ow_faults_detection/FaultDetector.h
@@ -33,7 +33,7 @@ class FaultDetector
 {
 public:
   FaultDetector(ros::NodeHandle& nh);
-  ~FaultDetector(){}
+  ~FaultDetector() = default;
   
   FaultDetector (const FaultDetector&) = delete;
   FaultDetector& operator= (const FaultDetector&) = delete;
@@ -69,8 +69,8 @@ private:
   // OWLAT MESSAGE FUNCTIONS AND PUBLISHERS
   template<typename fault_msg>
   void setFaultsMessageHeader(fault_msg& msg);
-  template<typename faults_msg_pub_t, typename faults_msg_t, typename faults_flags_t>
-  void publishFaultsMessage(faults_msg_pub_t& faults_msg_pub, faults_msg_t& faults_msg, faults_flags_t faults_flags);
+  template<typename pub_t, typename msg_t, typename flags_t>
+  void publishFaultsMessage(pub_t& faults_pub, msg_t faults_msg, flags_t faults_flags);
 
   // PUBLISHERS AND SUBSCRIBERS
   
--- a/ow_faults_detection/include/ow_faults_detection/FaultDetector.h
+++ b/ow_faults_detection/include/ow_faults_detection/FaultDetector.h
@@ -57,24 +57,20 @@ private:
   // Get index from m_joint_index_map. If found, modify out_index and return
   // true. Otherwise, return false.
   bool findJointIndex(const unsigned int joint, unsigned int& out_index);
-
-  // Antenna
-  void antPublishFaultMessages();
   
   // Camera
   void cameraTriggerCb(const std_msgs::Empty& msg);
   void cameraRawCb(const sensor_msgs::Image& msg);
-  void cameraPublishFaultMessages(bool is_fault);
 
   // Power
-  void publishPowerSystemFault();
   void powerSOCListener(const owl_msgs::BatteryStateOfCharge& msg);
   void powerTemperatureListener(const owl_msgs::BatteryTemperature& msg);
 
   // OWLAT MESSAGE FUNCTIONS AND PUBLISHERS
-  void publishSystemFaultsMessage();
   template<typename fault_msg>
   void setFaultsMessageHeader(fault_msg& msg);
+  template<typename faults_msg_pub_t, typename faults_msg_t, typename faults_flags_t>
+  void publishFaultsMessage(faults_msg_pub_t& faults_msg_pub, faults_msg_t& faults_msg, faults_flags_t faults_flags);
 
   // PUBLISHERS AND SUBSCRIBERS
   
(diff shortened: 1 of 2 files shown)


ow_faults_detection/include/ow_faults_detection/FaultDetector.h
```diff
@@ -57,24 +57,20 @@ class FaultDetector
...
-  void publishPowerSystemFault();
   void powerSOCListener(const owl_msgs::BatteryStateOfCharge& msg);
   void powerTemperatureListener(const owl_msgs::BatteryTemperature& msg);
 
   // OWLAT MESSAGE FUNCTIONS AND PUBLISHERS
-  void publishSystemFaultsMessage();
   template<typename fault_msg>
   void setFaultsMessageHeader(fault_msg& msg);
+  template<typename faults_msg_pub_t, typename faults_msg_t, typename faults_flags_t>
+  void publishFaultsMessage(faults_msg_pub_t& faults_msg_pub, faults_msg_t& faults_msg, faults_flags_t faults_flags);
```
kmdalal, line 73: Just  readability suggestion: parameter names can be omitted in declarations, and are not helpful here since the typename tells all.  This would also shorten the line.

ow_faults_detection/include/ow_faults_detection/FaultDetector.h
```diff
@@ -85,32 +81,23 @@ class FaultDetector
...
-
-  // Camera
+  // faults topic subscribers;
+  ros::Subscriber m_joint_states_sub; // for arm and antenna
   ros::Subscriber m_camera_original_trigger_sub;
   ros::Subscriber m_camera_raw_sub;
-
-  // Power
   ros::Subscriber m_power_soc_sub;
   ros::Subscriber m_power_temperature_sub;
```
kmdalal, line 89: Unrelated to the intended change, but these members are not necessary since they are are used only to hold a value and are never referenced individually after assignment.   They could be replaced with a single member: std::list<ros::Subscriber>, and each subscriber can simply be pushed into this list.

ow_faults_detection/src/FaultDetector.cpp
```diff
@@ -59,52 +59,20 @@ void FaultDetector::setFaultsMessageHeader(fault_msg& msg)
   msg.header.frame_id = "world";
 }
 
-// publish system messages
-void FaultDetector::publishSystemFaultsMessage()
+template<typename faults_msg_pub_t, typename faults_msg_t, typename faults_flags_t>
+void FaultDetector::publishFaultsMessage(faults_msg_pub_t& faults_msg_pub, faults_msg_t& faults_msg, faults_flags_t faults_flags)
```
kmdalal, line 63: Readability suggestion: the parameter names could be a lot shorter and simpler, e.g. `pub or publisher, msg or message, flags`.  And the line should be shortened in any case -- an easy way is to stack the parameters.

ow_faults_detection/src/FaultDetector.cpp
```diff
@@ -133,31 +101,55 @@ void FaultDetector::jointStatesFlagCb(const ow_faults_detection::JointStatesFlag
...
+  }
+  if (PanTiltFaultsStatus::NONE == m_antenna_faults_flags) {
+    m_system_faults_flags &= ~SystemFaultsStatus::PAN_TILT_EXECUTION_ERROR;
+  } else {
+    m_system_faults_flags |= SystemFaultsStatus::PAN_TILT_EXECUTION_ERROR;
+  }
+
+  // publish updated faults messages
+  owl_msgs::ArmFaultsStatus arm_faults_msg;
+  publishFaultsMessage(m_arm_faults_msg_pub, arm_faults_msg, m_arm_faults_flags);
```
kmdalal, line 148: This comment applies to many similar instances in this file.  These two lines could be combined into just:

`publishFaultsMessage(m_arm_faults_msg_pub, owl_msgs::ArmFaultsStatus(), m_arm_faults_flags);`

In either case, passing a static (stack-allocated) object can be unsafe, but it's fine in this case because the object is no longer referenced after the called function returns.

ow_faults_detection/include/ow_faults_detection/FaultDetector.h
```diff
@@ -85,32 +81,23 @@ class FaultDetector
...
-  bool m_pan_fault;
-  bool m_tilt_fault;
-  
-  // Camera
   bool m_camera_data_pending = false;
-  
-  // Power
-  uint64_t m_power_faults_flags = 0;
   float m_last_soc = std::numeric_limits<float>::quiet_NaN();
 };
```
kmdalal, line 103: Minor: GitHub won't let me comment on the relevant line because it's not part of the diff.  But `~FaultDetector(){}` could be replaced with `~FaultDetector() = default;` which is bit more proper.

anthonykoutroulis:
@kmdalal I updated with your suggestions and removed some unneeded scopes for further readability

kmdalal (approved):
Tests pass, good refactoring!

PR #325, opened 2023-03-08
More rigorous transform waiting

AstroStucky:
## Linked Issues:
Jira Ticket 🎟️ | [OCEANWATER-1119](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1119)

## Reviewer Instructions
@mikedalal Only perform the test and report on results.
@mogumbo Only review the code (and approach).

## Summary of Changes
* `rospy.Time.now()` used in favor of `rospy.Time(0)`. 
* Increased the default timeout of FrameTransformer to 1 second. 
* Made timestamp and non-optional parameter for FrameTransformer methods, which makes it easier to understand what a method call is doing. 

## Test
The following actions may use a frame transform depending on the parameters passed: TaskScoopCircular, TaskScoopLinear, TaskDiscardSample, ArmMoveCartesian, ArmMoveCartesianGuarded, ArmFindSurface, and PanTiltCartesian. This test will call each one with the appropriate parameters to make it uses a frame transform. The test passes if no errors result from these calls and movement looks expected. 

1. Grind a trench so that scoop tasks can be called 
`rosrun ow_lander task_grind.py`

2. Perform circular scoop
`rosrun ow_lander task_scoop_circular.py`

3. Discard sample
`rosrun ow_lander task_discard_sample.py`

4. Perform linear scoop
`rosrun ow_lander task_scoop_linear.py`

5. Perform a relative motion, un-guarded
`rosrun ow_lander arm_move_cartesian.py -f 1 -y 0.4`

6. Move back to the previous position, guarded
`rosrun ow_lander arm_move_cartesian_guarded.py -f 1 -y -0.4`

7. Orient scoop into a better position for a TOOL-relative motion.
`rosrun ow_lander arm_move_cartesian.py -x 1.7 -y 0.4 -z 0.2 -e 3.14 0 0`

8. Interrogate surface directly below scoop
`rosrun ow_lander arm_find_surface.py -f 1 -z 0.3`

9. Look at scoop
`rosrun ow_lander pan_tilt_move_cartesian.py -f 1`

### DIFF
--- a/ow_lander/src/ow_lander/common.py
+++ b/ow_lander/src/ow_lander/common.py
@@ -4,12 +4,13 @@
 
 """Defines functions required by multiple modules within the package"""
 
-from math import pi, tau, acos, sqrt
+from math import pi, tau
 from ow_lander import constants
 from std_msgs.msg import Header
 from rospy import Time
-# import roslib; roslib.load_manifest('urdfdom_py')
+
 from urdf_parser_py.urdf import URDF
+
 
 class Singleton(type):
   """When passed to the metaclass parameter in the class definition, the class
@@ -52,5 +53,5 @@ def in_closed_range(val, lo, hi):
   """
   return val >= lo and val <= hi
 
-def create_header(frame_id, timestamp):
+def create_header(frame_id, timestamp=Time(0)):
   return Header(0, timestamp, frame_id)
--- a/ow_lander/src/ow_lander/common.py
+++ b/ow_lander/src/ow_lander/common.py
@@ -52,5 +52,5 @@ def in_closed_range(val, lo, hi):
   """
   return val >= lo and val <= hi
 
-def create_most_recent_header(frame_id):
-  return Header(0, Time(0), frame_id)
+def create_header(frame_id, timestamp):
+  return Header(0, timestamp, frame_id)
--- a/ow_lander/src/ow_lander/frame_transformer.py
+++ b/ow_lander/src/ow_lander/frame_transformer.py
@@ -45,16 +45,16 @@ class FrameTransformer(metaclass = Singleton):
     stamped = None
     if isinstance(geometry, Point):
       stamped = PointStamped(
-        header=create_header(source_frame, rospy.Time.now()), point=geometry)
+        header=create_header(source_frame, timestamp), point=geometry)
     elif isinstance(geometry, Pose):
       stamped = PoseStamped(
-        header=create_header(source_frame, rospy.Time.now()), pose=geometry)
+        header=create_header(source_frame, timestamp), pose=geometry)
     elif isinstance(geometry, Vector3):
       stamped = Vector3Stamped(
-        header=create_header(source_frame, rospy.Time.now()), vector=geometry)
+        header=create_header(source_frame, timestamp), vector=geometry)
     elif isinstance(geometry, Wrench):
       stamped = WrenchStamped(
-        header=create_header(source_frame, rospy.Time.now()), wrench=geometry)
+        header=create_header(source_frame, timestamp), wrench=geometry)
     else:
       rospy.logerr(f"Unsupported geometry type {type(geometry)}")
       return None
--- a/ow_lander/src/ow_lander/frame_transformer.py
+++ b/ow_lander/src/ow_lander/frame_transformer.py
@@ -30,6 +30,20 @@ class FrameTransformer(metaclass = Singleton):
     geometry -- A geometry_msgs object. Supported: Point, Pose, Vector3, Wrench
     target_frame -- ROS identifier string for the frame to be transformed into
     source_frame -- ROS identifier string from the frame being transformed from
+    timestamp -- The time of the desired transform. Assigning rospy.Time(0) will
+                 request the most recent transform available in the cache. Even
+                 older transforms may be requested, but be careful not to
+                 request transforms further in the past than the cache saves.
+                 Using ropsy.Time.now() will acquire the next (newest)
+                 transform. If a rospy.Time.now() or a transform in the future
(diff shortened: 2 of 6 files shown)


kmdalal:
The arm_find_surface failed:
```
[INFO]: ArmFindSurface action started
[ERROR]: ArmFindSurface: Aborted - Arm failed to reach pose despite neither force nor torque thresholds being breached. Likely the thresholds were set too high or a planning error occurred. Try a lower threshold.
[INFO]: ArmFindSurface action complete
```

AstroStucky:
I think it is unrelated to the PR, and it already has a ticket [OCEANWATER-1117](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1117)

kmdalal (approved):
Tests passed, except for one that was a known and unrelated problem.

Foregoing a review of the code.

ow_lander/src/ow_lander/common.py
```diff
@@ -52,5 +52,5 @@ def in_closed_range(val, lo, hi):
   """
   return val >= lo and val <= hi
 
-def create_most_recent_header(frame_id):
-  return Header(0, Time(0), frame_id)
+def create_header(frame_id, timestamp):
```
mogumbo, line 55: How about making timestamp default to rospy.Time.now()? Then you won't have to put it on every line that calls create_header()

ow_lander/src/ow_lander/frame_transformer.py
```diff
@@ -23,14 +23,14 @@ class FrameTransformer(metaclass = Singleton):
...
+  DEFAULT_TIMEOUT = rospy.Duration(1.0)
 
   def __init__(self):
     self._buffer = tf2_ros.Buffer()
     self._listener = tf2_ros.TransformListener(self._buffer)
 
-  def transform_present(self, geometry, target_frame, source_frame, \
-      timeout=DEFAULT_TIMEOUT):
+  def transform_geometry(self, geometry, target_frame, source_frame, \
+      timestamp, timeout=DEFAULT_TIMEOUT):
```
mogumbo, line 33: Should this timestamp also default to rospy.Time.now()?

mogumbo (changes requested):

ow_lander/src/ow_lander/common.py
AstroStucky: `Time(0)` is a constant expression, but `rospy.Time.now()` is not. It will return whatever time it is now. If I did this the definition would have to look like this 
```suggestion
def create_header(frame_id, timestamp=None):
  if timestamp is None:
     return Header(0, rospy.Time.now(), frame_id)
```

I believe it is perfectly fine to define a function like this, but considering using the default value in this case hides the fact that when you call the function changes its return value in a way that is not very easy to communicate in the function name, I would much rather require a timestamp each time it is called.

mogumbo: Thanks. I understand a little better what you're doing now. I still prefer default function parameters just to reduce code, but I don't prefer it strongly enough to hold up this PR.

Most of the time I would expect very little difference between using Time.now() and Time(0). Did you see any change in behavior from switching to Time.now()?

AstroStucky: Not really, just that `Time(0)` did not behave how the tf2 documentation described. Its meant to grab the "most recent" timestamp, but if that's what it indeed does then why would an error occur saying the frame does not exist? And why should increasing the timeout prevent that error from occuring? If it has to wait at all, then there is no way it is grabbing the most recent.

Since it was not behaving how the documentation said, I just felt it was safer to use something that does behave how the documentation describes e.g. `Time.now()`.

ow_lander/src/ow_lander/frame_transformer.py
```diff
@@ -45,16 +45,16 @@ def transform_present(self, geometry, target_frame, source_frame, \
     stamped = None
     if isinstance(geometry, Point):
       stamped = PointStamped(
-        header=create_most_recent_header(source_frame), point=geometry)
+        header=create_header(source_frame, rospy.Time.now()), point=geometry)
```
mogumbo, line 48: Shouldn't `timestamp` be used instead of `rospy.Time.now()`? In fact, `timestamp` isn't used at all in this method.

AstroStucky:
@mogumbo pointed out that `Time(0)` was likely not working as documented because a non-zero duration was being passed to `timeout`. Turns out this was indeed the case. Great catch! 

Now the only calls that request a `Time.now()` transform are those that are called immediately following a movement. This is in case the most recent transform was recorded during said movement. All other calls simply request the most recent transform since that should be sufficient. 

This PR requires a re-review and re-test.

kmdalal (approved):
Approved just from the test viewpoint, but see my comment on a potential unrelated bug.

mogumbo:
I think functions that look up transforms should now at least have a clear comment that explains how to use them safely. But it would be even better to have error checking that warns or quits when using Time.now() with duration 0 or when using Time(0) with a non-zero duration. What do you think?

AstroStucky:
> I think functions that look up transforms should now at least have a clear comment that explains how to use them safely. But it would be even better to have error checking that warns or quits when using Time.now() with duration 0 or when using Time(0) with a non-zero duration. What do you think?

Detailed comments added. 

I stopped short of error checking because it seemed unnecessary and complicated to do. As for the specific scenarios you mentioned: it would be difficult to tell if Time.now() is used, since its a method called outside of these functions. For the latter scenario, a non-zero duration with Time(0) should be perfectly fine to do, it will just return immediately and the timeout is unnecessary. 

Hopefully my method comments communicate the pitfalls of these arguments well enough without error checking.

mogumbo (approved):

PR #327, opened 2023-03-28
Work-around for follow joint trajectory commonly resulting in error code

AstroStucky:
## Linked Issues:
* [OW-1090](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1090) - follow_joint_trajectory action server usually results in an error code
* [OW-1064](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1064) - Arm trajectories sometimes cannot be stopped
* [OW-1097](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1097) - /arm_controller/follow_joint_trajectory action sometimes returns a result before trajectory has completed

## Summary of Changes
* The `goal_tolerance` defined in `ros_controllers.yaml` is now used instead of being overridden by a much smaller value in a method call. This fixes GOAL_TOLERANCE_VIOLATION error code resulting for stow, unstow, and likely other actions. 
* All values of `traj_tolerance` defined in `ros_controllers.yaml` changed to zero, effectively disabling path tolerance checking. This works around PATH_TOLERANCE_VIOLATION, which was an error that was not handled anyways and would only result in being unable to stop the arm. 

## Caveat
Ideally we would have path tolerance checking enabled. The difficulty is some of our generated plans result in a discontinuity in joint positions, typically j_hand_yaw's positions. The discontinuity appears to simply be the joint position wrapping around by 2*pi when it goes too far in one direction or the other, but the follow_joint_trajectory actions interpret this as a path tolerance violation.

This may take some time to figure out. For now its worthwhile to just ignore path checking, which we already were doing before by not handling the error code. 

**Note:** The error code from follow_joint_trajectory is still not handled. This will be in another PR and is capture by [OW-1091](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1091)

## Test 1 - Consecutive Actions Calls
This will verify that OW-1097 is solved. 
Open any simulator world
1. Monitor the arm_controller result topic
```
rostopic echo /arm_controller/follow_joint_trajectory/result
```
2. Monitor the grinder_controller result topic
```
rostopic echo /grinder_controller/follow_joint_trajectory/result
```
3. Call several arm trajectories in quick succession. The following command will call them immediately after the prior using the `;` operator. 
 ```
rosrun ow_lander task_grind.py ; rosrun ow_lander task_scoop_linear.py ; rosrun ow_lander task_deliver_sample.py ; rosrun ow_lander arm_move_cartesian.py -x 1.8 -z 0.5
```
### Expected Outcome
a. Watch the output of the two rostopics and confirm the end of each of the 4 actions results in `error_code: 0`
b. Confirm actions all execute one after the other. The simulator output should look like so, message order matters.
```
[INFO:/lander_action_servers] TaskGrind action started
[INFO:/lander_action_servers] TaskGrind: Succeeded - TaskGrind trajectory succeeded
[INFO:/lander_action_servers] TaskGrind action complete
[INFO:/lander_action_servers] TaskScoopLinear action started
[INFO:/lander_action_servers] TaskScoopLinear: Succeeded - Arm trajectory succeeded
[INFO:/lander_action_servers] TaskScoopLinear action complete
[INFO:/lander_action_servers] TaskDeliverSample action started
[INFO:/lander_action_servers] TaskDeliverSample: Succeeded - Arm trajectory succeeded
[INFO:/lander_action_servers] TaskDeliverSample action complete
[INFO:/lander_action_servers] ArmMoveCartesian action started
[INFO:/lander_action_servers] ArmMoveCartesian: Succeeded - ArmMoveCartesian trajectory succeeded
[INFO:/lander_action_servers] ArmMoveCartesian action complete
```

## Test 2 - Stop action
The following can be done in the same simulation instance. We will be replicating one of the specific cases where `arm_stop.py` would be ignored, and this will require some precise timing. 
1. In a separate terminal type out the command `rosrun ow_lander arm_stop.py`, but do not press enter.
2. Call the following command sequence, and pay close attention to the Gazebo client.
```
rosrun ow_lander task_grind.py ; rosrun ow_lander task_scoop_linear.py 
```
3. Allow the TaskGrind action to complete. 
4. At the start of the TaskScoopLinear action the scoop will move from this pose
default_gzclient_camera(1)-2023-03-28T14_58_54 501728

to this pose

default_gzclient_camera(1)-2023-03-28T14_59_12 519041
In between these two poses, send the `arm_stop.py` command we previously typed. 
### Expected Outcome
The arm should stop. Previously ArmStop would be ignored at this point in the trajectory.

### DIFF
--- a/ow_lander/config/ros_controllers.yaml
+++ b/ow_lander/config/ros_controllers.yaml
@@ -80,6 +80,10 @@ arm_controller:
     # Velocity to be considered approximately equal to zero.
     stopped_velocity_tolerance: 0.001
     # Position tolerance for a particular joint to reach the goal and throughout the trajectory
+    # NOTE: A zero trajectory tolerance is a work-around for joint trajectories
+    #       being generated that contain a discontinuity in their positions. The
+    #       discontinuity would always result in a PATH_TOLERANCE_VIOLATION.
+    # j_shou_yaw: {trajectory: &traj_tolerance 0.5, goal: &goal_tolerance 0.1}
     j_shou_yaw: {trajectory: &traj_tolerance 0, goal: &goal_tolerance 0.1}
     j_shou_pitch: {trajectory: *traj_tolerance, goal: *goal_tolerance}
     j_prox_pitch: {trajectory: *traj_tolerance, goal: *goal_tolerance}
--- a/ow_lander/config/ros_controllers.yaml
+++ b/ow_lander/config/ros_controllers.yaml
@@ -80,7 +80,7 @@ arm_controller:
     # Velocity to be considered approximately equal to zero.
     stopped_velocity_tolerance: 0.001
     # Position tolerance for a particular joint to reach the goal and throughout the trajectory
-    j_shou_yaw: {trajectory: &traj_tolerance 0.5, goal: &goal_tolerance 0.1}
+    j_shou_yaw: {trajectory: &traj_tolerance 0, goal: &goal_tolerance 0.1}
     j_shou_pitch: {trajectory: *traj_tolerance, goal: *goal_tolerance}
     j_prox_pitch: {trajectory: *traj_tolerance, goal: *goal_tolerance}
     j_dist_pitch: {trajectory: *traj_tolerance, goal: *goal_tolerance}
--- a/ow_lander/src/ow_lander/arm_interface.py
+++ b/ow_lander/src/ow_lander/arm_interface.py
@@ -269,16 +269,13 @@ class ArmTrajectoryExecutor(metaclass=Singleton):
     def get_active_controller(self):
         return self._active_controller
 
-    def execute(self, trajectory, goal_time_tolerance=rospy.Time(0.1),
+    def execute(self, trajectory,
                 done_cb=None, active_cb=None, feedback_cb=None):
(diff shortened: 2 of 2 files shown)


kmdalal:
It seems this should be a hotfix to Release 11...

kmdalal:
Test 1 passed, though for some reason all INFO messages started printing in green instead of white beginning with `TaskScoopLinear: Succeeded - Arm trajectory succeeded`.  I doubt it's related to this PR but seems quite odd.

AstroStucky:
> Test 1 passed, though for some reason all INFO messages started printing in green instead of white beginning with `TaskScoopLinear: Succeeded - Arm trajectory succeeded`. I doubt it's related to this PR but seems quite odd.

I see this very commonly. It happens with everything I do in ROS, so as best as I can tell its a ROS logging bug.

kmdalal (approved):
Test 2 also passed, and good job isolating the place where I had seen Stop fail.  I can't assess the code changes due to unfamiliarity with this subsystem, but they look reasonable.  Good investigation; I hope this fix puts an end to a variety of long-standing intermittent issues!

Noted that you're converting this to a hotfix and will need a second approver.

PR #329, opened 2023-04-06
Oceanwater 1120 light params switch lights

mogumbo:
## Linked Issues:
| Jira Ticket 🎟️ | [OCEANWATER-1120](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1120) |

## Summary of Changes
* Control light intensities by setting an Ogre-controlled shader uniform instead of a custom uniform. This ensures that light intensities go to the correct lights.
* You will also need sibling PR: https://github.com/nasa/ow_europa/pull/23

## Test
* Run any ow sim world
* rosrun ow_lander light_set_intensity.py left 0.5
* Use rqt Message Publisher to set antenna tilt to 1.3
* Use rqt Message Publisher to set antenna pan to 1.0
* Use rqt Message Publisher to set antenna pan to -1.0

Without the PR you will see light intensities on top of the lander flicker between lights. With this PR both light intensities stay with the correct lights.

### DIFF
--- a/ow_gazebo_plugins/src/OWLightControlPlugin/OWLightControlPlugin.cpp
+++ b/ow_gazebo_plugins/src/OWLightControlPlugin/OWLightControlPlugin.cpp
@@ -1,0 +1,82 @@
+// The Notices and Disclaimers for Ocean Worlds Autonomy Testbed for Exploration
+// Research and Simulation can be found in README.md in the root directory of
+// this repository.
+
+#include "OWLightControlPlugin.h"
+#include <gazebo/rendering/rendering.hh>
+
+
+namespace gazebo
+{
+
+GZ_REGISTER_VISUAL_PLUGIN(OWLightControlPlugin)
+
+OWLightControlPlugin::OWLightControlPlugin() :
+  VisualPlugin()
+{
+}
+
+OWLightControlPlugin::~OWLightControlPlugin()
+{
+}
+
+void OWLightControlPlugin::Load(gazebo::rendering::VisualPtr visual, sdf::ElementPtr sdf)
+{
+  m_nodeHandle.reset(new ros::NodeHandle("OWLightControlPlugin"));
+
+  m_lightName.push_back("lander::lander_lights_link::lander_light_light0");
+  m_paramName.push_back("left_light");
+  m_lightIntensity.push_back(1.0);
+  m_lightName.push_back("lander::lander_lights_link::lander_light_light1");
+  m_paramName.push_back("right_light");
+  m_lightIntensity.push_back(1.0);
+
+  for (size_t i = 0; i < m_paramName.size(); i++)
+  {
+    m_nodeHandle->setParam(m_paramName[i], m_lightIntensity[i]);
+  }
+
+  mPreRenderConnection = gazebo::event::Events::ConnectPreRender(
+    boost::bind(&OWLightControlPlugin::onPreRender, this));
+}
+
+void OWLightControlPlugin::onPreRender()
+{
+  double intensity;
+  gazebo::rendering::ScenePtr scene;
+
+  for (size_t i = 0; i < m_paramName.size(); i++) {
+    m_nodeHandle->getParamCached(m_paramName[i], intensity);
+
+    if (m_lightIntensity[i] != intensity)
+    {
+      m_lightIntensity[i] = intensity;
+
+      if (!scene) {
+        scene = gazebo::rendering::get_scene();
+        if (!scene || !scene->Initialized()) {
+          gzwarn << "Gazebo scene not initialized." << std::endl;
+          return;
+        }
+
+        // Helper code for outputting light names
+        /*uint32_t lightCnt = scene->LightCount();
+        for (uint32_t i = 0; i < lightCnt; ++i) {
+          gzmsg << scene->LightByIndex(i)->Name() << std::endl;
+        }*/
+      }
+
+      gazebo::rendering::LightPtr light = scene->LightByName(m_lightName[i]);
+      if (!light) {
+        gzwarn << "Found no light named " << m_lightName[i] << std::endl;
+        continue;
+      }
+
+      intensity = std::clamp(intensity, 0.0, 1.0);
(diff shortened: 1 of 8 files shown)


AstroStucky:
Test works

ow_gazebo_plugins/src/OWLightControlPlugin/OWLightControlPlugin.h
```diff
@@ -0,0 +1,37 @@
...
+#include <ros/ros.h>
+
+
+namespace gazebo {
+  
+class OWLightControlPlugin : public gazebo::VisualPlugin
+{
+public:
+  OWLightControlPlugin();
+  ~OWLightControlPlugin();
```
AstroStucky, line 18: Empty methods in the cpp are unnecessary if you just default them.
```suggestion
  OWLightControlPlugin() = default;
  ~OWLightControlPlugin() = default;
```

ow_gazebo_plugins/src/OWLightControlPlugin/OWLightControlPlugin.cpp
```diff
@@ -0,0 +1,82 @@
...
+{
+}
+
+OWLightControlPlugin::~OWLightControlPlugin()
+{
+}
+
+void OWLightControlPlugin::Load(gazebo::rendering::VisualPtr visual, sdf::ElementPtr sdf)
+{
+  m_nodeHandle.reset(new ros::NodeHandle("OWLightControlPlugin"));
```
AstroStucky, line 25: Is this somehow preferable to `make_unique`?

ow_gazebo_plugins/src/OWLightControlPlugin/OWLightControlPlugin.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+  m_lightName.push_back("lander::lander_lights_link::lander_light_light0");
+  m_paramName.push_back("left_light");
+  m_lightIntensity.push_back(1.0);
```
AstroStucky, line 29: [optional] Could be a member function.

ow_gazebo_plugins/src/OWLightControlPlugin/OWLightControlPlugin.cpp
```diff
@@ -0,0 +1,82 @@
...
+{
+  double intensity;
+  gazebo::rendering::ScenePtr scene;
+
+  for (size_t i = 0; i < m_paramName.size(); i++) {
+    m_nodeHandle->getParamCached(m_paramName[i], intensity);
+
+    if (m_lightIntensity[i] != intensity)
+    {
+      m_lightIntensity[i] = intensity;
```
AstroStucky, line 53: Should this be after the `clamp` call?

ow_lander/materials/scripts/ow_lander.material
```diff
@@ -50,14 +50,14 @@ fragment_program ow_lander_frag glsl
     param_named irradianceMap                  int 3
 
     // lander lights
-    param_named spotlightIntensityScale[0]     float 1.0
-    param_named spotlightIntensityScale[1]     float 1.0
+    // Control status of rover lights (spotlights)
```
AstroStucky, line 53:
```suggestion
    // Control status of lander lights (spotlights)
```

AstroStucky (approved):
Just had some small suggestions and questions for the reviewer that they're welcome to resolve at their own discretion.

Code approved 🎊

ow_gazebo_plugins/src/OWLightControlPlugin/OWLightControlPlugin.cpp
mogumbo: Could be. If I did it more than twice I would probably make it a function.

ow_lander/materials/scripts/ow_lander.material
mogumbo: fixed

ow_gazebo_plugins/src/OWLightControlPlugin/OWLightControlPlugin.cpp
mogumbo: Good idea. Fixed.

ow_gazebo_plugins/src/OWLightControlPlugin/OWLightControlPlugin.h
mogumbo: Changed to `{}` in the header. I believe `default` is intended for trivial types and we don't really need it here.

ow_gazebo_plugins/src/OWLightControlPlugin/OWLightControlPlugin.h
```diff
@@ -0,0 +1,37 @@
...
+
+namespace gazebo {
+  
+class OWLightControlPlugin : public gazebo::VisualPlugin
+{
+public:
+  OWLightControlPlugin() {}
+  ~OWLightControlPlugin() {}
+
+  virtual void Load(gazebo::rendering::VisualPtr visual, sdf::ElementPtr sdf);
```
kmdalal, line 20: Purely stylistic tip: in declarations where the purpose of the parameter is obvious, the argument names can be omitted, for a shorter line and less to read.  E.g you wouldn't do this with a generic type like `int` or `string`.

ow_gazebo_plugins/src/OWLightControlPlugin/OWLightControlPlugin.h
```diff
@@ -0,0 +1,37 @@
...
+#include <gazebo/common/Plugin.hh>
+#include <ros/ros.h>
+
+
+namespace gazebo {
+  
+class OWLightControlPlugin : public gazebo::VisualPlugin
+{
+public:
+  OWLightControlPlugin() {}
```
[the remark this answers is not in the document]
mogumbo: After looking around I see there is a *lot* of information and opinions about trivial vs standard-layout vs POD classes. I don't think understanding all of it is within the realm of this PR. If we want to write style rules about when to use `= default` and `= delete` I think it should be separate work.

PR #330, opened 2023-04-12
Trajectory planner refactor

AstroStucky:
## Linked Issues:
Jira Ticket 🎟️    |    [OCEANWATER-1130](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1130)

## Summary of Changes
* TrajectorySequence class created to handle sequencing of trajectories and make trajectory programming self-documenting.
* Rewrote all arm trajectories to use the new TrajectorySequence. All actions should still behave exactly the same as before. [^2]
* Moved trajectory planning of actions into the action definition and action mixins.
* Created 4 new exception types specific to OceanWATERS.

## Test
Perform [Release 12 Test 4](https://babelfish.arc.nasa.gov/confluence/display/OCEANWATERS/draft+OW-TEST-012-4+ROS+Actions)[^1] starting at the first arm action, **4.10**, and finishing with the final arm action, **4.24**. There is no need to update **Tester #** columns with your results, just report them here.

[^1]: Changes to the behavior of the `--relative` flag on Cartesian arm actions required an updated test.
[^2]: This is for the most part true, but there is a planning error produced by linear path planning that was not handled before and can result in some trajectories that previously worked now being rejected. Test 4 has been updated to avoid this error.


anthonykoutroulis:
There are several changes to variables that only add a second underscore (ie: `_executor` to `-__executor`). Is that a replace error, or is there a reason for that change?

anthonykoutroulis:
Test works and is easy to follow. Code looks good outside of the one suspected find/replace error I noted.

AstroStucky:
> There are several changes in `arm_interface.py` to variables that only add a second underscore (ie: `_executor` to `__executor`). Is that a replace error, or is there a reason for that change?

It's intentional. I am using the `self.__really_private_field` notation in [ROS's Python style guide](http://wiki.ros.org/PyStyleGuide#Coding_Style). I do not believe it has a real effect on the privateness of a member, it's purely notation.

anthonykoutroulis (approved):

mogumbo:
The test steps mostly work. Step 4.15 completed but the arm pose was different than that one pictured in the test. (This is intermittent and I don't know if it was caused by this PR.)
Screenshot_20230419_090746

Step 4.17 succeeded, but with an error (also intermittent, and I don't know if it was caused by this PR).
`[INFO] [1681920590.331938, 1916525174.917500]: ArmFindSurface action started
[ERROR] [1681920596.670658, 1916525181.232500]: Received comm state PREEMPTING when in simple state DONE with SimpleActionClient in NS /arm_controller/follow_joint_trajectory
[INFO] [1681920598.948209, 1916525183.507500]: ArmFindSurface: Succeeded - ArmFindSurface trajectory stopped by a force of 312.98 N. Surface found at (1.467, -0.404, -0.051)
[INFO] [1681920598.964265, 1916525183.542500]: ArmFindSurface action complete`

After arm_stop.py in Step 4.23 I started getting this warning every 2 seconds (also intermittent, and I don't know if it was caused by this PR):
`[ WARN] [1681921412.585759732, 1916525984.222500086]: Power system node computed excessive power input for GSAP, 22.0606W. Capping GSAP input at 20W.`

AstroStucky:
All good catches. These are all intermittent bugs I have experience before these changes, so my best guess is that they are unrelated. 

One already had a ticket, and the other two I have been meaning to make tickets for. I referenced some of your results in the tickets, so thank you for the data collection. The tickets, in order of each bug you encountered are:
- https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1099
- https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1138
- https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1139

mogumbo:
> It's intentional. I am using the `self.__really_private_field` notation in [ROS's Python style guide](http://wiki.ros.org/PyStyleGuide#Coding_Style). I do not believe it has a real effect on the privateness of a member, it's purely notation.

I'm not a fan of this notation. Private should be absolute. Using __really_private implies that _private has been demoted to just a suggestion. But I don't hate it enough to hold up this PR.

ow_lander/src/ow_lander/arm_interface.py
```diff
@@ -171,7 +193,7 @@ def __init__(self):
         self._continue_arm_in_fault = config[
             ArmFaultMonitor.CONTINUE_ARM_IN_FAULT_NAME]
 
-    def _system_faults_callback(self, data):
+    def _system__faults_callback(self, data):
```
mogumbo, line 196: Is your extra underscore in the wrong place here? I thought you might want it at the beginning to make this method "really private."

ow_lander/src/ow_lander/constants.py
```diff
@@ -5,12 +5,12 @@
...
-J_HAND_YAW = 4
-J_DIST_PITCH = 3
-J_PROX_PITCH = 2
-J_SHOU_PITCH = 1
 J_SHOU_YAW = 0
+J_SHOU_PITCH = 1
+J_PROX_PITCH = 2
+J_DIST_PITCH = 3
+J_HAND_YAW = 4
+J_SCOOP_YAW = 5
```
mogumbo, line 13: Should J_GRINDER also be 5? If so, you should probably add a comment to explain why they are the same value.

mogumbo (approved):

ow_lander/src/ow_lander/arm_interface.py
AstroStucky: Oops! This actually is a find-and-replace mistake. Good catch.

ow_lander/src/ow_lander/constants.py
AstroStucky: > I'm not a fan of this notation. Private should be absolute. Using __really_private implies that _private has been demoted to just a suggestion. But I don't hate it enough to hold up this PR.

There is no real way to enforce private members in Python, but `_` and `__` partially serve that function. `_` is purely notionally, and indicates to a programer that this value/method should not be touched outside of the class its in or its subclasses. `__` mangles the name of the member in a way that make it still possible to access, but requires both its name and the name of the class its defined in. 

I have read (not in the ROS Py style guide) that the difference between `_` and `__` class member prefixes are analogous to protected and private members in C++. So anything you expect only a subclass to reference is prefixed with `_`, and anything entirely private to its class should be prefixed `__`. 

I have to admit I haven't been entirely rigorous with this in the ow_lander package.

**Source** https://stackoverflow.com/questions/1301346/what-is-the-meaning-of-single-and-double-underscore-before-an-object-name

AstroStucky:
> I'm not a fan of this notation. Private should be absolute. Using __really_private implies that _private has been demoted to just a suggestion. But I don't hate it enough to hold up this PR.

There is no real way to enforce private members in Python, but `_` and `__` partially serve that function. `_` is purely notionally, and indicates to a programer that this value/method should not be touched outside of the class its in or its subclasses. `__` mangles the name of the member in a way that make it still possible to access, but requires both its name and the name of the class its defined in. 

I have read (not in the ROS Py style guide) that the difference between `_` and `__` class member prefixes are analogous to protected and private members in C++. So anything you expect only a subclass to reference is prefixed with `_`, and anything entirely private to its class should be prefixed `__`. 

I have to admit I haven't been entirely rigorous with this in the ow_lander package.

**Source** https://stackoverflow.com/questions/1301346/what-is-the-meaning-of-single-and-double-underscore-before-an-object-name

ow_lander/src/ow_lander/arm_interface.py
AstroStucky: Fixed

PR #331, opened 2023-04-14
ResetPhysicsState call removed for more consistent scoop motion

AstroStucky:
## Linked Issues:
| EPIC ⚡ | [OCEANWATER-1134](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1134) |
| Jira Ticket 🎟️ | [OCEANWATER-1069](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1069) |

`ResetPhysicsState` was causing the scoop to be pushed towards the lander at the start of a perpendicular **TaskScoopCircular**. It is not clear why this happened, but the intention behind the call was to prevent the scoop from moving backwards as a result of dig forces, and this is already accomplished by simply setting dig forces to zero during backward motion.

Additionally, the backward motion if-statement is entered during non-dig trajectories and even while the scoop is stationary. The former due to any downward arc of the scoop being detected as a dig, and the latter due to floating point precision errors. Messing with the physics of the scoop outside of digging like this could result in non-physical dynamics, so now the if-statement only has an internal effect on the system and does not affect the scoop state except in cases of actual digging (when terrain volume is removed). A 1 mm velocity threshold was also added to avoid the if-statement from triggering due to floating point precision errors. 

## Test
1. Call a `rosrun ow_lander task_grind.py 1.65 0 0.05 .6 False` in atacama_y1a
2. Followed by `rosrun ow_lander task_scoop_circular.py --perpendicular` 
3. Verify the trajectory of **TaskScoopCircular** is normal and does not get pushed off-course towards the lander. Some jittering of the scoop might be observed at the end of the trajectory. This is expected, and is probably a result of the dig force.

### DIFF
--- a/ow_gazebo_plugins/src/BalovnevModelPlugin/BalovnevModelPlugin.cpp
+++ b/ow_gazebo_plugins/src/BalovnevModelPlugin/BalovnevModelPlugin.cpp
@@ -120,7 +120,7 @@ void BalovnevModelPlugin::onUpdate()
   if (link_vel.Dot(SCOOP_FORWARD) < 0.0) {
     // reset force if pushback occurs
     resetForces();
-    m_link->ResetPhysicsStates();
+    m_link->SetLinearVel(ignition::math::Vector3d(0, 0, 0));
   }
   if (m_horizontal_force == 0.0 && m_vertical_force == 0.0)
     return; // no force to apply, early return
--- a/ow_gazebo_plugins/src/BalovnevModelPlugin/BalovnevModelPlugin.cpp
+++ b/ow_gazebo_plugins/src/BalovnevModelPlugin/BalovnevModelPlugin.cpp
@@ -117,10 +117,11 @@ void BalovnevModelPlugin::onUpdate()
   }
   // check for pushback
   auto link_vel = m_link->RelativeLinearVel();
-  if (link_vel.Dot(SCOOP_FORWARD) < 0.0) {
+  double BACKWARD_MOTION_CHECK_TRHESHOLD = 0.001; // meters
+  if (link_vel.Length() > BACKWARD_MOTION_CHECK_TRHESHOLD
+      && link_vel.Dot(SCOOP_FORWARD) < 0.0) {
     // reset force if pushback occurs
     resetForces();
-    m_link->SetLinearVel(ignition::math::Vector3d(0, 0, 0));
   }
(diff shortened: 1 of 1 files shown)


ow_gazebo_plugins/src/BalovnevModelPlugin/BalovnevModelPlugin.cpp
```diff
@@ -117,10 +117,11 @@ void BalovnevModelPlugin::onUpdate()
   }
   // check for pushback
   auto link_vel = m_link->RelativeLinearVel();
-  if (link_vel.Dot(SCOOP_FORWARD) < 0.0) {
+  double BACKWARD_MOTION_CHECK_TRHESHOLD = 0.001; // meters
+  if (link_vel.Length() > BACKWARD_MOTION_CHECK_TRHESHOLD
```
kmdalal, line 121: typo TRHESHOLD

kmdalal:
In the given test, the scoop meets the ground and then the action completes normally.  There was no backward or erratic motion.  Attached screenshot shows final position.  Is this the full intended test and result?

dig-circular

AstroStucky:
> In the given test, the scoop meets the ground and then the action completes normally. There was no backward or erratic motion. Attached screenshot shows final position. Is this the full intended test and result?

Oh whoops! I have been testing with the collision mesh disabled for my own convenience and forgot that testers have to call as **TaskGrind** before calling **TaskScoopCicrular**. The test procedure has been updated.

kmdalal:
I think the test passes now, though I watched the scoop carefully and notice that the ground material continually disappears ahead of where the scoop is actually making contact.  The following end position shows that the trench extends a little further than where the scoop actually made contact.  I'm assuming this is an unrelated issue, and I'd consider it a minor one.

dig-circular png

AstroStucky:
It is unrelated to this PR. It has to do with ow_dynamic_terrain, and as far as I know has been this way since that package first saw the light of day. I do wish to address it if there is time to. Surprisingly, we've never had a ticket for it, so I've just made it a low priority one. https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1135

kmdalal:
And I just confirmed that the behavior I last reported is found in Release 11.

kmdalal (approved):

PR #335, opened 2023-05-17
All dig points now correspond to the center of the trench

AstroStucky:
## Linked Issues:
| EPIC ⚡ | [OCEANWATER-1146](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1146) |
| Jira Ticket 🎟️ | [OCEANWATER-1153](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1153) |

## Summary of Changes
* X and Y for all dig actions (TaskGrind, TaskScoopLinear, and TaskScoopCircular) now correspond to the center of their resulting trench.
  * This allows for X and Y values used for a TaskGrind to be used for either a TaskScoopLinear or TaskScoopCircular without adjustment (see Caveat).
* Default action client script values updated to work with new semantics. 

## Caveat
The `length` value used for a TaskGrind cannot in general be used for a TaskScoopLinear in the same spot. This is because the TaskGrind `length` is the distance between the grinder entry point and its exit point, but the TaskScoopLinear `length` is the distance between the end of it's entry circular trajectory, and the start of the exit circular trajectory. This means a TaskScoopLinear `length` value must be slightly shorter than the TaskGrind length that processed the trench it digs into. Empirically the TaskScoopLinear `length` should be about 3 cm less than the TaskGrind `length`, but can be less if only shallow digs are desired. Default trench lengths have been changed based on this heuristic.

## Test
All of the following should succeed and look nominal.
1.  Launch Atacama_y1a
2. Command a grind `rosrun ow_lander task_grind.py -x 1.7 -y 0.6`
3. Command a linear scoop using the same coordinates `rosrun ow_lander task_scoop_linear.py -x 1.7 -y 0.6`
4. Sample the surface height of mound near the lander 
`rosrun ow_lander arm_find_surface.py -x 1.65 -y -0.4 -z 0.1`
here is my result
```
[INFO:/lander_action_servers] ArmFindSurface: Succeeded - ArmFindSurface trajectory stopped by a force of 206.46 N. Surface found at (1.660, -0.404, -0.070)
```
6. Use the resulting z-value to grind that spot `rosrun ow_lander task_grind.py -x 1.65 -y -0.4 -z -0.07`
7. Perform a circular scoop on that spot `rosrun ow_lander task_scoop_circular.py -x 1.65 -y -0.4 -z -0.07`

### DIFF
--- a/ow_lander/scripts/task_grind.py
+++ b/ow_lander/scripts/task_grind.py
@@ -13,7 +13,7 @@ from ow_lander import constants
 parser = argparse.ArgumentParser(
   formatter_class=argparse.ArgumentDefaultsHelpFormatter,
   description="The grinder is used process the terrain for excavation.")
-parser.add_argument('-x', type=float, default=1.45,
+parser.add_argument('-x', type=float, default=1.75,
   help="X-coordinate of grinding starting point in base_link frame")
 parser.add_argument('-y', type=float, default=0.0,
   help="Y-coordinate of grinding starting point in base_link frame")
--- a/ow_lander/scripts/task_scoop_circular.py
+++ b/ow_lander/scripts/task_scoop_circular.py
@@ -22,7 +22,7 @@ parser.add_argument('--frame', '-f', type=int, default=0,
        + str(constants.FRAME_ID_MAP))
 parser.add_argument('--relative', '-r', action='store_true', default=False,
   help="Position will be interpreted as relative to the current position")
-parser.add_argument('-x', type=float, default=1.65,
+parser.add_argument('-x', type=float, default=1.75,
   help="X-coordinate on surface where trench is centered")
 parser.add_argument('-y', type=float, default=0,
   help="Y-coordinate on surface where trench is centered")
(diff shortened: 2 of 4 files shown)


kmdalal (approved):
Code review was light but the changes looked straightforward and reasonable.

Tests passed, though I had to use a deeper depth in the last dig to see terrain deformation, and I wondered why no regolith was collected (this is out of scope).

Though the new X/Y semantics makes the operations more usable to me, it remains to be seen how well they will work out for our users, especially those also using OWLAT.

AstroStucky:
> Tests passed, though I had to use a deeper depth in the last dig to see terrain deformation, and

I hardly saw any too, but I think that's just because it is a raised mound. I probably could've pick a better spot for the test. 

> I wondered why no regolith was collected (this is out of scope).

We can discuss this further. Pretty much there was not enough terrain deformed to overcome the threshold to spawn a particle. There are ways to improve on this. 
 
> Though the new X/Y semantics makes the operations more usable to me, it remains to be seen how well they will work out for our users, especially those also using OWLAT.

If nothing else I hope this PR demonstrates how simple it is to change trench geometry and parameter semantics on the action definition level should we need to make modifications in the future. I think these semantics are easier to work with than what we had before, and I hope our users will agree when they get a chance to use them. 

Thanks for the review!

PR #336, opened 2023-05-18
Oceanwater 1123 publish goal errors

anthonykoutroulis:
## Linked Issues:
| Jira Ticket 🎟️ | [OCEANWATER-1123](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1123) |

## Summary of Changes
* Added a ros topic to the ActionServerBase class in server.py for publishing the result of inheriting actions in actions.py. ActionServerBase is modified publish the resulting state and action source as they are resolved and their statuses are set in ActionServerBase
* Added an additional field to the relevant actions in actions.py to describe what action "group" they belong to (arm, task, camera, pan_tilt).
* Added a ros message "ActionGoalStatus.msg" to hold action result statuses
* Modified the fault detector node to subscribe to the new topic.
* Added logic to fault detector to set/clear GOAL_ERROR flags from action status message.

### Setup
* Launch any sim, ie: `roslaunch ow atacama_y1a.launch`, and then open 4 additional terminals.
* Terminal 1 - In the first additional terminal, once the sim is running, monitor the topic that publishes the goal states from the action server module: `rostopic echo /action_goal_status`
  * The terminal will warn that no messages are being published, that is expected. Action status messages will publish as actions are made.
* Terminal 2 - In the second terminal, monitor the system fault flags captured by the fault detector module: `rostopic echo /system_faults_status`
  * This will rapidly print the integer value` of the system faults bit flags, 0 for now.
* Terminal 3 - In the third terminal, run any executable (will succeed) lander action, ie: `rosrun ow_lander arm_unstow.py`
## # Run
* The topic in Terminal 1 will update once. The message data will have a `status_list` that is an array of statuses grouped by component type (in order: 0=ARM, 1=TASK, 2=CAMERA, 3=PAN_TILT). The `status` value of each will reflect the resulting status of the action call.
  * The primary values that will be observed in this test are 0 for `Pending`, 3 for `Succeeded`, and 4 for `Aborted`. The other values (see [actionlib_msgs/GoalStatus documentation](https://docs.ros.org/en/noetic/api/actionlib_msgs/html/msg/GoalStatus.html) are supported and should be published, but they are not used by the fault detection and I don't have a test for them.
  * Observe that the status value for the successful action is `3`. The entry should correspond to the enumeration above, but the text field will be updated with debug text to describe the source.
  * The other status values will be 0 for Pending. They will not be populated with debug text until an appropriate action is called.
  * Observe that the value in Terminal 3 is still 0, indicating no system fault detected.
* To test a failed action, either send an action that you know will fail, send an action during another that is using the same component (ie: task_grind during arm_stow), or begin an arm action and use Terminal 4 to interrupt it before completion with `rosrun ow_lander arm_stop.py`
  * Observe that the status of the failed/interrupted action is now `4 `.
  * Observe that the system_faults_status in Terminal 2 has changed to a non-zero value corresponding to the Goal Error Type (2=ARM, 8=TASK, 16=CAMERA, 64=PAN_TILT), or the bitwise representation of a combination of error types.
* Run a successful action of the same component type to clear reported errors.
  * Observe the status value is updated in Terminal 1, and that the system fault bit is cleared in Terminal 2.


ow_lander/src/ow_lander/server.py
```diff
@@ -12,13 +14,23 @@ class ActionServerBase(ABC):
   and logging for all OceanWATERS actions.
   """
 
+  ACTION_GOAL_STATUS_TOPIC = "/action_goal_status"
+  # allocate and initialize the container for goal status msgs
+  _goal_status_array = []
+  for _ in range(owl_msgs.msg.ActionGoalStatus.NUM_GOAL_TYPES):
+      _goal_status_array.append(actionlib_msgs.msg.GoalStatus())
```
anthonykoutroulis, line 21: @AstroStucky I feel like having initialization code outside of __init__ is strange, but I'm not seeing any other python way to enforce the size/structure of the shared class-member array. Thoughts?

AstroStucky:
`ActionGoalStatus` seems more like an internal topic than a user facing one. That said I wonder if it makes more sense to define it in the ow_lander package? 

Would be curious to hear @kmdalal's thoughts on this

ow_lander/src/ow_lander/server.py
AstroStucky: I believe this is fine to do. You wouldn't want to use `__init__` because that initializes and instance, but here you want to initialize a class member, and that should only happen once, ideally when `ActionServerBase` is imported.

This could be shortened to a single line if you'd like.
```Python
  _goal_status_array = [actionlib_msgs.msg.GoalStatus() for _ in range(owl_msgs.msg.ActionGoalStatus.NUM_GOAL_TYPES)
```

ow_faults_detection/src/FaultDetector.cpp
```diff
@@ -72,6 +77,55 @@ void FaultDetector::publishFaultsMessage(pub_t& fault_pub, msg_t fault_msg, flag
...
+        case msg.CAMERA_GOAL:
+          m_system_faults_flags &= ~SystemFaultsStatus::CAMERA_GOAL_ERROR;
+          break;
+        case msg.PAN_TILT_GOAL:
+          m_system_faults_flags &= ~SystemFaultsStatus::PAN_TILT_GOAL_ERROR;
+          break;
+        default:
+          break;
+      } // else: other statuses not currently supported but should be in the future
+    }
```
AstroStucky, line 122:
```suggestion
    // set/exonerate goal errors
    switch (goal_index) {
      case msg.ARM_GOAL:
        update_fault_flag_from_goal_status(SystemFaultsStatus::ARM_GOAL_ERROR,  goal_status);
        break;
      case msg.TASK_GOAL:
        update_fault_flag_from_goal_status(SystemFaultsStatus::TASK_GOAL_ERROR, goal_status);
        break;
      case msg.CAMERA_GOAL:
        update_fault_flag_from_goal_status(SystemFaultsStatus::CAMERA_GOAL_ERROR, goal_status);
        break;
      case msg.PAN_TILT_GOAL:
        update_fault_flag_from_goal_status(SystemFaultsStatus::PAN_TILT_GOAL_ERROR, goal_status);
        break;
      default:
        break;
    }
```
And `update_fault_flag_from_goal_status` could be a private member function FaultDetector 
```C++
static void FaultDetector::update_fault_flag_from_goal_status(const uint64 flag, const uint8 goal_status) {
  if (goal_status == actionlib_msgs::GoalStatus::ABORTED) { // set fault
    m_system_faults_flags |= flag;
  } else if (goal_status ==  actionlib_msgs::GoalStatus::SUCCEEDED) { // exonerate fault 
    m_system_faults_flags &= ~flag;
  }
};

```

AstroStucky (commented):
Tests works exactly as described. Great work!

Just a few simplifications to the code were suggested.

ow_faults_detection/src/FaultDetector.cpp
AstroStucky: (feel free to think up a better function name)

AstroStucky:
_It seems I accidentally left this comment in drafts. It relates directly to a comment on actions.py:87, so sorry if that caused confusion._

The use of `ArmActionServer` is two-fold.
1. Provide OW defined actions with a common base to unify subtleties in action server handling, and
2. Provide users with a base class to define their own actions in a similar manner to our own. 
So I am very glad to see that state reporting to the fault node was made optional by setting `goal_group_id` to `None` as some users may have no interested in hooking their actions up to our own fault reporting. 

That said, I think this ~use~ non-use of `goal_group_id` would self-document if its default value was defined in `ActionServerBase`. The following could go underneath the other properties, and would define a property that is automatically set to `None` if the child class never sets it. 
```Python 
  @property
  def goal_group_id(self):
    return None
```

So this way any child class that does not set `goal_group_id` will have it set to none, and the state publishing code will be automatically disabled for it.

ow_faults_detection/include/ow_faults_detection/FaultDetector.h
```diff
@@ -13,6 +13,7 @@
 #include <std_msgs/Float64.h>
 #include <std_msgs/Empty.h>
 #include <ow_faults_detection/JointStatesFlag.h>
+#include <owl_msgs/ActionGoalStatus.h>
```
kmdalal, line 16: I think this should be moved to `ow_lander` for now, to remain consistent since it's not a unified topic.

ow_lander/src/ow_lander/actions.py
```diff
@@ -839,6 +854,7 @@ class LightSetIntensityServer(ActionServerBase):
   goal_type     = owl_msgs.msg.LightSetIntensityGoal
   feedback_type = owl_msgs.msg.LightSetIntensityFeedback
   result_type   = owl_msgs.msg.LightSetIntensityResult
+  goal_group_id = None
```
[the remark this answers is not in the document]
kmdalal: We'd have to define what goal errors apply to lighting, and how to detect them.  Definitely a separate project, and probably low priority.

ow_faults_detection/src/FaultDetector.cpp
anthonykoutroulis: This reminds me, I actually did this for the camera faults and was considering doing this for the every `system_faults_flag` update, but decided to hold off until all the the fault unification changes were done.

```c++
is_system_fault = actionlib_msgs::GoalStatus::ABORTED == goal_status;
switch (goal_index) {
      case msg.ARM_GOAL:
        update_system_faults_flags(SystemFaultsStatus::ARM_GOAL_ERROR,  is_system_fault);
        break;
...

is_system_fault = msg.value > POWER_THERMAL_MAX;
update_system_faults_flags(SystemFaultsStatus::POWER_EXECUTION_ERROR,  is_system_fault);

```

I'm going to hold off on including this change in this PR, as it can be applied to every component and so the goal status has the same 'bit flipping' style consistency for now. I can do a minor cleanup PR separately. In fact, we still need a release test for faults, so that could work nicely there as a regression test change too.

anthonykoutroulis:
> _It seems I accidentally left this comment in drafts. It relates directly to a comment on actions.py:87, so sorry if that caused confusion._
> 
> The use of `ArmActionServer` is two-fold
> 
>     1. Provide OW defined actions with a common base to unify subtleties in action server handling, and
> 
>     2. Provide users with a base class to define their own actions in a similar manner to our own.
> 
> 
> So I am very glad to see that state reporting to the fault node was made optional by setting `goal_group_id` to `None` as some users may have no interested in hooking their actions up to our own fault reporting.
> 
> That said, I think this ~use~ non-use of `goal_group_id` would self-document if its default value was defined in `ActionServerBase`. The following could go underneath the other properties, and would define a property that is automatically set to `None` if the child class never sets it.
> 
> ```python
>   @property
>   def goal_group_id(self):
>     return None
> ```
> 
> So this way any child class that does not set `goal_group_id` will have it set to none, and the state publishing code will be automatically disabled for it.

Perfect, this solves a lot. @AstroStucky 

It also solves the need retroactive support caught by @kmdalal 

> I'm running `ReferenceMission1` and got the following error on `GuardedMove`:
> 
> ```
> [INFO]: GuardedMove action started
> [INFO]: GuardedMove: Succeeded - Ground detected
> [ERROR]: Exception in your execute callback: 'GuardedMoveServer' object has no attribute 'goal_group_id'
> Traceback (most recent call last):
>   File "/opt/ros/noetic/lib/python3/dist-packages/actionlib/simple_action_server.py", line 289, in executeLoop
>     self.execute_callback(goal)
>   File "/usr/local/home/kdalal/ow/src/ow_simulator/ow_lander/src/ow_lander/server.py", line 133, in __on_action_called
>     self.execute_action(goal)
>   File "/usr/local/home/kdalal/ow/src/ow_simulator/ow_lander/src/ow_lander/actions.py", line 181, in execute_action
>     self._set_succeeded("Ground detected", final=ground_pos, success=True)
>   File "/usr/local/home/kdalal/ow/src/ow_simulator/ow_lander/src/ow_lander/server.py", line 99, in _set_succeeded
>     self.__publish_state(actionlib_msgs.msg.GoalStatus.SUCCEEDED)
>   File "/usr/local/home/kdalal/ow/src/ow_simulator/ow_lander/src/ow_lander/server.py", line 149, in __publish_state
>     if self.goal_group_id is not None:
> AttributeError: 'GuardedMoveServer' object has no attribute 'goal_group_id'
> 
> [ERROR]: To transition to an aborted state, the goal must be in a preempting or active state, it is currently in state: 3
> ```

ow_faults_detection/src/FaultDetector.cpp
AstroStucky: That sounds like a good plan to me. I hadn't bothered to look at how other fault flags were being set/unset before making this comment, but I agree that using the same method for it would be best.

kmdalal (approved):
Tested by running ReferenceMission1 and stopping several arm operations.  The goal error reporting and resetting seemed to work fine.  Nice work!

ow_lander/src/ow_lander/server.py
```diff
@@ -16,9 +16,10 @@ class ActionServerBase(ABC):
 
   ACTION_GOAL_STATUS_TOPIC = "/action_goal_status"
   # allocate and initialize the container for goal status msgs
-  _goal_status_array = []
-  for _ in range(owl_msgs.msg.ActionGoalStatus.NUM_GOAL_TYPES):
-      _goal_status_array.append(actionlib_msgs.msg.GoalStatus())
+  _goal_status_array = [
+    actionlib_msgs.msg.GoalStatus() for _ in range(ow_lander.msg.ActionGoalStatus.NUM_GOAL_TYPES) 
+    ]
```
AstroStucky, line 21:
```suggestion
  ]
```

AstroStucky (approved):
Did a quick retest and all works. All the code looks good. 

There's just one thing that bothered me about the PR on a second look, and that's that I find it weird that stopping an arm movement results in an arm fault flag. It seems to me that ceasing arm movement should be considered normal behavior if arm movement does in fact cease its movement. However, I do not have enough fault handling experience to know for sure whether this makes sense or not. This is approved, but this might be worth discussing. 

Nice work!

PR #341, opened 2023-06-01
Oceanwater 975 limit system faults publish rate

anthonykoutroulis:
## Linked Issues:
| Jira Ticket 🎟️ | [OCEANWATER-975](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-975) |

## Summary of Changes
* Have the Joint Faults module subscribe to the original Joint States message, rather than just the gazebo sim updates.
* Change the /flags/joint_states publishing to happen as Joint States messages are received (same rate), rather than on every update
* Flags are still internally updated at sim time.
* Fault detector unchanged, but now /system_faults_status publishes at the correct 50Hz, corresponding to the rate the lander updates rather than as fast as it can to keep up with gazebo sim ticks (was ~400Hz)

## Test
* Run a scenario, ie: `roslaunch ow atacama_y1a.launch`
* In another terminal, monitor the system faults message content: `rostopic echo /system_faults_status`
* In another terminal, monitor the system faults message rate: `rostopic hz /system_faults_status`
* Observe a fault value of `value: 0` (no faults), and a topic rate of 50Hz
* In the rqt_gui, activate an arm fault: `Dynamic Reconfigure -> faults -> arm_faults`
* Observe the fault value updates correctly: `value: 4`

### DIFF
--- a/ow_faults_injection/include/ow_faults_injection/JointsFaults.h
+++ b/ow_faults_injection/include/ow_faults_injection/JointsFaults.h
@@ -60,6 +60,7 @@ private:
   std::unique_ptr<ros::NodeHandle> m_node_handle;
   ros::CallbackQueue m_callback_queue;
 
+  ros::Subscriber m_joint_states_sub;
   ros::Publisher m_joint_state_flags_pub;
 
   std::vector<unsigned int> m_joint_state_indices;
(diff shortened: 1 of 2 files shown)


kmdalal (approved):
Test passed, looks good!

PR #342, opened 2023-06-22
Oceanwater 1162 camera sync warning

anthonykoutroulis:
## Linked Issues:
| Jira Ticket 🎟️ | [OCEANWATER-1162](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1162) |

## Summary of Changes
The `/StereoCamera/left/camera_info` and `/StereoCamera/left/image_raw` topics were previously not synchronized, resulting in  fault injection only suppressing the image_raw topic and causing a repeated ros warning. Fault injection 
has been updated to also suppress camera_info when during a trigger fault.

## Test
* Launch a sim in a terminal
* Monitor the fault topics in other terminal/s (ie: `rostopic echo /system_faults_status`)
* In rqt, activate the /StereoCamera/left/image_trigger topic
* In rqt's Dynamic Reconfigure pane, under Faults, activate the camera_left_trigger_failure fault.
* Observe, for a minute or two, the console output of the simulation terminal.
  * Observe that the warning does not appear as it normally would.
 
Additionally, changes can be observed by monitoring the various camera topics with `rqt_bag` and observing the `/StereoCamera/left/camera_info` and `/StereoCamera/left/image_raw topics` now are approximately synchronized in time and synchronized in suppression.

### DIFF
--- a/ow_faults_injection/include/ow_faults_injection/FaultInjector.h
+++ b/ow_faults_injection/include/ow_faults_injection/FaultInjector.h
@@ -18,6 +18,7 @@
 #include <ow_lander/lander_joints.h>
 #include <sensor_msgs/JointState.h>
 #include <sensor_msgs/Image.h>
+#include <sensor_msgs/CameraInfo.h>
 #include <geometry_msgs/WrenchStamped.h>
 #include <unordered_map>
 #include <random>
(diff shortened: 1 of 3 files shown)


ow_faults_injection/include/ow_faults_injection/FaultInjector.h
```diff
@@ -80,7 +83,10 @@ class FaultInjector
 
   // Camera
   ros::Subscriber m_camera_raw_sub;
-  ros::Publisher m_camera_trigger_remapped_pub;
+  ros::Subscriber m_camera_info_sub;
```
kmdalal, line 86: Unrelated to this change, but a minor suggestion: since all that is ever done with subscribers is a single subscription, there is no need for individual data members for each.  You can replace all of them with a list of `ros::Subscriber` that you just push into with each subscription.  The subscribers do need to exist, but they can be anonymous.

kmdalal (approved):
Test passed.  Straightforward fix for a very annoying warning!

PR #345, opened 2023-07-06
Create tests for actions, include: arm_find_surface, arm_move_cartesi…

Hao12450:
## Linked Issues:

## Summary of Changes

1. Now all the rostest files have the prepend test_, for example, test_arm_find_surface.test.
2. All the .py file moved to the script folder
3. All the tests are initial call arm_unstow and end with arm_stow

* arm_unstow_action. This test is to check the stow and unstow function work as expected.
* arm_move_cartesian. Several random valid locations were used to test if the arm was able to move as we expected, also as I mentioned in the tag-up meeting, from what I saw in the rqt topic publish monitor, the result and feedback show an incorrect value for the pose, is very different compared to the input value and the pose from the l_scoop_tip (using command rosrun tf tf_echo base_link l_scoop tip, the information using this command is matching the input value of the arm_move_cartesian)
* arm_move_cartesian_guarded. Additional tests for arm moving. The first one is similar to arm_move_cartesian which uses one valid location in the space. The rest try to push or rotate against the ground. The objective is to ascertain whether the arm will cease operation upon experiencing significant force and torque
* arm_find_surface. Test the performance of arm_find_surface action with different parameters. To make each test more independent, after each arm_find_surface action, use arm_move_cartesian to raise the scoop back to the space. This makes more sense to me: if another arm_find_surface is called immediately after the previous one, ros will try to find the shortest path but ignore the ground, which means we anticipate observing the scoop dragging across the ground for a certain distance. Also makes more sense if the scoop is already touching the ground, no need to call the arm_find_action
* arm_move_joint. Control all the joints independently to reach the 'unstow' location.
* pan_tilt_move_cartesian. Move the pan tilt and let it facing to the input vector.
## Test
* rostest ow_sim_tests test_arm_find_surface_action.test
(For other tests, replace 'test_arm_find_surface_action.test' with other tests files name)


AstroStucky:
Please remove the `.vscode` directory. Is there anyway to prevent VSCode from cluttering the repo with its metadata?

Hao12450:
> 

I can put the .vscode in the .gitignore file.

AstroStucky:
> 

We could... I suppose there is nothing wrong with doing that, but I know other people on the team use vscode, and this was never an issue, which is why I think there might be a different way to import the work directory so that the metadata does not wind up there. I don't use vscode myself, so I cannot help here. 

Do whichever you think is best.

kmdalal:
If I may chime in, I'm not familiar with vscode but I don't think its directory should be in the repo -- I've never seen that in a repo.   Is it somehow needed for the build?

Hao12450:
> If I may chime in, I'm not familiar with vscode but I don't think its directory should be in the repo -- I've never seen that in a repo. Is it somehow needed for the build?

The folder contains the workspace setting, I will delete it before I do git add .

AstroStucky:
> The folder contains the workspace setting, I will delete it before I do git add .

There is no need to delete it, and doing so may mess up your VSCode configuration. Instead, for the future, you can stage particular files and directories by passing them as arguments to `git add`. _e.g. `git add ow_sim_tests` will only stage changes in the `ow_sim_tests` directory._ Should you stage something you did not intend to, you can also use `git reset` to unstage a file/directory before committing. _e.g. `git reset .vscode` unstages that directory._

AstroStucky:
As we discussed, there is an error in pan/tilt actions that needs to be fixed before we can test it. [The ticket for that error is here. ](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1181)

For this PR, lets go ahead and remove `test_pan_tilt_move_cartesian.test` and its corresponding python script. You can do this with a `git rm --cached` for the two files followed by a commit. This will preserve the file in your working tree, so you can commit them later, but remove it from the repository.

ow_sim_tests/scripts/test_arm_move_cartesian_guarded.py
```diff
@@ -0,0 +1,207 @@
...
+import rospy
+import roslib
+import unittest
+import ow_lander.msg
+import owl_msgs.msg
+import argparse
+
+from action_testing import *
+from ow_lander import constants
+from geometry_msgs.msg import Vector3, Pose, Quaternion
```
AstroStucky, line 17:
```suggestion
import rospy
import roslib
import unittest
import owl_msgs.msg

from action_testing import *
from ow_lander import constants
from geometry_msgs.msg import Pose, Quaternion
```

ow_sim_tests/scripts/test_arm_move_cartesian_guarded.py
```diff
@@ -0,0 +1,207 @@
...
+    while rospy.get_time() == 0:
+      rospy.sleep(0.1)
+
+  def test_01_arm_unstow(self):
+    arm_unstow_result = test_arm_action(self,
+      'ArmUnstow', owl_msgs.msg.ArmUnstowAction,
+      owl_msgs.msg.ArmUnstowGoal(),
+      TEST_NAME, "test_01_arm_unstow",
+      server_timeout = 50.0 # (seconds) first action call needs longer timeout
+    )
```
AstroStucky, line 40:
```suggestion
```

ow_sim_tests/scripts/test_arm_find_surface.py
```diff
@@ -0,0 +1,201 @@
...
+    while rospy.get_time() == 0:
+      rospy.sleep(0.1)
+
+  def test_01_arm_unstow(self):
+    arm_unstow_result = test_arm_action(self,
+      'ArmUnstow', owl_msgs.msg.ArmUnstowAction,
+      owl_msgs.msg.ArmUnstowGoal(),
+      TEST_NAME, "test_01_arm_unstow",
+      server_timeout = 50.0 # (seconds) first action call needs longer timeout
+    )
```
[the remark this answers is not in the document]
AstroStucky: That `server_timeout = 50.0` line may need to be added to the next test. It always seems to be necessary for the first call to `test_arm_action` or `test_action`.

ow_sim_tests/scripts/test_arm_move_cartesian_guarded.py
```diff
@@ -0,0 +1,207 @@
...
+          position = Point(1.797, 1.277, -0.160),
+          orientation = Quaternion(0.959, -0.284, 0.003, 0.006)
+        ),
+        force_threshold = 200,
+        torque_threshold = 10
+      ),
+      TEST_NAME, 'test_08_arm_move_cartesian_guarded'
+    )
+
+  def test_09_arm_move_cartesian_guarded(self):
```
AstroStucky, line 150: It looked to me like this test was trying to trigger an torque threshold breach, but that did not happen in my testing. You could try lowering the commanded z-coordinate by a little bit, so that the scoop is pushed more and more into the terrain as it moves, or lower the `torque_threshold` parameter.

ow_sim_tests/scripts/test_arm_move_cartesian_guarded.py
```diff
@@ -0,0 +1,207 @@
...
+      ),
+      TEST_NAME, 'test_11_arm_move_cartesian_guarded'
+    )
+
+  def test_99_arm_stow(self):
+    arm_stow_result = test_arm_action(self,
+      'ArmStow', owl_msgs.msg.ArmStowAction,
+      owl_msgs.msg.ArmStowGoal(),
+      TEST_NAME, 'test_99_arm_stow'
+    )
```
AstroStucky, line 203:
```suggestion
```

ow_sim_tests/scripts/test_arm_move_joint.py
```diff
@@ -0,0 +1,110 @@
...
+import rospy
+import roslib
+import unittest
+import ow_lander.msg
+import owl_msgs.msg
+import argparse
+
+from action_testing import *
+from ow_lander import constants
+from geometry_msgs.msg import Vector3, Pose, Quaternion
```
AstroStucky, line 17:
```suggestion
import rospy
import roslib
import unittest
import owl_msgs.msg

from action_testing import *
```

ow_sim_tests/scripts/test_arm_move_joint.py
```diff
@@ -0,0 +1,110 @@
...
+      ),
+      TEST_NAME, 'test_06_arm_move_joint'
+    )
+
+  def test_07_arm_stow(self):
+    arm_stow_result = test_arm_action(self,
+      'ArmStow', owl_msgs.msg.ArmStowAction,
+      owl_msgs.msg.ArmStowGoal(),
+      TEST_NAME, 'test_07_arm_stow'
+    )
```
AstroStucky, line 106:
```suggestion
```

ow_sim_tests/scripts/test_arm_move_joints.py
```diff
@@ -0,0 +1,94 @@
...
+# Research and Simulation can be found in README.md in the root directory of
+# this repository.
+
+import sys
+import rospy
+import roslib
+import unittest
+import ow_lander.msg
+import owl_msgs.msg
+import argparse
```
AstroStucky, line 13:
```suggestion
import rospy
import roslib
import unittest
import owl_msgs.msg
```

ow_sim_tests/scripts/test_arm_move_joints.py
```diff
@@ -0,0 +1,94 @@
...
+    )
+
+  def test_99_arm_move_joints(self):
+    arm_move_joints_result = test_arm_action(self,
+      'ArmMoveJoints', owl_msgs.msg.ArmMoveJointsAction,
+      owl_msgs.msg.ArmMoveJointsGoal(
+        relative = False,
+        angles = [-1.50098, 1.5708, -2.6529, 2.89725, 0, 0]
+      ),
+      TEST_NAME, 'test_99_arm_move_joints'
```
AstroStucky, line 90:
```suggestion
  def test_06_arm_move_joints(self):
    arm_move_joints_result = test_arm_action(self,
      'ArmMoveJoints', owl_msgs.msg.ArmMoveJointsAction,
      owl_msgs.msg.ArmMoveJointsGoal(
        relative = False,
        angles = [-1.50098, 1.5708, -2.6529, 2.89725, 0, 0]
      ),
      TEST_NAME, 'test_06_arm_move_joints'
```

ow_sim_tests/scripts/test_arm_unstow.py
```diff
@@ -0,0 +1,49 @@
...
+# Research and Simulation can be found in README.md in the root directory of
+# this repository.
+
+import sys
+import rospy
+import roslib
+import unittest
+import ow_lander.msg
+import owl_msgs.msg
+import argparse
```
AstroStucky, line 13:
```suggestion
import rospy
import roslib
import unittest
import owl_msgs.msg
```

AstroStucky (changes requested):
All test units of all tests passed; however, there is a reason for this. I do not believe there is a way for them to fail at the moment. This is because `test_arm_action` and `test_action` do not check for an action's success, and I really think that it should. I'm going to commit a change to that module so that it does, and then test units will fail if the action fails. 

For the most part I like the range of parameters being provided to actions. I have suggestions for specific tests to cover more parameters.

### test_arm_find_surface.py
 - Add at least one unit that drives into the ground at an angle by using a normal that is not (0, 0, -1).
 - Add at least one unit where the scoop fails to find the surface. The action should still succeed, but the result message will be different. 

### test_arm_move_cartesian.py
 - test_03 fails, but that does not cause a test unit failure. This will be fixed by a change to `action_testing.py` I plan to make on this branch. Please stand by for it. I do suggest you pick different parameters for test_03 so that it does not fail. 
 - Add at least one call that uses `frame = 1`.

### test_arm_move_guarded.py
 - test_09 I believe is supposed to be stopped by a torque breach, but it is not. I left a code comment about this too. 
 - Add at least one call that uses `frame = 1`.

### test_arm_move_joint.py
 - Add at least 2 test units where `relative = True`.

Really good work so far! Just a little bit more is needed to get this across the finish line.

ow_sim_tests/scripts/test_arm_find_surface.py
```diff
@@ -0,0 +1,201 @@
...
+# Research and Simulation can be found in README.md in the root directory of
+# this repository.
+
+import sys
+import rospy
+import roslib
+import unittest
+import ow_lander.msg
+import owl_msgs.msg
+import argparse
```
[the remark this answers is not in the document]
Hao12450: The vscode will make the fonts a little darker if the import not use, so it easy fix for me. Also, I find out one thing, since we do 'from action_testing import * ' do you think it also is nice if I clean some of the duplicate import? For example import rospy is not necessary for all the unit test files because when we do 'from action_testing import * ' it will do the 'import rospy'

ow_sim_tests/scripts/test_arm_move_cartesian_guarded.py
Hao12450: Ye, not sure why but looks like every time I run the same test, the force and torque are not guaranteed to be the same value, I am going to run some more tests on this.

ow_sim_tests/scripts/test_arm_find_surface.py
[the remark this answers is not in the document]
AstroStucky: > The vscode will make the fonts a little darker if the import not use, so it easy fix for me. Also, I find out one thing, since we do 'from action_testing import * ' do you think it also is nice if I clean some of the duplicate import? For example import rospy is not necessary for all the unit test files because when we do 'from action_testing import * ' it will do the 'import rospy'

Hmmm, that's a good point. I would rather we keep `from action_testing import *` and the redundant imports for now. Using `*` in import statements is actually discouraged in the Python world, but I use it here simply to make lines shorter. When I make it a more explicit import statement, _e.g._ `import action_testing` then suddenly the files will require those other import statements again. The small refactor of `action_testing.py` that I have in mind will use less global scope, so I plan to make it an explicit import then. 

So to summarize, please keep the redundant imports for now.

AstroStucky:
Please remove the `.vscode` directory from this branch.

AstroStucky:
I have added a feature to action_testing.py that checks for the expected end state of an action. This defaults to zero, but the `expected_end_state` parameter can be provided to, for example, test an action call that is expected to fail. Any actions that abort that weren't supposed to abort will not cause a failure of that test unit.

AstroStucky:
Going to look through your changes, and provide my review now

ow_sim_tests/scripts/test_arm_find_surface.py
```diff
@@ -0,0 +1,217 @@
...
+        frame = 0,
+        relative = True,
+        pose= Pose(
+          position = Point(0, 0, 0.2),
+          orientation = Quaternion(0, 0, 0, 1)
+        )
+      ),
+      TEST_NAME, 'test_02_arm_move_cartesian'
+      # The arm_move_cartesian after every arm_find_surface is to get the arm off the ground
+    )
```
AstroStucky, line 60:
```suggestion
   # Uisng arm_move_cartesian after every arm_find_surface to get the arm off the ground
  def test_02_arm_move_cartesian(self):
    arm_move_cartesian_result = test_arm_action(self,
      'ArmMoveCartesian', owl_msgs.msg.ArmMoveCartesianAction,
      owl_msgs.msg.ArmMoveCartesianGoal(
        frame = 0,
        relative = True,
        pose= Pose(
          position = Point(0, 0, 0.2),
          orientation = Quaternion(0, 0, 0, 1)
        )
      ),
      TEST_NAME, 'test_02_arm_move_cartesian'
    )
```

AstroStucky: The diff makes it hard to see, but I simply moved the comment to the top of the test unit, and did some minor rewording

AstroStucky (changes requested):
Mostly minor corrections. The fact that we have to provide the name of the test unit in string form led to some missed numbering changes. I hope to remove these after a small rework of `action_testing.py` but for now I would still like to keep them consistent (at the same time nothing breaks if we don't). 

Hoping to see just one `relative = True` test unit in `test_arm_move_joints.py`.

Will do a retest of everything once I see those changes. Thanks!

ow_sim_tests/test/test_camera_capture.test
```diff
@@ -4,12 +4,12 @@
     <arg name="gzclient" default="true" />
 
     <include file="$(find ow)/launch/europa_terminator_workspace.launch">
-        <arg name="rqt_gui" value="false" />
+        <arg name="rqt_gui" value="true" />
```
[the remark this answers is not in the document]
Hao12450: Yes, I am interested, do you want to talk on the next dev day?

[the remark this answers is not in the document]
AstroStucky: Sure thing. Dev day would be perfect because I would like to hear other team members weigh in on it.

AstroStucky (approved):
Tests all work, and I really like the range of parameters their units test. Excellent work!

I did leave 2 minor code corrections that I think were just a misunderstanding. Please correct them. I do not feel a retest is necessary at least from myself after those corrections are made.

Thanks for your hard work!

PR #346, opened 2023-07-11
HOTFIX ArmFindSurface was not beginning its approach a distance away from surface

AstroStucky:
## Linked Issues:
Jira Ticket 🎟️ | [OCEANWATER-1183](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1183)

## Summary of Changes
Previously the scoop would initiate its approach towards the surface at the provided **position**. It was never intended to work this way, it should instead begin its approach a **distance** along the anti-normal direction away from **position**. This change makes it so. 

There may be a separate PR for merging this back into noetic-devel because the code has changed enough between the two branches that these changes will unfortunately not carry over directly.

## Test
This is a hotfix, so don't forget to have ow_simulator set to this branch, and all the other repos set to `master`.
1. In atacama_y1a `rosrun ow_lander arm_find_surface.py -x 1.6 -y -0.2 -z -0.17 --distance 0.2`.
     The scoop should begin its approach towards the surface approximately 20 cm above the surface.
2. `rosrun ow_lander arm_find_surface.py -x 1.6 -y -0.2 -z -0.17 --distance 0.4`
     The scoop should now begin its approach towards the surface twice as far from the surface as the previous command.

### DIFF
--- a/ow_lander/src/ow_lander/actions.py
+++ b/ow_lander/src/ow_lander/actions.py
@@ -405,22 +405,22 @@ class ArmFindSurfaceServer(mixins.FrameMixin, mixins.ArmActionMixin,
     # orient scoop so that the bottom points in the opposite to the normal
     # NOTE: regardless of frame parameter orientation is in the base_link frame
     orientation = math3d.quaternion_rotation_between(SCOOP_DOWNWARD, normal)
-    start = goal.position
+    estimated_surface_pos = goal.position
     if relative:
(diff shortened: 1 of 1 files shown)


kmdalal (approved):
Test passed, and I'm assuming one test is sufficient.   The code change seems reasonable.

Hao12450 (approved):
All tests passed, and I did one more test using the parameters that wrote in my unit test.

PR #347, opened 2023-07-15
Oceanwater 1175 additional rostests

Hao12450:
## Linked Issues:

## Summary of Changes

* Additional tests for actions. 
* Some errors observed during testing: 
   - Warning in ros topic [ WARN] [1689639620.134610941, 1916525242.934999943]: The input topic '/StereoCamera/left/camera_info' is not yet advertised
   - CameraCapture aborted error (see OW-1186)
* Please let me know if any additional tests are still required to develop. I assumed for this branch all the actions have it own rostest now, except the pan_tile_move actions.

## Test
* rostest ow_sim_tests test_task_deliver_sample.test
* rostest ow_sim_tests test_task_discard_sample.test
* rostest ow_sim_tests test_task_grind.test
* rostest ow_sim_tests test_task_scoop_circular.test
* rostest ow_sim_tests test_task_scoop_linear.test
* rostest ow_sim_tests test_camera_set_exposure.test
* rostest ow_sim_tests test_light_set_intensity.test

### DIFF
--- a/ow_sim_tests/scripts/test_camera_set_exposure.py
+++ b/ow_sim_tests/scripts/test_camera_set_exposure.py
@@ -25,12 +25,14 @@ class TestCameraSetExposure(unittest.TestCase):
     # proceed with test only when ros clock has been initialized
     while rospy.get_time() == 0:
       rospy.sleep(0.1)
-  
+      # NOTE: additional sleep is required due to CameraCapture failing if called too early (see OW-1186)
+    rospy.sleep(10.0)
+    
   def test_01_camera_set_exposure(self):
     camera_set_exposure_result = test_action(self,
       'CameraSetExposure', owl_msgs.msg.CameraSetExposureAction,
       owl_msgs.msg.CameraSetExposureGoal(
-        automatic = True,
+        automatic = False,
         exposure = -1
       ),
       TEST_NAME, 'test_01_camera_set_exposure',
@@ -54,6 +56,7 @@ class TestCameraSetExposure(unittest.TestCase):
       owl_msgs.msg.CameraCaptureGoal(),
       TEST_NAME, 'test_03_camera_capture'
     )
+    rospy.sleep(3.0)
 
   def test_04_camera_set_exposure(self):
     camera_set_exposure_result = test_action(self,
@@ -71,6 +74,7 @@ class TestCameraSetExposure(unittest.TestCase):
       owl_msgs.msg.CameraCaptureGoal(),
       TEST_NAME, 'test_05_camera_capture'
     )
+    rospy.sleep(3.0)
 
   def test_06_camera_set_exposure(self):
     camera_set_exposure_result = test_action(self,
@@ -88,6 +92,7 @@ class TestCameraSetExposure(unittest.TestCase):
       owl_msgs.msg.CameraCaptureGoal(),
       TEST_NAME, 'test_07_camera_capture'
     )
+    rospy.sleep(3.0)
 
   def test_08_camera_set_exposure(self):
     camera_set_exposure_result = test_action(self,
@@ -105,6 +110,7 @@ class TestCameraSetExposure(unittest.TestCase):
       owl_msgs.msg.CameraCaptureGoal(),
       TEST_NAME, 'test_09_camera_capture'
     )
+    rospy.sleep(3.0)
 
   def test_10_camera_set_exposure(self):
     camera_set_exposure_result = test_action(self,
@@ -115,6 +121,14 @@ class TestCameraSetExposure(unittest.TestCase):
       ),
       TEST_NAME, 'test_10_camera_set_exposure'
     )
+
+  def test_11_camera_capture(self):
+    camera_capture_result = test_action(self,
+      'CameraCapture', owl_msgs.msg.CameraCaptureAction,
+      owl_msgs.msg.CameraCaptureGoal(),
+      TEST_NAME, 'test_11_camera_capture'
+    )
+    rospy.sleep(3.0)
 
 if __name__ == '__main__':
   import rostest
--- a/ow_sim_tests/scripts/test_camera_set_exposure.py
+++ b/ow_sim_tests/scripts/test_camera_set_exposure.py
@@ -1,0 +1,121 @@
+#!/usr/bin/env python
+
+# The Notices and Disclaimers for Ocean Worlds Autonomy Testbed for Exploration
+# Research and Simulation can be found in README.md in the root directory of
+# this repository.
+
+import rospy
+import roslib
+import unittest
+import owl_msgs.msg
+
+from action_testing import *
+
+PKG = 'ow_sim_tests'
+TEST_NAME = 'camera_set_exposure'
+roslib.load_manifest(PKG)
+
+class TestCameraSetExposure(unittest.TestCase):
+
+  @classmethod
+  def setUpClass(cls):
+    rospy.init_node("test_camera_set_exposure")
+
+    set_ignore_action_checks(True)
+    # proceed with test only when ros clock has been initialized
+    while rospy.get_time() == 0:
+      rospy.sleep(0.1)
+  
+  def test_01_camera_set_exposure(self):
+    camera_set_exposure_result = test_action(self,
+      'CameraSetExposure', owl_msgs.msg.CameraSetExposureAction,
+      owl_msgs.msg.CameraSetExposureGoal(
+        automatic = True,
+        exposure = -1
+      ),
+      TEST_NAME, 'test_01_camera_set_exposure',
+      server_timeout = 50.0 # (seconds) first action call needs longer timeout
+    )
+
+  # Move the antenna so the camera can see the robot arm.
+  def test_02_pan_tilt_move_cartesian(self):
+    pan_tilt_move_cartesian_result = test_action(self,
+      'PanTiltMoveCartesian', owl_msgs.msg.PanTiltMoveCartesianAction,
+      owl_msgs.msg.PanTiltMoveCartesianGoal(
+        frame = 1,
+        point = Point(0, 0, 0)
+      ),
+      TEST_NAME, "test_02_pan_tilt_move_cartesian"
+    )
+  
+  def test_03_camera_capture(self):
+    camera_capture_result = test_action(self,
+      'CameraCapture', owl_msgs.msg.CameraCaptureAction,
+      owl_msgs.msg.CameraCaptureGoal(),
+      TEST_NAME, 'test_03_camera_capture'
+    )
+
+  def test_04_camera_set_exposure(self):
+    camera_set_exposure_result = test_action(self,
+      'CameraSetExposure', owl_msgs.msg.CameraSetExposureAction,
+      owl_msgs.msg.CameraSetExposureGoal(
+        automatic = False,
+        exposure = 0.01
+      ),
+      TEST_NAME, 'test_04_camera_set_exposure'
+    )
+
+  def test_05_camera_capture(self):
+    camera_capture_result = test_action(self,
+      'CameraCapture', owl_msgs.msg.CameraCaptureAction,
+      owl_msgs.msg.CameraCaptureGoal(),
+      TEST_NAME, 'test_05_camera_capture'
+    )
+
+  def test_06_camera_set_exposure(self):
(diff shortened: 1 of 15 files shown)


AstroStucky:
## test_task_discard_sample.test
I would like to see the default action client script parameters used for at least one of the test units. You can see what the defaults are with `rosrun ow_lander task_discard_sample.py --help` while the simulator is running. 

## test_camera_set_exposure.test
I do not see every image show in the RQT GUI before the next image is captured in the test. Sometimes no images show at all. This could simply be a problem with rqt's update rate. Seeing as we do not have an automated way to verify images it takes currently, i think it makes sense to put `rospy.sleep(3.0)` or some similar time at the end of each camera capture test unit. Curious if you have differing thoughts on this @kmdalal ?

I did have one run of the test where the PanTiltMoveCartesian failed to move the antenna mast at all until it timed out, but I haven't been able to replicate this result.

ow_sim_tests/scripts/test_camera_set_exposure.py
```diff
@@ -0,0 +1,121 @@
...
+  def test_10_camera_set_exposure(self):
+    camera_set_exposure_result = test_action(self,
+      'CameraSetExposure', owl_msgs.msg.CameraSetExposureAction,
+      owl_msgs.msg.CameraSetExposureGoal(
+        automatic = False,
+        exposure = 0.05
+      ),
+      TEST_NAME, 'test_10_camera_set_exposure'
+    )
+
```
AstroStucky, line 118: You set the exposure in test_10, why not capture a picture for test_11 to finish off the test?

ow_sim_tests/scripts/test_light_set_intensity.py
```diff
@@ -0,0 +1,143 @@
...
+  def test_11_light_set_intensity(self):
+    light_set_intensity_result = test_action(self,
+      'LightSetIntensity', owl_msgs.msg.LightSetIntensityAction,
+      owl_msgs.msg.LightSetIntensityGoal(
+        name = 'right',
+        intensity = 1.0
+      ),
+      TEST_NAME, 'test_11_light_set_intensity'
+    )
+
```
AstroStucky, line 140: Please add two more tests here that both turn off the lights by setting their intensity values to `0.0`

ow_sim_tests/scripts/test_task_scoop_linear.py
```diff
@@ -0,0 +1,140 @@
...
+
+  @classmethod
+  def setUpClass(cls):
+    rospy.init_node("test_task_scoop_linear")
+
+    set_ignore_action_checks(True)
+    # proceed with test only when ros clock has been initialized
+    while rospy.get_time() == 0:
+      rospy.sleep(0.1)
+
```
AstroStucky, line 29: I liked how `test_task_scoop_circular.py` started with scooping above the terrain. Can we do that here too?

ow_sim_tests/test/test_camera_set_exposure.test
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0"?>
+<launch>
+
+    <arg name="gzclient" default="true" />
+
+    <include file="$(find ow)/launch/europa_terminator_workspace.launch">
+        <arg name="rqt_gui" value="true" />
```
AstroStucky, line 7: It just occurred to me, for the purpose of server-side automated testing it will be important to make rqt_gui a command line argument.
```suggestion
    <arg name="rqt_gui" defaullt="true" />
    
    <include file="$(find ow)/launch/europa_terminator_workspace.launch">
        <arg name="rqt_gui" value="$(arg rqt_gui)" />
```
Please make the same change for `test_camera_capture.test`.

ow_sim_tests/scripts/test_task_discard_sample.py
```diff
@@ -0,0 +1,117 @@
...
+      'TaskDiscardSample', owl_msgs.msg.TaskDiscardSampleAction,
+      owl_msgs.msg.TaskDiscardSampleGoal(
+        frame = 1,
+        relative = False,
+        point = Point(0.1, -0.1, 0.1),
+        height = 0.2
+      ),
+      TEST_NAME, "test_07_task_discard_sample"
+    )
+
```
AstroStucky, line 114: As I mentioned in my thread comment, lets have one more test at the end here that uses the default command values.

AstroStucky (changes requested):
Tests all work for the most part, see my most recent comment for details on that. 

Requesting just a few more test units to be added to some of the tests. Otherwise, these look great, and cover a good range of parameters. Nice work!

ow_sim_tests/scripts/test_camera_set_exposure.py
Hao12450: Oh, I forgot to add one more camera_capture at the end, thank you for pointing it out.

Hao12450:
>

ow_sim_tests/test/test_camera_set_exposure.test
Hao12450: Can I know why this change is important?

ow_sim_tests/scripts/test_camera_set_exposure.py
```diff
@@ -0,0 +1,121 @@
...
+    set_ignore_action_checks(True)
+    # proceed with test only when ros clock has been initialized
+    while rospy.get_time() == 0:
+      rospy.sleep(0.1)
+  
+  def test_01_camera_set_exposure(self):
+    camera_set_exposure_result = test_action(self,
+      'CameraSetExposure', owl_msgs.msg.CameraSetExposureAction,
+      owl_msgs.msg.CameraSetExposureGoal(
+        automatic = True,
```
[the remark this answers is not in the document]
kmdalal: Actually, the `automatic` field is completely ignored, in both the PLEXIL interface and the action script.  It shouldn't make a difference what it is set to, though I agree False is the proper value.  I think we should put some comments in the implementation that this field is ignored (as a separate task).

ow_sim_tests/scripts/test_task_deliver_sample.py
```diff
@@ -0,0 +1,76 @@
...
+  @classmethod
+  def setUpClass(cls):
+    rospy.init_node("test_task_deliver_sample")
+
+    set_ignore_action_checks(True)
+    # proceed with test only when ros clock has been initialized
+    while rospy.get_time() == 0:
+      rospy.sleep(0.1)
+
+  def test_01_task_deliver_sample(self):
```
kmdalal, line 30: I find it odd that a deliver is attempted from the stowed position -- this would never occur in practice.  Is this test useful, @AstroStucky ?

kmdalal:
Regarding @AstroStucky's comment:

> I do not see every image show in the RQT GUI before the next image is captured in the test. Sometimes no images show at all. This could simply be a problem with rqt's update rate. Seeing as we do not have an automated way to verify images it takes currently, i think it makes sense to put `rospy.sleep(3.0)` or some similar time at the end of each camera capture test unit. Curious if you have differing thoughts on this @kmdalal ?

I got the same behavior and agree that adding a pause after each image could resolve this issue and make it easier to verify the test.  This strikes me as a test that, while automated in execution, requires a human for full verification.

kmdalal:
All the tests except `test_camera_set_exposure` passed for me.  This one failed twice in a row, the same way.  I've attached a transcript.
[test_camera_set_exposure.log](https://github.com/nasa/ow_simulator/files/12101739/test_camera_set_exposure.log)

ow_sim_tests/scripts/test_light_set_intensity.py
kmdalal: I'd like to see a delay between each intensity setting so that one can see it change.  The test ran so fast that all I observed was one, high, light intensity.

ow_sim_tests/scripts/test_task_scoop_linear.py
kmdalal: Outside the scope of this PR, but the linear scoop looks much like a circular one to me, at least in that most of the dig seems to be complete in the circular and not linear movement.  I recall linear and circular looking more different in the past, but maybe it's the parameter values in these tests.

Hao12450:
> All the tests except `test_camera_set_exposure` passed for me. This one failed twice in a row, the same way. I've attached a transcript. [test_camera_set_exposure.log](https://github.com/nasa/ow_simulator/files/12101739/test_camera_set_exposure.log)

I believe is because the camera_set_exposure call too soon before all ros simulation startup (see OW-1186 with similar bug on CameraCapture), I will add an addition delay before the first test is called.

ow_sim_tests/scripts/test_task_scoop_linear.py
Hao12450: > 

I can change the 'length' parameter to make it more different.

ow_sim_tests/scripts/test_camera_set_exposure.py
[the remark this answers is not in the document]
AstroStucky: Sending `automatic = true` to  `CameraSetExposure`'s action server actually results in an action aborted.
```
[ERROR:/lander_action_servers] CameraSetExposure: Aborted - CameraSetExposure action does not support autoexposure.
```
Perhaps it ought not work this way. Mind you it is impossible to get this result by calling the action client script because it simply does not expose the option, it is perfectly possible to arrive at this result by calling the action from C++ code, Python code, or by publishing directly to the topic `CameraSetExposure/goal`.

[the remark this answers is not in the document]
AstroStucky: This is why the `test_camera_set_exposure.test` fails.

ow_sim_tests/scripts/test_task_scoop_linear.py
AstroStucky: Honestly I wouldn't bother. The thing is both of these tests will have to be at least slightly modified after I merge in the branch `OCEANWATER-1146_new_dig_model_integration`. This is because I rewrote the path planning for **TaskScoopLinear**, **TaskScoopCircular**, and **TaskGrind**  in that branch and as a result parameters will have to vary slightly from what we previously used. Thankfully I don't anticipate this to be a whole lot of work. 

It's worth noting that in the "new dig model" the shape of dig trajectories are invariant to their position in the workspace; however, as they are in `noetic-devel` you cannot claim that because there are some movements that command joint position instead of end-effector pose.

ow_sim_tests/test/test_camera_set_exposure.test
AstroStucky: Code test servers will run these tests "headless" meaning there is no display connected to the server, and even if there were a display connected there will be no human to observe it, so spending computation overhead on rendering and GUI logic would be a waste. Wasting computation is one thing, but many program will just crash if they attempt to use a display that is not there, so allowing these rostests to run entirely headless is a must to enable them to run on test servers.

ow_sim_tests/scripts/test_task_deliver_sample.py
AstroStucky: Some of the bugs I found were found from calling actions immediately at simulator start-up, so I asked Hao to always call the action being tested at the very start of its test. Basically it has been handy at uncovering simulation start-up timing errors, like a topic being subscribed to before it is published. Fixing these can only make our code more rigorous.

AstroStucky:
> > All the tests except `test_camera_set_exposure` passed for me. This one failed twice in a row, the same way. I've attached a transcript. [test_camera_set_exposure.log](https://github.com/nasa/ow_simulator/files/12101739/test_camera_set_exposure.log)
> 
> I believe is because the camera_set_exposure call too soon before all ros simulation startup (see OW-1186 with similar bug on CameraCapture), I will add an addition delay before the first test is called.

Hmmm... Unrelated, but I believe I need to convey the aborted message through the `action_testing.py` module, because I can see now that it's not at all clear why the action has failed unless you scroll all the way up in the log.
```
[ERROR:/lander_action_servers] CameraSetExposure: Aborted - CameraSetExposure action does not support autoexposure.
```
See my code comment about this. It is not because it's being called too soon.

kmdalal (approved):
I only retested `test_camera_set_exposure` and it now passes and visually looks good.  I believe the other tests have been sufficiently tried, and I have no further code comments.  Great work!

PR #350, opened 2023-07-25
Oceanwater 1191 fix hole orientation and position

mogumbo:
## Linked Issues:
| Jira Ticket 🎟️ | [OCEANWATER-1191](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1191) |

## Summary of Changes
* Fixed orientation and position problems with dynamic terrain holes.

## Test
* There's not a good way to test this without extra changes to other files. Instead, here are some before and after screenshots:

Before/after positioning changes. Positions of holes should be points on a regular grid.
Screenshot_20230725_111311
Screenshot_20230725_133531

Before/after flip across x-axis. I patched the terrain with europa_test_dem since you can see its proper orientation using `roslaunch ow europa_test_dem.launch`
Screenshot_20230725_111057
Screenshot_20230725_113738

### DIFF
--- a/ow_dynamic_terrain/src/TerrainModifier.cpp
+++ b/ow_dynamic_terrain/src/TerrainModifier.cpp
@@ -243,7 +243,6 @@ bool TerrainModifier::applyImageToHeightmap(Heightmap* heightmap, const cv::Poin
     return false;
   }
 
-  int heightmap_size = static_cast<int>(terrain->getSize());
   int left = center.x - image.cols / 2;
   int top = center.y - image.rows / 2;
   int right = left + image.cols - 1;
--- a/ow_dynamic_terrain/src/TerrainModifier.cpp
+++ b/ow_dynamic_terrain/src/TerrainModifier.cpp
@@ -257,7 +257,9 @@ bool TerrainModifier::applyImageToHeightmap(Heightmap* heightmap, const cv::Poin
     {
       // Using bottom - y instead of y - top because cv::Mat has a different
       // layout than Ogre Heightmaps.
-      auto pixel_value = image.at<float>(bottom - y, x - left);
+      auto cv_index_y = bottom - y;
+      auto cv_index_x = x - left;
+      auto pixel_value = image.at<float>(cv_index_y, cv_index_x);
       if (skip_zeros && ignition::math::equal<float>(pixel_value, 0.0f))
         continue;
 
@@ -268,7 +270,7 @@ bool TerrainModifier::applyImageToHeightmap(Heightmap* heightmap, const cv::Poin
         continue; // no change is necessary
 
       set_height_value(x, y, new_height);
-      diff.at<float>(y - top, x - left) = new_height - old_height;
+      diff.at<float>(cv_index_y, cv_index_x) = new_height - old_height;
       
       // if we make it here, flag that a change has occurred
       change_occurred = true;
--- a/ow_dynamic_terrain/src/TerrainModifier.cpp
+++ b/ow_dynamic_terrain/src/TerrainModifier.cpp
@@ -243,29 +243,31 @@ bool TerrainModifier::applyImageToHeightmap(Heightmap* heightmap, const cv::Poin
     return false;
   }
 
-  auto heightmap_size = static_cast<int>(terrain->getSize());
-  auto left = max(center.x - image.cols / 2, 0);
-  auto top = max(center.y - image.rows / 2, 0);
-  auto right = min(center.x - image.cols / 2 + image.cols - 1, heightmap_size);
-  auto bottom = min(center.y - image.rows / 2 + image.rows - 1, heightmap_size);
+  int heightmap_size = static_cast<int>(terrain->getSize());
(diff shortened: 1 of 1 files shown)


ow_dynamic_terrain/src/TerrainModifier.cpp
```diff
@@ -243,29 +243,31 @@ bool TerrainModifier::applyImageToHeightmap(Heightmap* heightmap, const cv::Poin
     return false;
   }
 
-  auto heightmap_size = static_cast<int>(terrain->getSize());
-  auto left = max(center.x - image.cols / 2, 0);
-  auto top = max(center.y - image.rows / 2, 0);
-  auto right = min(center.x - image.cols / 2 + image.cols - 1, heightmap_size);
-  auto bottom = min(center.y - image.rows / 2 + image.rows - 1, heightmap_size);
+  int heightmap_size = static_cast<int>(terrain->getSize());
```
kmdalal, line 246: heightmap_size is not used and so it seems this line can be deleted.

ow_dynamic_terrain/src/TerrainModifier.cpp
```diff
@@ -243,29 +243,31 @@ bool TerrainModifier::applyImageToHeightmap(Heightmap* heightmap, const cv::Poin
...
+  int left = center.x - image.cols / 2;
+  int top = center.y - image.rows / 2;
+  int right = left + image.cols - 1;
+  int bottom = top + image.rows - 1;
```
kmdalal, line 250: This is the first time I've seen changing `auto` to a specific type fix a problem, though I can see how it would in this case.

There are many uses of `auto` earlier in the file.  Should some of those be converted too?

ow_dynamic_terrain/src/TerrainModifier.cpp
mogumbo: Thanks, fixed. I meant to remove that and messed it up somehow.

ow_dynamic_terrain/src/TerrainModifier.cpp
mogumbo: To be fair, changing auto to int didn't help but it was one of the things I tried. The changes to the math are what fixed the problem. I left the changes from auto to int because int is easier to type and makes the code easier to understand. I agree that auto is overused in this file.

kmdalal (approved):

ow_dynamic_terrain/src/TerrainModifier.cpp
```diff
@@ -243,29 +243,30 @@ bool TerrainModifier::applyImageToHeightmap(Heightmap* heightmap, const cv::Poin
...
 
-  for (auto y = top; y <= bottom; ++y)
-    for (auto x = left; x <= right; ++x)
+  for (int y = top; y <= bottom; ++y)
+    for (int x = left; x <= right; ++x)
     {
-      auto pixel_value = image.at<float>(y - top, x - left);
+      // Using bottom - y instead of y - top because cv::Mat has a different
+      // layout than Ogre Heightmaps.
+      auto pixel_value = image.at<float>(bottom - y, x - left);
```
AstroStucky, line 260: The same calculation is used later for the diff image, so it might make sense to make them variables. 
```suggestion
      // Using bottom - y instead of y - top because cv::Mat has a different
      // layout than Ogre Heightmaps.
      auto iy = bottom - y;
      auto ix = x - left;
      auto pixel_value = image.at<float>(iy, ix);
```

ow_dynamic_terrain/src/TerrainModifier.cpp
```diff
@@ -243,29 +243,30 @@ bool TerrainModifier::applyImageToHeightmap(Heightmap* heightmap, const cv::Poin
...
+
       auto old_height = get_height_value(x, y);
       auto new_height = merge_method(old_height, pixel_value + z_bias);
 
       if (old_height == new_height) 
         continue; // no change is necessary
-      
+
       set_height_value(x, y, new_height);
       diff.at<float>(y - top, x - left) = new_height - old_height;
```
AstroStucky, line 271: Related to the above suggestion.
```suggestion
      diff.at<float>(iy, ix) = new_height - old_height;
```

AstroStucky (approved):
I regression tested the branch by commanding a series of perpendicular TaskGrinds and TaskScoopCirculars that extended to the outer range of the workspace. Terrain was modified as expected which resulted in modification differential messages as expected.

There is a necessary change that was missed involving the differential image, see my code suggestions. 

Great find! Coordinate transforms you went with are much simpler, and apparently more correct.

ow_dynamic_terrain/src/TerrainModifier.cpp
mogumbo: Good eye. Fixed.

ow_dynamic_terrain/src/TerrainModifier.cpp
mogumbo: fixed

PR #351, opened 2023-07-26
OCEANWATER-1175_Additional_rostests_2

Hao12450:
## Linked Issues:

## Summary of Changes
Additional testing for `arm_stop`, `arm_move_joints_guarded`, `pan_tilt_move_cartesian` and `pan_tilt_move_joints`.
* `test_pan_tilt_move_cartesian`: Antenna facing the given valid vector in space.
* `test_pan_tilt_move_joints`: Change antenna position by assigning the angle of pan and tilt, The first 9 tests using `PanTiltMoveJointsAction,` in which the pan and tilt move simultaneously, the `test_10` with `PanAction`only move `Pan` and `test_11`with `TiltAction` only move `Tilt`.
* `test_arm_move_joints_guarded`: Scoop tries to dig the ground but with different parameters, `test 01` pre-position the scoop at the specified location and then proceeds with the digging with `test 02`. `Test 02` rotates the hand_yaw and tries to dig, expect to see the action stop since the force_threshold is low. `Test 03` and `Test 04` will rotate the shoulder_yaw, and we expect to see `Test 03` drag the scoop at a small distance and `Test 04` stop the arm very quickly because of the low force and torque threshold.
* `test_arm_stop`: Except for the first test that calls the `arm_stop` directly, the test can split into two parts, at first the function call one of the arm move action, and the second part use `arm_stop` to stop that action. `client.wait_for_result(timeout=rospy.Duration.from_sec(5))` is used between the arms move and `arm_stop` in which allow `arm_stop` call before arm move finish, the number in `Duration.from_sec( )`specify the time in second wait between two actions.
## Test
* `rostest ow_sim_tests test_pan_tilt_move_cartesian.test`
* `rostest ow_sim_tests test_pan_tilt_move_joints.test`
* `rostest ow_sim_tests test_arm_move_joints_guarded.test`
* `rostest ow_sim_tests test_arm_stop.test`

### DIFF
--- a/ow_sim_tests/scripts/test_arm_stop.py
+++ b/ow_sim_tests/scripts/test_arm_stop.py
@@ -8,6 +8,7 @@ import rospy
 import roslib
 import unittest
 import owl_msgs.msg
+import actionlib
 
 from action_testing import *
 
@@ -25,19 +26,16 @@ class TestArmStop(unittest.TestCase):
     # proceed with test only when ros clock has been initialized
     while rospy.get_time() == 0:
       rospy.sleep(0.1)
-    
+
+  # GoalStatus as ABORTED because arm_stop is not stop any action
   def test_01_arm_stop(self):
-
-    client_stop = actionlib.SimpleActionClient('ArmStop', owl_msgs.msg.ArmStopAction)
-
-    connected_stop = client_stop.wait_for_server(timeout = rospy.Duration(50))
-
-    client_stop.send_goal(owl_msgs.msg.ArmStopGoal())
-    client_stop.wait_for_result()
-
-    result = client_stop.get_result()
-
-    self.assertTrue(result, "No arm trajectory to stop")
+    arm_stop_result = test_arm_action(self,
+      'ArmStop', owl_msgs.msg.ArmStopAction,
+      owl_msgs.msg.ArmStopGoal(),
+      TEST_NAME, "test_01_arm_stop",
+      server_timeout = 50.0, # (seconds) first action call needs longer timeout,
+      expected_end_state = actionlib.GoalStatus.ABORTED
+    )
 
   def test_02_arm_unstow_and_stop(self):
 
@@ -55,7 +53,7 @@ class TestArmStop(unittest.TestCase):
 
     result = client_stop.get_result()
 
-    self.assertTrue(result, "Arm trajectory stopped")
+    self.assertEqual(client_stop.get_goal_status_text(), "Arm trajectory stopped")
 
   def test_03_task_deliver_sample_and_stop(self):
 
@@ -73,7 +71,7 @@ class TestArmStop(unittest.TestCase):
 
     result = client_stop.get_result()
 
-    self.assertTrue(result, "Arm trajectory stopped")
+    self.assertEqual(client_stop.get_goal_status_text(), "Arm trajectory stopped")
 
   def test_04_arm_stow_and_stop(self):
 
@@ -91,7 +89,7 @@ class TestArmStop(unittest.TestCase):
 
     result = client_stop.get_result()
 
-    self.assertTrue(result, "Arm trajectory stopped")
+    self.assertEqual(client_stop.get_goal_status_text(), "Arm trajectory stopped")
 
 if __name__ == '__main__':
   import rostest
--- a/ow_sim_tests/scripts/test_arm_stop.py
+++ b/ow_sim_tests/scripts/test_arm_stop.py
@@ -16,6 +16,10 @@ PKG = 'ow_sim_tests'
 TEST_NAME = 'arm_stop'
 roslib.load_manifest(PKG)
 
+# NOTE: Running this test at 3x real-time factor results in a slightly different
+# log outcome, but otherwise the test still succeeds. This should be fixed by
+# OW-1193
+
 class TestArmStop(unittest.TestCase):
 
   @classmethod
--- a/ow_sim_tests/scripts/test_arm_stop.py
+++ b/ow_sim_tests/scripts/test_arm_stop.py
@@ -51,8 +51,6 @@ class TestArmStop(unittest.TestCase):
     client_stop.send_goal(owl_msgs.msg.ArmStopGoal())
     client_stop.wait_for_result()
 
-    result = client_stop.get_result()
-
     self.assertEqual(client_stop.get_goal_status_text(), "Arm trajectory stopped")
 
   def test_03_task_deliver_sample_and_stop(self):
@@ -69,8 +67,6 @@ class TestArmStop(unittest.TestCase):
     client_stop.send_goal(owl_msgs.msg.ArmStopGoal())
     client_stop.wait_for_result()
 
-    result = client_stop.get_result()
-
     self.assertEqual(client_stop.get_goal_status_text(), "Arm trajectory stopped")
 
   def test_04_arm_stow_and_stop(self):
@@ -86,8 +82,6 @@ class TestArmStop(unittest.TestCase):
 
     client_stop.send_goal(owl_msgs.msg.ArmStopGoal())
     client_stop.wait_for_result()
-
-    result = client_stop.get_result()
 
     self.assertEqual(client_stop.get_goal_status_text(), "Arm trajectory stopped")
 
--- a/ow_sim_tests/scripts/test_arm_stop.py
+++ b/ow_sim_tests/scripts/test_arm_stop.py
@@ -1,0 +1,98 @@
+#!/usr/bin/env python
+
+# The Notices and Disclaimers for Ocean Worlds Autonomy Testbed for Exploration
+# Research and Simulation can be found in README.md in the root directory of
+# this repository.
+
+import rospy
+import roslib
+import unittest
+import owl_msgs.msg
+
+from action_testing import *
+
+PKG = 'ow_sim_tests'
+TEST_NAME = 'arm_stop'
+roslib.load_manifest(PKG)
+
+class TestArmStop(unittest.TestCase):
+
+  @classmethod
+  def setUpClass(cls):
+    rospy.init_node("arm_stop_test")
+
+    set_ignore_action_checks(True)
+    # proceed with test only when ros clock has been initialized
+    while rospy.get_time() == 0:
+      rospy.sleep(0.1)
+    
+  def test_01_arm_stop(self):
+
+    client_stop = actionlib.SimpleActionClient('ArmStop', owl_msgs.msg.ArmStopAction)
+
+    connected_stop = client_stop.wait_for_server(timeout = rospy.Duration(50))
+
+    client_stop.send_goal(owl_msgs.msg.ArmStopGoal())
+    client_stop.wait_for_result()
+
+    result = client_stop.get_result()
+
+    self.assertTrue(result, "No arm trajectory to stop")
+
+  def test_02_arm_unstow_and_stop(self):
+
+    client_unstow = actionlib.SimpleActionClient('ArmUnstow', owl_msgs.msg.ArmUnstowAction)
+    client_stop = actionlib.SimpleActionClient('ArmStop', owl_msgs.msg.ArmStopAction)
+
+    connected_unstow = client_unstow.wait_for_server(timeout = rospy.Duration(10))
+    connected_stop = client_stop.wait_for_server(timeout = rospy.Duration(10))
+    
+    client_unstow.send_goal(owl_msgs.msg.ArmUnstowGoal())
+    client_unstow.wait_for_result(timeout=rospy.Duration.from_sec(5))
+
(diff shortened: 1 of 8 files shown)


kmdalal (approved):
All tests passed on first try.  My code review was light and is better handled by @AstroStucky, who can also better assess the choice of tests.  Glad to see more and better tests keep coming!

kmdalal:
I forgot to comment that I consistently saw this error when the simulator started.  I'd never seen it before and it doesn't happen on noetic-devel.  Maybe @Lumsterling has an idea what would cause it.
```
[ERROR]: OW_POWER_SYSTEM ERROR: _Map_base::at while attempting to read system.cfg in the node!
Ensure system.cfg is formatted properly before restarting the simulation.   
```

Lumsterling:
> ```
> [ERROR]: OW_POWER_SYSTEM ERROR: _Map_base::at while attempting to read system.cfg in the node!
> Ensure system.cfg is formatted properly before restarting the simulation.   
> ```

I'll double check this tomorrow, but I'm pretty sure this means that the power system wasn't able to read system.cfg in the power system due to the ``_Map_base::at`` error (it was thrown while trying to read it). I *think* this specific error just means it couldn't find system.cfg at the directory it's supposed to be at but I don't recall for certain.

As a heads up, this error terminates the power system. It can't start properly if it can't read the starting parameters, after all. I might switch to this branch and run the simulation myself to see if I'm getting the same error you are, because it doesn't look like anything here modified the power system.

Lumsterling:
Just switched to this branch and did the tests as well as ran the simulation by itself, and I never observed the power system error you mentioned @kmdalal.

In fact, I should have caught this earlier: the error you saw is only present in the multi-cell model implementation, as part of the QoL changes I'm working on. It's not part of this branch at all (it can occur but it won't have that error message). Were you still on the OW-1187 branch from the previous PR review by chance?

kmdalal:
Thanks for investigating and also running the tests, @Lumsterling .  I was definitely on this branch (OW-1175) when I saw the errors.  But I realize the problem: I had been on the OW-1187 branch previously, and probably didn't rebuild given that this branch only adds Python scripts.

And I forgot to mention that I switched to noetic-devel, rebuilt, and did not see this error.  So it seems like this error only occurs (sometimes?) on OW-1187.  Sorry for confusing this PR!

Lumsterling:
> And I forgot to mention that I switched to noetic-devel, rebuilt, and did not see this error. So it seems like this error only occurs (sometimes?) on OW-1187. Sorry for confusing this PR!

I'll go take a quick look at OW-1187 and see if I can replicate the error. I'm pretty sure it's an issue with ``system.cfg`` not being in the correct spot, which might have happened when switching branches (maybe the directory is different in the ``noetic-devel``-based branches compared to my multi-cell branch), but if I can replicate it and confirm it as an issue I'll let you know.

Lumsterling:
Confirmed @kmdalal that the error occurs when you compile in the OW-1187 branch, then move to the OW-1175 branch and start the simulator without rebuilding. I'm not sure why exactly this happens given the location of ``system.cfg`` doesn't change at all between them, but at the same time, I'm unsure if I should commit time to looking at this given it doesn't happen if you build the branch as intended.

kmdalal:
Thanks for the further investigation, @Lumsterling .  This doesn't seem further effort, since it's not showing up in noetic-devel.  Let's see if it does after this branch is merged.

ow_sim_tests/scripts/test_arm_stop.py
```diff
@@ -0,0 +1,98 @@
...
+
+    connected_stop = client_stop.wait_for_server(timeout = rospy.Duration(50))
+
+    client_stop.send_goal(owl_msgs.msg.ArmStopGoal())
+    client_stop.wait_for_result()
+
+    result = client_stop.get_result()
+
+    self.assertTrue(result, "No arm trajectory to stop")
+
```
AstroStucky, line 41: [optional]
If you'd like, the **action_testing.py** can be used here, but only here since the other units are calling actions asynchronously. But here, you call a single action synchronously. 

```suggestion
    arm_stop_result = test_arm_action(self,
      'ArmStop', owl_msgs.msg.ArmStopAction,
      owl_msgs.msg.ArmStopGoal(),
      TEST_NAME, "test_01_arm_stop",
      server_timeout = 50.0 # (seconds) first action call needs longer timeout,
      expected_end_state = actionlib.GoalStatus.SUCCEEDED
    )

```

There is a drawback, which is `test_arm_action` returns the action result, which I previously thought contained the status and status message of the action, but in fact does not. You'll see in **action_testing.py** on line 133, there is a TODO to instead return the client object. 

So if you would rather test the result message, the change could also just be what I discuss in the next comment down. 

```suggestion
    client_stop = actionlib.SimpleActionClient('ArmStop', owl_msgs.msg.ArmStopAction)

    connected_stop = client_stop.wait_for_server(timeout = rospy.Duration(50))

    client_stop.send_goal(owl_msgs.msg.ArmStopGoal())
    client_stop.wait_for_result()

    result = client_stop.get_result()

    self.assertEquals(client_stop.get_goal_status_text(), "No arm trajectory to stop")

```

ow_sim_tests/scripts/test_arm_stop.py
```diff
@@ -0,0 +1,98 @@
...
+    
+    client_unstow.send_goal(owl_msgs.msg.ArmUnstowGoal())
+    client_unstow.wait_for_result(timeout=rospy.Duration.from_sec(5))
+
+    client_stop.send_goal(owl_msgs.msg.ArmStopGoal())
+    client_stop.wait_for_result()
+
+    result = client_stop.get_result()
+
+    self.assertTrue(result, "Arm trajectory stopped")
```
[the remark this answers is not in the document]
AstroStucky: Please apply to the remaining action units as needed.

AstroStucky (changes requested):
I think the approach to testing ArmStop is solid, and I am glad to see it gave you the chance to work directly with the `actionlib` python library. This was the right choice since `action_testing.py` does not support calling of actions asynchronously. 

All the tests passed fine for me. I just have a handful of code comments.

ow_sim_tests/scripts/test_arm_stop.py
Hao12450: Thank you so much for the correction. I also make changes on ` expected_end_state = actionlib.GoalStatus.SUCCEEDED`, in my understanding should be `expected_end_state = actionlib.GoalStatus.ABORTED` since `test_01` does not stop any action, therefore we expect to see the end state as `ABORTED`.

AstroStucky: Yes! Good catch, I mean for that to be ABORTED. Fixed in the original message.

ow_sim_tests/scripts/test_arm_stop.py
```diff
@@ -55,7 +53,7 @@ def test_02_arm_unstow_and_stop(self):
 
     result = client_stop.get_result()
```
AstroStucky, line 54: `result` is no longer used, please remove it. Same for the other test units.

ow_sim_tests/scripts/test_pan_tilt_move_joints.py
```diff
@@ -0,0 +1,141 @@
...
+class TestPanTiltMoveJoints(unittest.TestCase):
+
+  @classmethod
+  def setUpClass(cls):
+    rospy.init_node("pan_tilt_move_joints_test")
+
+    set_ignore_action_checks(True)
+    # proceed with test only when ros clock has been initialized
+    while rospy.get_time() == 0:
+      rospy.sleep(30.0)
```
AstroStucky, line 27: Same change here please
```suggestion
    while rospy.get_time() == 0:
      rospy.sleep(0.1)
    # NOTE: work around for pan/tilt actions failing when called too early (see OW-1181)
    rospy.sleep(30.0)
```

ow_sim_tests/scripts/test_pan_tilt_move_joints.py
```diff
@@ -0,0 +1,141 @@
...
+
+  def test_01_pan_tilt_move_joints(self):
+    pan_tilt_move_joints_result = test_action(self,
+      'PanTiltMoveJoints', owl_msgs.msg.PanTiltMoveJointsAction,
+      owl_msgs.msg.PanTiltMoveJointsGoal(
+        pan = 0.785,
+        tilt = 1
+      ),
+      TEST_NAME, "test_01_pan_tilt_move_joints",
+      server_timeout = 400.0 # (seconds) first action call needs longer timeout
```
AstroStucky, line 37:
```suggestion
      server_timeout = 50.0 # (seconds) first action call needs longer timeout
```

AstroStucky (commented):
Some small changes requested

ow_sim_tests/scripts/test_pan_tilt_move_joints.py
Hao12450: Thank you for the review!

AstroStucky:
I noticed `test_arm_stop_test.test` results in slightly different log outcomes when running at real-time compared to running at 3x real-time.

Output at 1x time
```
[INFO:/lander_action_servers] ArmStop action started
[ERROR:/lander_action_servers] ArmStop: Aborted - No arm trajectory to stop
[INFO:/lander_action_servers] ArmStop action complete
[INFO:/lander_action_servers] ArmUnstow action started
[INFO:/lander_action_servers] ArmStop action started
[INFO:/lander_action_servers] ArmStop: Succeeded - Arm trajectory stopped
[INFO:/lander_action_servers] ArmStop action complete
[ERROR:/lander_action_servers] ArmUnstow: Aborted - Stop was called; trajectory execution ceased
[INFO:/lander_action_servers] ArmUnstow action complete
[INFO:/lander_action_servers] TaskDeliverSample action started
[INFO:/lander_action_servers] ArmStop action started
[INFO:/lander_action_servers] ArmStop: Succeeded - Arm trajectory stopped
[INFO:/lander_action_servers] ArmStop action complete
[ERROR:/lander_action_servers] TaskDeliverSample: Aborted - Stop was called; trajectory execution ceased
[INFO:/lander_action_servers] TaskDeliverSample action complete
[INFO:/lander_action_servers] ArmStow action started
[INFO:/lander_action_servers] ArmStop action started
[INFO:/lander_action_servers] ArmStop: Succeeded - Arm trajectory stopped
[INFO:/lander_action_servers] ArmStop action complete
[ERROR:/lander_action_servers] ArmStow: Aborted - Stop was called; trajectory execution ceased
[INFO:/lander_action_servers] ArmStow action complete
```

Output running at 3x time
```
[INFO:/lander_action_servers] ArmStop action started
[ERROR:/lander_action_servers] ArmStop: Aborted - No arm trajectory to stop
[INFO:/lander_action_servers] ArmStop action complete
[INFO:/lander_action_servers] ArmUnstow action started
[INFO:/lander_action_servers] ArmStop action started
[INFO:/lander_action_servers] ArmStop: Succeeded - Arm trajectory stopped
[ERROR:/lander_action_servers] ArmUnstow: Aborted - Stop was called; trajectory execution ceased
[INFO:/lander_action_servers] ArmStop action complete
[INFO:/lander_action_servers] ArmUnstow action complete
[INFO:/lander_action_servers] TaskDeliverSample action started
[INFO:/lander_action_servers] ArmStop action started
[INFO:/lander_action_servers] ArmStop: Succeeded - Arm trajectory stopped
[INFO:/lander_action_servers] ArmStop action complete
[INFO:/lander_action_servers] ArmStow action started
[ERROR:/lander_action_servers] ArmStow: Aborted - Arm is already checked out by the TaskDeliverSample action server
[INFO:/lander_action_servers] ArmStow action complete
[INFO:/lander_action_servers] ArmStop action started
[INFO:/lander_action_servers] ArmStop: Succeeded - Arm trajectory stopped
[INFO:/lander_action_servers] ArmStop action complete
```

This is a problem with lander arm actions in general that is now captured by [OW-1193](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1193). To summarize it, ArmStop cannot interrupt an action until it has finished planning, and the effect can only be seen at 3x speed because TaskSampleDeliver takes a long time to plan (in real-time), and the wait-time between calling an action and calling ArmStop is using simulation time.

ow_sim_tests/scripts/test_arm_stop.py
```diff
@@ -0,0 +1,90 @@
...
+import unittest
+import owl_msgs.msg
+import actionlib
+
+from action_testing import *
+
+PKG = 'ow_sim_tests'
+TEST_NAME = 'arm_stop'
+roslib.load_manifest(PKG)
+
```
AstroStucky, line 18:
```suggestion

# NOTE: Running this test at 3x real-time factor results in a slightly different
# log outcome, but otherwise the test still succeeds. This should be fixed by
# OW-1193

```

PR #353, opened 2023-08-08
Made terrain texture mask

mogumbo:
| Jira Ticket 🎟️ | [OCEANWATER-1142](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1142) |

## Summary of Changes
* Made terrain texture mask, which is updated when heightmap pixels are modified.

## Test
* This can't be tested without modifying a material file and shader. Instead, I offer this image as proof that it works. Terrain is drawn in yellow where the mask has been set to 0. 
Screenshot_20230808_141448

### DIFF
--- a/ow_dynamic_terrain/src/DynamicTerrainVisual.cpp
+++ b/ow_dynamic_terrain/src/DynamicTerrainVisual.cpp
@@ -2,7 +2,7 @@
 // Research and Simulation can be found in README.md in the root directory of
 // this repository.
 
-#include "DynamicTerrainBase.h"
+#include "DynamicTerrainVisual.h"
 #include "TerrainModifier.h"
 #include <gazebo/rendering/RenderingIface.hh>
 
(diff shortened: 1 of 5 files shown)


AstroStucky:
You did it! 😀 You had me thinking this would be abandoned, so this PR is a welcomed sight.

I take it having this happen when the grinder carves terrain is still under development?

mogumbo:
> You did it! grinning You had me thinking this would be abandoned, so this PR is a welcomed sight.
> 
> I take it having this happen when the grinder carves terrain is still under development?

Yes, I'm afraid this PR isn't any use on its own. We still need to a) always modify the visual terrain in a plausible way when grinding happens, and b) update the terrain shader to draw the grinder tailings.

ow_dynamic_terrain/src/DynamicTerrainVisual.h
mogumbo, line 1: I was adding enough material to DynamicTerrainVisual that I thought it was cleaner to give it its own header file.

ow_dynamic_terrain/src/DynamicTerrainVisual.cpp
```diff
@@ -13,129 +13,133 @@ using namespace Ogre;
...
+}
 
-  static inline void setHeightFromWorldCoords(Ogre::Terrain* terrain, int x, int y, float value)
-  {
-    value -= terrain->getPosition().z;
-    terrain->setHeightAtPoint(x, y, value);
-  }
+float DynamicTerrainVisual::getHeightInWorldCoords(const Ogre::Terrain* terrain, int x, int y)
+{
+  auto value = terrain->getHeightAtPoint(x, y);
```
kmdalal, line 35: Is it not valid or safe to declare `value` a float?

ow_dynamic_terrain/src/DynamicTerrainVisual.h
```diff
@@ -0,0 +1,66 @@
...
+#include "TerrainModifier.h"
+#include <gazebo/rendering/RenderingIface.hh>
+
+namespace ow_dynamic_terrain
+{
+class DynamicTerrainVisual : public gazebo::VisualPlugin, public DynamicTerrainBase
+{
+public:
+  DynamicTerrainVisual();
+
```
kmdalal, line 15: Though I don't think it's in our coding guidelines, I like to see all classes either define, `delete`, or `default` the copy constructor, destructor, and assignment operator.

ow_dynamic_terrain/src/DynamicTerrainVisual.h
```diff
@@ -0,0 +1,66 @@
...
+  void Load(gazebo::rendering::VisualPtr /*visual*/, sdf::ElementPtr /*sdf*/) override;
```
kmdalal, line 16: Minor: why not just omit the parameter names instead of commenting them out?

ow_dynamic_terrain/src/DynamicTerrainVisual.cpp
mogumbo: This was existing code, but it got split into .h and .cpp files. I agree this should be float, but I would prefer to do this sort of cleanup in a separate ticket.

ow_dynamic_terrain/src/DynamicTerrainVisual.h
mogumbo: This was existing code, but it got split into .h and .cpp files. I would prefer to do this sort of cleanup in a separate ticket.

ow_dynamic_terrain/src/DynamicTerrainVisual.h
kmdalal: I'll propose it in a future meeting.  It simply makes explicit the intention.  Without anything you get `default` but it's good to know that was intended.

kmdalal:
During ReferenceMission1, I get this for the first time:
```
[ERROR]: TaskScoopCircular: Aborted - The follow_joint_trajectory action server for arm_controller resulted in error code -5 with the message: j_dist_pitch goal error 0.704364
```
I'll retest on noetic-devel as it may not be related to this branch.

AstroStucky:
> During ReferenceMission1, I get this for the first time:
> 
> ```
> [ERROR]: TaskScoopCircular: Aborted - The follow_joint_trajectory action server for arm_controller resulted in error code -5 with the message: j_dist_pitch goal error 0.704364
> ```
> 
> I'll retest on noetic-devel as it may not be related to this branch.

This indicates the arm is not completing its planned trajectory. It may be getting jammed on collision terrain.

kmdalal:
Please see my revised comment above about the goal errors I'm seeing on this branch.

ow_dynamic_terrain/src/DynamicTerrainVisual.cpp
AstroStucky: I have a mixture of opinions about `auto`, but it does seem unnecessary here unless an `auto` return type is used, but then that would make the function's prototype far less descriptive... 

My feelings are `auto` usage should be very case-by-case. Sometimes I don't care about the type, other times I want to be very explicit about it if for no other reason than the code self-documents better.

This is just discussion by the way, not a change request. I'm fine with keeping the definition as it was.

ow_dynamic_terrain/src/DynamicTerrainVisual.h
AstroStucky: I think it's a good practice when you have vital code in a class's custom constructor, but not in a default constructor, such that if the default constructor were to be used you would more then certainly wind up with an error.

That said, we are talking about a Gazebo plugin here, so the only thing that will ever construct it is Gazebo code that neither us nor our users should ever touch.

AstroStucky: Actually I just talked myself into feeling that deleting unused constructors is good even for a Gazebo plugin like this. As unlikely as it may be, if Gazebo had an update where it now wants to pass some argument to its plugin constructors, or it for some reason has to copy them now, you would know what the problem is immediately from a compile error, but if you had not deleted unwanted constructors you may have to painstakingly find out by observing that your class is now erring and working backwards to find out why.

ow_dynamic_terrain/src/TerrainModifier.cpp
```diff
@@ -250,9 +252,25 @@ bool TerrainModifier::applyImageToHeightmap(Heightmap* heightmap, const cv::Poin
 
   auto diff = OpenCV_Util::createZerosMatLike(image);
 
+  uint32 mask_size = 0;
+  HardwarePixelBufferSharedPtr pixel_buffer;
+  uint8* mask_pixels = nullptr;
```
AstroStucky, line 257: Apparently `uint*` is provided by some compilers, but `uint*_t` is in the standard library. I think we should use the standard library versions. This comment applies to all instances of these types that in other parts of the code.
```suggestion
  uint32_t mask_size = 0;
  HardwarePixelBufferSharedPtr pixel_buffer;
  uint8_t* mask_pixels = nullptr;
```

We could also leave this for a refactor.

ow_dynamic_terrain/src/TerrainModifier.cpp
```diff
@@ -250,9 +252,25 @@ bool TerrainModifier::applyImageToHeightmap(Heightmap* heightmap, const cv::Poin
...
+  {
+    mask_size = maskmap->getWidth();
+    // Lock the mask map pixel buffer
+    pixel_buffer = maskmap->getBuffer();
+    pixel_buffer->lock(HardwareBuffer::HBL_DISCARD);
+    // Get pixel pointer
+    const PixelBox& pixel_box = pixel_buffer->getCurrentLock();
+    mask_pixels = static_cast<uint8*>(pixel_box.data);
+  }
+
```
AstroStucky, line 268: I am not sure what image library you're using here, but I am wondering why it's not OpenCV, like the rest of the code uses? Is it a special case for the mask? Should the other code use it instead of OpenCV?

AstroStucky (commented):
I was not familiar with a lot of the image operations that were going on, but I reviewed what I could understand.

AstroStucky:
Seeing as this is an incomplete part of a whole, I think it needs to be merged into an epic branch instead of noetic-devel.

AstroStucky:
`rostest ow_sim_tests sample_collection.test ignore_action_checks:=true` succeeded (the flag is necessary due to an incompatibility of this test with Release 11+ actions, but I assure you that action result is still checked for success).

I reattempted Mike's ReferenceMission1.plx test on atacama_y1a and got a different error.

```
[INFO:/lander_action_servers] TaskDiscardSample action started
[ WARN:/move_group] Solution is approximate. This usually indicates a failure.
[ WARN:/move_group_commander_wrappers_1692214908639257159] Fail: ABORTED: No motion plan found. No execution attempted.                                                                                                     
[ERROR:/lander_action_servers] TaskDiscardSample: Aborted - MoveIt planning failed with error code: val: -1
[INFO:/lander_action_servers] TaskDiscardSample action complete
```

This comes as a surprise to me. I am wondering if anyone else sees TaskDiscardSample randomly failing to plan a trajectory like this? 

At this point I am not entirely convinced the ReferenceMission1 errors are due to this PR. I would like to run more tests in both this branch and noetic-devel.

ow_dynamic_terrain/src/TerrainModifier.cpp
mogumbo: I'm using Ogre API to modify an Ogre texture, which is reasonable. I don't know the specifics of why OpenCV was used to modify an Ogre Heightmap. There might be a good reason for it.

mogumbo:
I have been able to reproduce all the errors described by @mikedalal and @AstroStucky using the noetic-devel branch. They only happen about 25% of the time for me, so it takes a while to get an error. I don't see any reason to believe these errors are related to this PR.

ow_dynamic_terrain/src/TerrainModifier.cpp
mogumbo: Fixed

mogumbo:
> Seeing as this is an incomplete part of a whole, I think it needs to be merged into an epic branch instead of noetic-devel.

I think it's a good practice to use an epic branch if parts of that epic will temporarily break the code for other developers or cause some other exceptional situation. Nothing like that is happening here, so I think it would just be unnecessary overhead.

kmdalal (approved):
I'm opening a new Jira for the ReferenceMission1 errors -- thanks @mogumbo and @AstroStucky for the testing.

I'm satisfied with this and agree it's fine to merge by itself.

AstroStucky (approved):

PR #354, opened 2023-08-15
OW-1196: battery pack partial disabling

Lumgineer:
## Linked Issues:
| EPIC ⚡ | n/a |
| Jira Ticket 🎟️ | [OCEANWATER-1196](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1196) |
| Github :octocat:  | # |

**Note**: If you are a new reviewer (i.e. anyone except Mike & Thomas), the only test you need to complete is the performance test at the bottom of this PR description.

This PR will be merged into [OW-1163](https://github.com/nasa/ow_simulator/pull/352), ideally before its merge into noetic-devel. It should provide an option for users to reduce the CPU usage of the new multi-cell battery system without fully sacrificing the new changes, while also adding a new fault type only possible with the new implementation.

## Summary of Changes
* A new variable has been added to ``system.cfg``: ``active_models``. This can be set from 1-4, causing the power system to instantiate that number of prognosers. 4 are needed to simulate a full battery, but the CPU load is quite high as a result (excessively so if the user is not planning on focusing on the power model), so the default here is 1.
  * Power distribution will remain as-is, meaning that even in the case of 1 prognoser, it will still only be fed 1/4th of the power draw (as if the other models were still running and being fed the other 3/4ths). In this sense, reducing the number of active models will let the user get predictions as if they were sampling one part of the battery instead.
* A new power fault has been added: battery disconnection. The user can specify in the RQT window a number of battery nodes to disconnect, between 0-3 (since there are 4 nodes total). This will have the effect of halting input data being sent into their respective GSAP prognosers, effectively removing them from the system (this will tangentially increase performance since the prognosers will not start new predictions, but it is not the purpose of the fault). The remaining node(s) will receive a larger fraction of any applied power draw from external sources. This simulates losing one or more parallel nodes in a battery pack, which would force any additional load onto the remaining cells.
  * For example, at the start of the simulation, any running prognoser receives 1/4th of any applied power draw. If 1 model is disconnected, the remaining models will now receive 1/3rd the power each. If 3 models disconnect, the remaining 1 model will receive 100% of the power draw. This will not have an instantaneous effect on predictions unless there is already a significant amount of power draw being applied (i.e. if the battery loses 3 nodes at 95% SoC, it will still read 95% SoC), but the battery will deteriorate faster as a result of higher power draw. Discussion with Chetan may be necessary to confirm this is the expected behavior with this kind of fault.
    * This behavior does not change if less than 4 prognosers are instantiated at start to save on performance. If only 1 is running & 3 prognosers are designed to be disconnected, the simulation will run as if power was being distributed across 4 prognosers anyways.

## Tests Overview
* For both tests, check ``ow_power_system/config/system.cfg`` and ensure ``print_debug`` is ``true`` to allow important battery stats to be printed to the terminal during runtime. For the purposes of these tests, all debug printout flags in ``system.cfg`` should be set to true for simplicity.

## Test 1 (Battery Disconnection Fault)
* In ``system.cfg``, set ``active_models`` to 4 for this test.
* Open up a terminal and build & run the simulator in the OceanWATERS directory. The world shouldn't matter, though I used the ``atacama`` world (``roslaunch ow atacama_y1a.launch``).
* Monitor the terminal and confirm the 4 separate models running predictions.
* Once the simulation has started, navigate to the ``Dynamic Reconfigure`` tab on the bottom of the RQT window, and from there, to the ``faults`` section. Scroll down until you see the section governing the new fault, ``disconnect_battery_nodes``.
  * Adjust the value of ``battery_nodes_to_disconnect`` to some nonzero value (note it only allows whole numbers) and activate it by checking ``disconnect_battery_nodes``.
  * **Monitor the debug statement outputs and confirm that the correct amount of models were deactivated** (e.g. if 1 node was disconnected, 3 should remain). **Confirm that adjusting the value of ``battery_nodes_to_disconnect`` after the fault has already been activated updates the number of disconnected nodes, only if the number was increased. Decreasing the number should not result in any changes to the system**.
  * Using the ``high_power_draw`` fault, apply an additional amount of power draw on the system.
  * **Confirm that the additional power draw is correctly distributed amongst the remaining models** (e.g. with 120W of additional power draw amongst 3 remaining models, each should see a 40W increase).
  * Deactivate ``disconnect_battery_nodes`` and observe the warning statement (though it may be easier to miss with the debug printouts). **Confirm that the disconnection remains**.
  * Adjust the value of ``battery_nodes_to_disconnect`` and reapply the fault. **Confirm, if the number of disconnected nodes was higher than before, that the new number is reflected in the debug printouts. If the number of disconnected nodes is lower than before, confirm that no change in behavior occurs**.

## Test 2 (Modifying Active Models)
* Ensure ``active_models`` is set to 1 in ``system.cfg``.
* Open up a terminal and build & run the simulator in the OceanWATERS directory.
* Monitor the terminal and confirm only 1 model is running predictions. **Monitor the CPU usage of the simulation. ``ow_power_system`` should be using much less of the CPU**.
* Like the previous test, go to the ``Dynamic Reconfigure`` tab in the RQT window and navigate to the ``faults`` section. Use ``high_power_draw`` to apply an additional power load on the system.
  * **Confirm the single model only receives 1/4th the additional power draw from the fault such that its individual behavior is identical to if performance mode was off**.
  * Using the new battery disconnection fault, remove one or more models from the simulation and check the power draw. **Confirm the power draw is still affected by the disconnection fault with identical distribution behavior to that seen in Test 1**. The only difference should be that only 1 model is actually running at any given time, regardless of the number of disconnected nodes.

## Performance Test (for new reviewers)
This test is just to informally measure the CPU usage of these changes to the power system, and should require little input other than idling your computer:
* Navigate to ``ow_power_system/config/system.cfg`` and set ``active_models`` to 4.
* Build & run the simulator in the OceanWATERS directory under this OW-1196 branch. The world might affect performance one way or another, though I used the ``atacama`` world across all my tests (``roslaunch ow atacama_y1a.launch``).
* Let the simulation run idly for at least 15 minutes. Afterwards, record the average CPU load, the % of the CPU the ``ow_power_system`` process used, the number of cores in your setup, and your RAM size.
  * The average CPU load can be determined by opening up a 2nd terminal and running the ``uptime`` command, which will display the CPU load average for the past 1, 5, & 15 minutes. We're looking for the 15 minute average.
  * ``ow_power_system``'s CPU usage can be seen by running the ``top`` command in a 2nd terminal while the simulation is running. It can also be checked via System Monitor. Note that ``top`` will show CPU usage relative to a single core while System Monitor will show CPU usage relative to all cores (to my understanding), so the percentages will be very different.
  * The amount of RAM your setup has can be determined by going to Settings -> About.
  * The number of CPU cores you have can be seen by calling `nproc` in the terminal.
* Feel free to record any qualitative observations you might notice while the simulation is idling (e.g. computer heating up, fans getting faster, etc.).
* Repeat this one time with ``active_models`` set to ``2`` in ``system.cfg`` instead, letting it idle for 15 minutes again to get the new load average from ``uptime``.
* Report your observations as a comment to this PR.

### DIFF
--- a/ow_power_system/src/PowerSystemNode.cpp
+++ b/ow_power_system/src/PowerSystemNode.cpp
@@ -589,39 +589,11 @@ void PowerSystemNode::injectFault (const std::string& fault_name)
   // Disconnect battery nodes case
   if (fault_name == FAULT_NAME_DBN_ACTIVATE)
   {
-    // Check if the fault has switched, and disconnect nodes if it was turned
-    // on. There is no continuous behavior to monitor.
+    // Check if the fault has switched.
     if (!m_disconnect_battery_nodes_fault_activated && fault_enabled)
     {
-      ros::param::getCached("/faults/" + FAULT_NAME_DBN, dbn_nodes);
-
-
-
+      ROS_INFO_STREAM(fault_name << " activated!");
       m_disconnect_battery_nodes_fault_activated = true;
-
-      // Warn the user if they tried to reduce the number of deactivated nodes
-      // from a previous fault activation.
-      if (m_deactivated_models >= dbn_nodes)
-      {
-        ROS_WARN_STREAM("Attempted to disconnect " << std::to_string(dbn_nodes)
-                        << " out of " << std::to_string(NUM_MODELS)
-                        << " nodes when there are already "
-                        << std::to_string(m_deactivated_models)
-                        << " out of " << std::to_string(NUM_MODELS)
-                        << " disconnected. Behavior has not changed.");
-      }
-      else
-      {
-        ROS_INFO_STREAM(fault_name << " activated! "
-                      << std::to_string(dbn_nodes)
-                      << " out of " << std::to_string(NUM_MODELS) << " models "
-                      << "are now disconnected.");
-      }
-
-      // Update the number of active and deactivated nodes.
-      m_deactivated_models = std::max(m_deactivated_models, dbn_nodes);
-      m_active_models = std::min(m_active_models,
-                                (NUM_MODELS - m_deactivated_models));
     }
     else if (m_disconnect_battery_nodes_fault_activated && !fault_enabled)
     {
@@ -630,9 +602,43 @@ void PowerSystemNode::injectFault (const std::string& fault_name)
                     << "is deactivated.");
       m_disconnect_battery_nodes_fault_activated = false;
     }
+
+    // Continual beahvior with DBN fault.
+    if (m_disconnect_battery_nodes_fault_activated && fault_enabled)
+    {
+      // Get the number of nodes set by the user.
+      ros::param::getCached("/faults/" + FAULT_NAME_DBN, dbn_nodes);
+
+      // Warn the user periodically if they try to reduce the number of
+      // deactivated nodes from a previous fault activation.
+      if (m_deactivated_models > dbn_nodes)
+      {
+        ROS_WARN_STREAM_THROTTLE(30, fault_name << " currently set to "
+                        << "disconnect " << std::to_string(dbn_nodes)
+                        << " out of " << std::to_string(NUM_MODELS)
+                        << " nodes when there are already "
+                        << std::to_string(m_deactivated_models)
+                        << " out of " << std::to_string(NUM_MODELS)
+                        << " disconnected. Behavior has not changed.");
+        ROS_WARN_STREAM_THROTTLE(30, "This warning will repeat periodically "
+                        << "until the fault value is reset to "
+                        << std::to_string(m_deactivated_models)
+                        << ", updated to disconnect more nodes, "
+                        << "or the fault is deactivated.");
+      }
+      else if (m_deactivated_models < dbn_nodes)
+      {
+        ROS_INFO_STREAM(fault_name << " updated! "
+                      << std::to_string(dbn_nodes)
+                      << " out of " << std::to_string(NUM_MODELS) << " models "
+                      << "are now disconnected.");
+        // Update the number of active and deactivated nodes.
+        m_deactivated_models = std::max(m_deactivated_models, dbn_nodes);
+        m_active_models = std::min(m_active_models,
+                                  (NUM_MODELS - m_deactivated_models));
+      }
+    }
   }
-
-
 }
 
 /*
--- a/ow_power_system/src/PowerSystemNode.cpp
+++ b/ow_power_system/src/PowerSystemNode.cpp
@@ -22,6 +22,8 @@ using namespace PCOE;
 
 const std::string FAULT_NAME_HPD           = "high_power_draw";
 const std::string FAULT_NAME_HPD_ACTIVATE  = "activate_high_power_draw";
+const std::string FAULT_NAME_DBN           = "battery_nodes_to_disconnect";
+const std::string FAULT_NAME_DBN_ACTIVATE  = "disconnect_battery_nodes";
 const int CUSTOM_FILE_EXPECTED_COLS           = 2;
 
 // Error flags.
@@ -216,7 +218,12 @@ void PowerSystemNode::initAndRun()
   while(ros::ok())
   {
     // Set up value modifiers in each model handler for injection later.
-    injectFaults();
+    // Skip the first loop to prevent activation/deactivation messages from
+    // printing during startup.
+    if (m_power_models[0].input_info.timestamp > 0)
+    {
+      injectFaults();
+    }
 
     // DEBUG PRINT
     if (m_timestamp_print_debug)
@@ -533,27 +540,30 @@ void PowerSystemNode::injectFault (const std::string& fault_name)
 {
   bool fault_enabled;
   double hpd_wattage = 0.0;
+  int dbn_nodes = 0;
 
   // Get the value of fault_enabled.
   ros::param::getCached("/faults/" + fault_name, fault_enabled);
 
-  if (!m_high_power_draw_activated && fault_enabled)
+  // High power draw case
+  if (fault_name == FAULT_NAME_HPD_ACTIVATE)
   {
-    ROS_INFO_STREAM(fault_name << " activated!");
-    m_high_power_draw_activated = true;
-  }
-  else if (m_high_power_draw_activated && !fault_enabled)
-  {
-    ROS_INFO_STREAM(fault_name << " deactivated!");
-    m_high_power_draw_activated = false;
-  }
-
-  if (m_high_power_draw_activated && fault_enabled)
-  {
-    // If the current fault being utilized is high_power_draw,
-    // simply update wattage based on the current value of the HPD slider.
-    if (fault_name == FAULT_NAME_HPD_ACTIVATE)
+    // Check if the fault has switched.
+    if (!m_high_power_draw_activated && fault_enabled)
     {
+      ROS_INFO_STREAM(fault_name << " activated!");
+      m_high_power_draw_activated = true;
+    }
+    else if (m_high_power_draw_activated && !fault_enabled)
+    {
+      ROS_INFO_STREAM(fault_name << " deactivated!");
+      m_high_power_draw_activated = false;
+    }
+
+    // Continual behavior with HPD fault.
+    if (m_high_power_draw_activated && fault_enabled)
+    {
+      // Update wattage based on the current value of the HPD slider.
       ros::param::getCached("/faults/" + FAULT_NAME_HPD, hpd_wattage);
       double split_wattage = hpd_wattage / (NUM_MODELS - m_deactivated_models);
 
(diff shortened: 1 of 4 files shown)


kmdalal:
Just starting this review, and my first reaction is that it seems like two distinct features, with only the performance mode applicable for merge into 1163.   But we can keep them together if it was much more convenient to do so.

Lumgineer:
Typically I imagine this kind of ticket would be split into 2 separate PRs, but it worked out such that the changes ended up being intertwined (both features work around modifying the number of running GSAP instances).

kmdalal:
Running an idle Atacama world for about 15 minutes with performance_mode yielded a 15-minute load average of about 4, which is consistent with the 16 I experienced running all 4 prognosers.  The fan was running at medium and the keyboard warm but not not hot.  This is what I would consider acceptable (but not great) performance for an application.   

I'll do more testing with the lander performing operations.

kmdalal:
Instead of a Boolean performance_mode, would it make sense to have a setting for number of prognosers (1-4)?  I'm going to suggest a value of true (or 1) for the default, so that we don't give new users a bad first experience!

Lumgineer:
> Running an idle Atacama world for about 15 minutes with performance_mode yielded a 15-minute load average of about 4, which is consistent with the 16 I experienced running all 4 prognosers. The fan was running at medium and the keyboard warm but not hot. This is what I would consider acceptable (but not great) performance for an application.

Do you experience better performance with the single-cell model in ``noetic-devel``? I observed essentially identical ``ow_power_system`` CPU load for both at 7%. I will say that running the simulation on my computer has always led to the fans turning on and the computer heating up somewhat, but I assumed this was expected behavior for the OW simulator overall.

> Instead of a Boolean performance_mode, would it make sense to have a setting for number of prognosers (1-4)? I'm going to suggest a value of true (or 1) for the default, so that we don't give new users a bad first experience!

My understanding is that we were hoping this project would result in a full battery simulation, which consists of four 6S1P models. Anything under that isn't the full simulation and just a sample of the full expected battery. I imagine implementing the ability to choose the exact number of prognosers wouldn't be hard, but I'm not sure I see much of a reason to simulate more than 1 if we're not simulating all 4 of them. Maybe you'd be able to simulate an intra-battery fault that affects only 1 node if you have at least 2 prognosers simulating.

Still, since I don't think it would be too difficult to implement, I can work on this tomorrow. It might also be pertinent to change the flag name off of ``performance_mode`` since limiting to 1 prognoser is probably the best option right now anyways.

kmdalal:
> Still, since I don't think it would be too difficult to implement, I can work on this tomorrow. It might also be pertinent to change the flag name off of `performance_mode` since limiting to 1 prognoser is probably the best option right now anyways.

It was just an idea, not a request.  I think @chetankul and @AstroStucky should weigh in.

kmdalal:
> Do you experience better performance with the single-cell model in `noetic-devel`?

I'll run some tests and compare.

AstroStucky:
> > Still, since I don't think it would be too difficult to implement, I can work on this tomorrow. It might also be pertinent to change the flag name off of `performance_mode` since limiting to 1 prognoser is probably the best option right now anyways.
> 
> It was just an idea, not a request. I think @chetankul and @AstroStucky should weigh in.


It seems a little bit complicated. I think we need to be sure the use-case is there, which I imagine @chetankul could answer better.

Lumgineer:
Since we're conflicted on if the option to specify exactly how many prognosers to run at once is a good addition or not, I'll hold off on implementing it for now. Maybe we can discuss it in the tag-up tomorrow.

kmdalal:
I've added @chetankul as a reviewer of the description in this PR and the discussion.

kmdalal:
The following warning seems to appear only once, but I think it would better appear every time the disconnection fault is unchecked.
```
[ WARN]: Note that disconnected battery nodes cannot be reconnected, even if the fault is deactivated.
```
And the accompanying warning doesn't seem applicable:
```disconnect_battery_nodes deactivated!
```
Finally, the message on reactivation of the fault (which I assume is meaningful only if one increases the number of disconnected nodes) could be improved:
```
disconnect_battery_nodes activated! Disconnecting 3 nodes...
```

kmdalal:
The `battery_nodes_to_disconnect` setting does not behave analogously to `high_power_draw` in that if it is changed while the fault is activated, the system does not respond -- instead the fault must be un-activated and re-activated.   Would be good if the behavior between these faults were consistent.

Lumgineer:
> The `battery_nodes_to_disconnect` setting does not behave analogously to `high_power_draw` in that if it is changed while the fault is activated, the system does not respond -- instead the fault must be un-activated and re-activated. Would be good if the behavior between these faults were consistent.

I'm trying to reproduce this on my end but it seems fully responsive to me. How exactly did you observe it?
I'm activating ``high_power_draw``, then disconnecting at least 1 node, then I'm adjusting the slider value while ``high_power_draw`` is still activated, and I'm seeing the applied power update to reflect the reduced number of nodes.

kmdalal:
> EDIT: I understand now. You're talking about how the slider doesn't update the number of disconnected nodes without re-activating the fault. I can change the behavior.

Yes, this is exactly what I meant.

Lumgineer:
As per Mike's request, I've added the rest of the OceanWATERS team as reviewers to this PR. However, we're only looking for additional performance data on different machines rather than more full reviews of all the changes here. I've attached a performance test at the bottom of the PR description for the new reviewers to complete with instructions on what to do.

kmdalal:
@Keeganfn @Hao12450 @mogumbo Note that you are only asked to do the performance test.  Getting more data for this PR is important before releasing it, because it introduced substantially more CPU usage from the simulator.

AstroStucky:
I misread the test instructions and had only `print_debug` set to true for a run. It turns out when you do that, the simulation prints white space the entire time. I consider this a minor bug, but would like to see it resolved if its not too much work.

AstroStucky:
Since `active_models` is 1 by default, please update test 1 to mention you need to first set the value to 4

Lumgineer:
@Hao12450 I don't see the ``ow_power_system`` process in your screenshots. Did it crash during startup?

Hao12450:
> @Hao12450 I don't see the `ow_power_system` process in your screenshots. Did it crash during startup? If so, it would have left a 'process has died' error message while the simulation is starting with the relevant reason for terminating.

Here is the log from ros: `ServiceCaller.on_refresh_services_button_clicked(): could not get class of service /power_system_node/get_loggers:
Unable to communicate with service [/power_system_node/get_loggers], address [rosrpc://192.168.1.160:49595]
ServiceCaller.on_refresh_services_button_clicked(): could not get class of service /power_system_node/set_logger_level:
Unable to communicate with service [/power_system_node/set_logger_level], address [rosrpc://192.168.1.160:49595]
[INFO] [1692725735.976554, 1916524808.285000]: Successfully connected to the arm_controller/follow_joint_trajectory action server.
[INFO] [1692725736.199580, 1916524808.507500]: Successfully connected to the grinder_controller/follow_joint_trajectory action server.
[INFO] [1692725736.238690, 1916524808.545000]: Successfully connected to the dynamic reconfigure server.
[ INFO] [1692725736.244993091, 1916524808.552500010]: Loading robot model 'lander'...
[ INFO] [1692725737.482909616, 1916524809.789999962]: Ready to take commands for planning group arm.
[ INFO] [1692725737.886983651, 1916524810.192500114]: Ready to take commands for planning group grinder.

`

Lumgineer:
That's strange...I don't think I've ever seen that before and nobody else has reported it. I imagine you've already done so, but have you already done a fresh rebuild of the simulator under this branch? (e.g. ``catkin clean`` followed by ``catkin build``)

Hao12450:
> That's strange...I don't think I've ever seen that before and nobody else has reported it. I imagine you've already done so, but have you already done a fresh rebuild of the simulator under this branch? (e.g. `catkin clean` followed by `catkin build`)

Sorry, I forgot to do a fresh rebuild, It shows now, will give you the new result later!

AstroStucky:
In test 1, I started ticked `disconnect_battery_nodes` when `battery_nodes_to_disconnect` was set to 2. The number of prognosers correctly became 2. I then increased it the value to 3, while the flag was still ticked, and the number of prognosers became 1. I believe this fails the following test condition. 

> Confirm that adjusting the value of battery_nodes_to_disconnect after the fault has already been activated does NOT result in any changes to the system.

I noticed decreasing the value of `disconnect_battery_nodes` does not bring any nodes back online.

Lumgineer:
> In test 1, I ticked `disconnect_battery_nodes` when `battery_nodes_to_disconnect` was set to 2. The number of prognosers correctly became 2. I then increased it the value to 3, while the flag was still ticked, and the number of prognosers became 1. I believe this fails the following test condition.
> 
> > Confirm that adjusting the value of battery_nodes_to_disconnect after the fault has already been activated does NOT result in any changes to the system.
> 
> I noticed decreasing the value of `disconnect_battery_nodes` does not bring any nodes back online.

I totally forgot to update that test, that's on me. Mike requested I change this behavior to be more consistent with other faults. It is now supposed to update if you adjust the value while the fault is already activated.

I'll update the test.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -511,39 +558,92 @@ void PowerSystemNode::injectFault (const std::string& fault_name)
...
+                        << " nodes when there are already "
+                        << std::to_string(m_deactivated_models)
+                        << " out of " << std::to_string(NUM_MODELS)
+                        << " disconnected. Behavior has not changed.");
+        ROS_WARN_STREAM_THROTTLE(30, "This warning will repeat periodically "
+                        << "until the fault value is reset to "
+                        << std::to_string(m_deactivated_models)
+                        << ", updated to disconnect more nodes, "
+                        << "or the fault is deactivated.");
+      }
```
AstroStucky, line 634:
```suggestion
      if (m_deactivated_models > dbn_nodes)
      {
        ROS_WARN_STREAM_THROTTLE(30, fault_name << " currently set to "
                        << "disconnect " << std::to_string(dbn_nodes)
                        << " out of " << std::to_string(NUM_MODELS)
                        << " nodes when there are already "
                        << std::to_string(m_deactivated_models)
                        << " out of " << std::to_string(NUM_MODELS)
                        << " disconnected. Behavior has not changed.");
        ROS_WARN_STREAM_THROTTLE(30, "This warning will repeat periodically "
                        << "until the value of \"battery_nodes_to_disconnect\" is reset "
                        << std::to_string(m_deactivated_models)
                        << ", updated to disconnect more nodes, "
                        << "or the fault is deactivated.");
      }
```

AstroStucky (approved):
Code changes are straight forward enough. Just suggested a minor change to a warning message. 

Tests performed as described. 

Nice work!

kmdalal:
15-minute performance test results with RAM-32GB, nproc=12:
 - 4 active models:  Load averages 12.06, 12.18, 8.90.  Core temperatures (0..5) evenly distributed between 85c-92c  (100c is critical).  Power node CPU usage ~400%.
 - 2 active models: Load averages 8.04, 7.97, 7.45.  Core temperatures (0..5) evenly distributed between 86c-95c.  Power node CPU usage ~200%.

I probably should have run the 2-model test first, but ran this one second, with a small gap between, so some of the high temperatures may be carryover.  But I think a rough test is good enough for this purpose.

kmdalal (approved):
I just gave the latest version a quick test.  I see some small areas for improvement in user messages, but we can save this for later work since the merging of this work is critical path for Release 12.   

I also think it is merge-able without any further performance test data, and performance tests can (and I think should) be added to our release tests.  Therefore I'm taking the extra reviewers off this PR, also because everyone is overloaded right now!

Awesome work!

PR #355, opened 2023-08-17
Oceanwater 1095 failures messages communication

Hao12450:
## Linked Issues:
| Jira Ticket 🎟️ | [OCEANWATER-1195](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1195) |

## Summary of Changes
The current failure messages relayed from `ow_lander` to `PLEXIL` are still incomplete. This branch primarily aims to establish a robust communication mechanism between the `ow_faults_detection `package and the `ow_lander` package.

Previously, the Arm action servers treated both `ArmExecutionError` and `ArmPlanningError` in the same manner. Furthermore, there was no linkage with the `ow_faults_detection` package. 

This PR introduces a communication method. Here's an overview of the design (Note: Changes currently apply to only `ArmMoveCartesianGurard` and `ArmMoveCartesian`):

* The Arm action can produce outcomes like `ArmExecutionError`, `ArmPlanningError`, or successful action. When ROS encounters an `ArmPlanningError`, `action.py` will publish an error flag to the ROS topic `arm_faults_interal`.
* `FaultsDetector` continuously publishes `ArmFaultsStatus` and `SystemFaultsStatus` since the simulator's initiation. When it detects an error flag on the topic `arm_faults_internal`, it updates the flags on both `arm_faults_status` and `system_faults_status`.
* In the event of an `arm_execution_error`, the arm will freeze, preventing subsequent action calls. However, if only `arm_goal_error` is detected and the next action is valid, the arm process and updates the error flag on both `arm_faults_status` and `system_faults_status`.

## Test
We need four terminals for testing:

1. `roslaunch ow europa_terminator_workspace.launch`
2. After the simulator fully starts, this can tell once the ros info shows `Ready to take commands for planning group arm.`. In the second terminal run the following command `rostopic echo /arm_faults_status`
3. To check the system faults messages, input the following command in the third terminal `rostopic echo /system_faults_status`
4. The fourth terminal, we want to call the arm actions. `rosrun ow_lander arm_move_cartesian_guarded.py`

* The defaults parameter of `rosrun ow_lander arm_move_cartesian_guarded.py` is invalid so we expected to see the following logs: Screenshot from 2023-08-18 15-35-53
The action returns an error because the input provides an invalid location, hence we see both `rostopic echo /arm_faults_status` and `rostopic echo /system_faults_status` have a value equal to 2, which represent `TRAJECTORY_GENERATION` error in `/arm_faults_status` and `ARM_GOAL_ERROR` in `/system_faults_status`.
* For the next action run the following command with a valid location `rosrun ow_lander arm_move_cartesian_guarded.py -x 1.86 -y -0.63 -z 0.045 -q 1 0.005 0.053 0.029`
We should expect to see the arm start moving in the gazebo desktop, and the expectation logs as below:
Screenshot from 2023-08-18 15-50-11
And both ros topics should change the value back to 0, meaning no error occurred.

### DIFF
--- a/ow_faults_detection/include/ow_faults_detection/FaultDetector.h
+++ b/ow_faults_detection/include/ow_faults_detection/FaultDetector.h
@@ -50,7 +50,7 @@ private:
   void jointStatesFlagCb(const ow_faults_detection::JointStatesFlagConstPtr& msg);
   bool isFlagSet(uint joint, const std::vector<uint8_t>& flags);
 
-  void ArmFaultsInternalCb(const owl_msgs::ArmFaultsStatus::ConstPtr& msg);
+  void armFaultsInternalCb(const owl_msgs::ArmFaultsStatus::ConstPtr& msg);
   
   // Find an item in an std::vector or other find-able data structure, and
   // return its index. Return -1 if not found.
@@ -93,7 +93,7 @@ private:
   ros::Subscriber m_camera_raw_sub;
   ros::Subscriber m_power_soc_sub;
   ros::Subscriber m_power_temperature_sub;
-  ros::Subscriber arm_faults_internal_sub;
+  ros::Subscriber m_arm_faults_internal_sub;
 
   // faults bitflags
   uint64_t m_arm_faults_flags     = owl_msgs::ArmFaultsStatus::NONE;
@@ -106,7 +106,8 @@ private:
   std::vector<unsigned int> m_joint_state_indices;
   bool m_camera_data_pending = false;
   float m_last_soc = std::numeric_limits<float>::quiet_NaN();
-  int trajectory_error_flag = 0; // for arm planning error, msg published by ow_lander
+  // for arm planning error, msg published by ow_lander
+  uint64_t m_arm_faults_internal_flag = owl_msgs::ArmFaultsStatus::NONE;
 };
 
 #endif
--- a/ow_faults_detection/include/ow_faults_detection/FaultDetector.h
+++ b/ow_faults_detection/include/ow_faults_detection/FaultDetector.h
@@ -49,6 +49,8 @@ private:
   // Arm
   void jointStatesFlagCb(const ow_faults_detection::JointStatesFlagConstPtr& msg);
   bool isFlagSet(uint joint, const std::vector<uint8_t>& flags);
+
+  void ArmFaultsInternalCb(const owl_msgs::ArmFaultsStatus::ConstPtr& msg);
   
   // Find an item in an std::vector or other find-able data structure, and
   // return its index. Return -1 if not found.
@@ -91,6 +93,7 @@ private:
   ros::Subscriber m_camera_raw_sub;
   ros::Subscriber m_power_soc_sub;
   ros::Subscriber m_power_temperature_sub;
+  ros::Subscriber arm_faults_internal_sub;
 
   // faults bitflags
   uint64_t m_arm_faults_flags     = owl_msgs::ArmFaultsStatus::NONE;
@@ -103,6 +106,7 @@ private:
   std::vector<unsigned int> m_joint_state_indices;
   bool m_camera_data_pending = false;
   float m_last_soc = std::numeric_limits<float>::quiet_NaN();
+  int trajectory_error_flag = 0; // for arm planning error, msg published by ow_lander
 };
 
 #endif
--- a/ow_faults_detection/src/FaultDetector.cpp
+++ b/ow_faults_detection/src/FaultDetector.cpp
@@ -203,8 +203,8 @@ void FaultDetector::jointStatesFlagCb(const ow_faults_detection::JointStatesFlag
   }
 
   // update faults from ow_lander
-  uint64_t hardward_fault_bit = m_arm_faults_flags & ArmFaultsStatus::HARDWARE;
-  m_arm_faults_flags = (m_arm_faults_internal_flag & ~ArmFaultsStatus::HARDWARE) | hardward_fault_bit;
+  uint64_t hardware_fault_bit = m_arm_faults_flags & ArmFaultsStatus::HARDWARE;
+  m_arm_faults_flags = (m_arm_faults_internal_flag & ~ArmFaultsStatus::HARDWARE) | hardware_fault_bit;
 
   // publish updated faults messages
   publishFaultsMessage(m_arm_faults_msg_pub, ArmFaultsStatus(), m_arm_faults_flags);
--- a/ow_faults_detection/src/FaultDetector.cpp
+++ b/ow_faults_detection/src/FaultDetector.cpp
@@ -148,7 +148,6 @@ bool FaultDetector::isFlagSet(uint joint, const std::vector<uint8_t>& flags)
 void FaultDetector::ArmFaultsInternalCb(const owl_msgs::ArmFaultsStatus::ConstPtr& msg)
 {
   //The actual publish on arm faults status intergrate in jointStatesFlagCb
-  ROS_INFO("arm faults internal error messages receieved");
   trajectory_error_flag = msg->value;
 }
 
@@ -196,11 +195,15 @@ void FaultDetector::jointStatesFlagCb(const ow_faults_detection::JointStatesFlag
   }
 
   // update system faults
-  if (ArmFaultsStatus::NONE == m_arm_faults_flags) {
-    m_system_faults_flags &= ~SystemFaultsStatus::ARM_EXECUTION_ERROR;
-  } else {
-    m_system_faults_flags |= SystemFaultsStatus::ARM_EXECUTION_ERROR;
-  }
+
+  // TODO: The comment out codes were not sure if necessary, but keep these  for now for
+  // reference
+
+  // if (ArmFaultsStatus::NONE == m_arm_faults_flags) {
+  //   m_system_faults_flags &= ~SystemFaultsStatus::ARM_EXECUTION_ERROR;
+  // } else {
+  //   m_system_faults_flags |= SystemFaultsStatus::ARM_EXECUTION_ERROR;
+  // }
   if (PanTiltFaultsStatus::NONE == m_antenna_faults_flags) {
     m_system_faults_flags &= ~SystemFaultsStatus::PAN_TILT_EXECUTION_ERROR;
   } else {
@@ -211,8 +214,10 @@ void FaultDetector::jointStatesFlagCb(const ow_faults_detection::JointStatesFlag
   // TODO: Currently only trajectory_generation faults under consideration
   if (trajectory_error_flag == ArmFaultsStatus::TRAJECTORY_GENERATION){
     m_arm_faults_flags |= ArmFaultsStatus::TRAJECTORY_GENERATION;
+    m_system_faults_flags |= SystemFaultsStatus::ARM_GOAL_ERROR;
   } else {
     m_arm_faults_flags &= ~ArmFaultsStatus::TRAJECTORY_GENERATION;
+    m_system_faults_flags &= ~SystemFaultsStatus::ARM_GOAL_ERROR;
   }
   // publish updated faults messages
   publishFaultsMessage(m_arm_faults_msg_pub, ArmFaultsStatus(), m_arm_faults_flags);
--- a/ow_faults_detection/src/FaultDetector.cpp
+++ b/ow_faults_detection/src/FaultDetector.cpp
@@ -147,7 +147,8 @@ bool FaultDetector::isFlagSet(uint joint, const std::vector<uint8_t>& flags)
 
 void FaultDetector::armFaultsInternalCb(const owl_msgs::ArmFaultsStatus::ConstPtr& msg)
 {
-  //The actual publish on arm faults status intergrate in jointStatesFlagCb
+  //arm_faults_internal was published from ow_lander/action.py
+  //m_arm_faults_internal_flag go to jointStatesFlagCb and publishes an updated message to '/arm_faults_status'.
   m_arm_faults_internal_flag = msg->value;
 }
 
@@ -195,7 +196,6 @@ void FaultDetector::jointStatesFlagCb(const ow_faults_detection::JointStatesFlag
   }
 
   // update system faults
-
   if (PanTiltFaultsStatus::NONE == m_antenna_faults_flags) {
     m_system_faults_flags &= ~SystemFaultsStatus::PAN_TILT_EXECUTION_ERROR;
   } else {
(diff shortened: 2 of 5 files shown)


AstroStucky:
> ... which represent TRAJECTORY_GENERATION error in /arm_faults_status and ARM_EXECUTION_ERROR in /system_faults_status.

I believe you mean `ARM_GOAL_ERROR` in /system_faults_status, which would correspond to a value of 2.

ow_lander/src/ow_lander/actions.py
```diff
@@ -533,10 +583,18 @@ def execute_action(self, goal):
       trajectory = self.plan_end_effector_to_pose(intended_pose_stamped)
       self._arm.execute_arm_trajectory(trajectory,
         action_feedback_cb=self.publish_feedback_cb)
-    except (ArmExecutionError, ArmPlanningError) as err:
+    except ArmExecutionError as err:
+      self._arm.checkin_arm(self.name)
```
AstroStucky, line 587: Since `_arm_faults_internal_messages_publish` is not called here, you could produce an error by following a modification of the PR test:
1. `rosrun ow_lander arm_move_cartesian_guarded.py`
2. `arm_move_cartesian_guarded.py -x 1.86 -y -0.63 -z 0.045 -q 1 0.005 0.053 0.029`
3. While ArmMoveCartesian executes, in a 5th terminal call `rosrun ow_lander arm_stop.py`
Now `/system_faults_status` happens to still be true, an **ARM_GOAL_ERROR** has occurred as a result of a stop command, but the `/arm_faults_status` message still shows **TRAJECTORY_GENERATION**, which is an incorrect reason for the fault.

What I am pointing out here is that the loop of the internal fault message has to be closed somewhere in every arm action call. And by closed, I mean at some point in the course of the arm action execution the message has to be updated with either a fail or success result. Or for the purposes of this PR either **NONE** or **TRAJECTORY_GENERATION**. The loop can either be closed at the end of the next called action, as it happens here (except in the case where an execution error has occurred) or it could happen at the beginning of the next called action. It turns out `/system_faults_status` closes this loop at the end of the next called action, but I think that is wrong. 

I feel it makes more sense to close this loop at the beginning of the next called arm action. That way the `/arm_faults_status` will show that no faults are active during arm movement.

ow_lander/src/ow_lander/actions.py
```diff
@@ -72,6 +73,54 @@ def _compute_workspace_shoulder_yaw(x, y):
     yaw += math.asin(l / h)
     _assert_shou_yaw_in_range(yaw)
     return yaw
+    
+class _arm_faults_internal_messages_publish:
+    """Publish the arm failure messages to rostopic 'arm_faults_internal'"""
+    def __init__(self, err):
+      self.err = err
+      self.arm_faults_internal_pub = rospy.Publisher('arm_faults_internal', ArmFaultsStatus, queue_size = 10, latch = True)
```
AstroStucky, line 81: As a member of `ArmActionMixin` we could simply create this publisher once. There is no need to instantiate it everytime we need to publish a message.

ow_lander/src/ow_lander/actions.py
```diff
@@ -72,6 +73,54 @@ def _compute_workspace_shoulder_yaw(x, y):
...
+          self.err_flag = 32
+      elif self.err == 'VELOCITY_LIMIT':
+          self.err_flag = 64
+      elif self.err == 'NO_FORCE_DATA':
+          self.err_flag = 128
+      elif self.err == 'FORCE_TORQUE_LIMIT':
+          self.err_flag = 256
+      else:
+          rospy.loginfo('Arm faults status message is invalid')
+          return
```
AstroStucky, line 106: I suggest we avoid using strings. We already have the bitflags defined in `owl_msgs.msg.ArmFaultsStatus`. It's best not to redefine them locally.

ow_lander/src/ow_lander/actions.py
```diff
@@ -72,6 +73,54 @@ def _compute_workspace_shoulder_yaw(x, y):
...
+    
+      self.message_received = False
+
+      rospy.loginfo(self.err + ' error message sending now')
```
AstroStucky, line 110: I can understand if this was used for your won debugging, but for the published code there is no need to log yet another message about the error here.

ow_lander/src/ow_lander/actions.py
```diff
@@ -72,6 +73,54 @@ def _compute_workspace_shoulder_yaw(x, y):
...
+
+      # To ensure that the subscriber has receives the message, keep publish until all set bits on
+      # self.err_flag also in msg.value now.
+      while not rospy.is_shutdown() and not self.message_received:
+        if self.arm_faults_internal_pub.get_num_connections() > 0:
+          self.arm_faults_internal_pub.publish(value = self.err_flag)
+          self.message_received = True
```
AstroStucky, line 117: I believe ROS already has ways to ensure messages get received, and considering I never see a message local to the machine dropped, I think this loop is unnecessary.

ow_lander/src/ow_lander/actions.py
```diff
@@ -72,6 +73,54 @@ def _compute_workspace_shoulder_yaw(x, y):
...
+        rospy.sleep(1)
+        
+    def _check_communicate_cb(self, msg):
+      if self.err_flag & msg.value == self.err_flag:
+        self.message_received = True
```
AstroStucky, line 122: Never used
```suggestion
```

AstroStucky (changes requested):
There is a bug that occurs when an execution error happens following a planning error. 

I suggested a slightly different approach where the new capabilities are simply put into `ArmActionMixin`.

ow_faults_detection/include/ow_faults_detection/FaultDetector.h
```diff
@@ -49,6 +49,8 @@ class FaultDetector
   // Arm
   void jointStatesFlagCb(const ow_faults_detection::JointStatesFlagConstPtr& msg);
   bool isFlagSet(uint joint, const std::vector<uint8_t>& flags);
+
+  void ArmFaultsInternalCb(const owl_msgs::ArmFaultsStatus::ConstPtr& msg);
```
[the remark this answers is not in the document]
kmdalal: Agreed, the function should start with lower-case to fit our coding standard.

ow_faults_detection/include/ow_faults_detection/FaultDetector.h
```diff
@@ -91,6 +93,7 @@ class FaultDetector
   ros::Subscriber m_camera_raw_sub;
   ros::Subscriber m_power_soc_sub;
   ros::Subscriber m_power_temperature_sub;
+  ros::Subscriber arm_faults_internal_sub;
```
kmdalal, line 96: Optional improvement: individual, named subscriber members are not necessary, since they are only created and never referenced again.  One could replace all these members with a container, e.g. `std::list<ros::Subscriber> m_subscribers;` and then just push the new subscribers into this list.  Easier to maintain.

ow_lander/src/ow_lander/actions.py
```diff
@@ -86,8 +88,11 @@ class ArmStopServer(mixins.ArmActionMixin, ActionServerBase):
   result_type   = owl_msgs.msg.ArmStopResult
 
   def execute_action(self, _goal):
+    # Reset faults messages before the arm start moving
+    self._arm_faults.reset_arm_faults_flags()
     if self._arm.stop_arm():
       self._set_succeeded("Arm trajectory stopped")
+      self._arm_faults.set_arm_faults_flag(ArmFaultsStatus.E_STOP)
```
AstroStucky, line 95: I don't think we have actually agreed we would implement E_STOP as a fault flag, and I don't know if it's right to call our stop command an "emergency stop". @kmdalal does this make sense? Should we keep it?

AstroStucky (changes requested):
This is looking great so far and the tests pass. Well done!

~~Before we call this done however, I would like to see the feature implemented into all arm commands, not just **ArmMoveCartesian** and **ArmMoveCartesianGuarded**.~~

~~What that means is, for those actions that inherit from **ArmActionMixin**, do exactly what you have done for **ArmMoveCartesian** and **ArmMoveCartesianGuarded**. For those actions that inherit from **ArmTrajectoryMixin**, **GrinderTrajectoryMixin**, and **ModifyJointValuesMixin**, you can save some lines of code and instead implement the same solution inside those mixin classes, since they already define an `execute_action` method themselves.~~

AstroStucky:
> This is looking great so far and the tests pass. Well done!
> 
> ~~Before we call this done however, I would like to see the feature implemented into all arm commands, not just **ArmMoveCartesian** and **ArmMoveCartesianGuarded**.~~
> 
> ~~What that means is, for those actions that inherit from **ArmActionMixin**, do exactly what you have done for **ArmMoveCartesian** and **ArmMoveCartesianGuarded**. For those actions that inherit from **ArmTrajectoryMixin**, **GrinderTrajectoryMixin**, and **ModifyJointValuesMixin**, you can save some lines of code and instead implement the same solution inside those mixin classes, since they already define an `execute_action` method themselves.~~

I would like to retract this request so we can get this PR merged before the end of Hao's internship tomorrow. Please still take a look at my other code feedback. Much of it is just naming suggestions.

ow_lander/src/ow_lander/actions.py
kmdalal: My thoughts:
 - I assume this refers to the ArmStop action.
 - That action should cause a goal error in the action that was stopped.
 - E_STOP is an OWLAT hardware concept (a physical stop button) and I don't think we'll ever model it in OceanWATERS.

If some discussion would be helpful, I'm available Friday morning.

AstroStucky: A goal error should already occur when the arm is stopped, although whether it is a `TASK_GOAL_ERROR` or `ARM_GOAL_ERROR` depends on the action that was stopped. 

So it sounds to me like we do not want an arm stop to tick the `E_STOP` flag in `/arm_fault_status`?

Hao12450:
> 

I finished the work on implementing all actions, but I only tested some of them, including arm_move_joints, arm_move_joint, arm_move_joints_guarded.

ow_faults_detection/src/FaultDetector.cpp
```diff
@@ -141,10 +145,15 @@ bool FaultDetector::isFlagSet(uint joint, const std::vector<uint8_t>& flags)
   return false;
 }
 
+void FaultDetector::armFaultsInternalCb(const owl_msgs::ArmFaultsStatus::ConstPtr& msg)
+{
+  //The actual publish on arm faults status intergrate in jointStatesFlagCb
```
kmdalal, line 150: Minor: the comment needs some spelling/grammar correction.

ow_faults_detection/include/ow_faults_detection/FaultDetector.h
AstroStucky: You are correct, but on the other hand I like being able to see exactly what topics a class subscribes to from its header.

AstroStucky (approved):
The code looks great. I went ahead and tested the TRAJECTORY_GENERATION signal for every action for which it is possible to generate, and they all worked. 

Great addition! Thank you Hao. 😃

kmdalal:
The given test worked for me, except for the following anomalies:
 - when the valid arm motion was commanded, `arm_faults_status` became 0 almost immediately, but `system_faults_status` took a while to become 0.  Why did they not become 0 together?
 - the "valid" arm motion actually caused the arm to start bouncing indefinitely on completion

Also, it seems we need to fix the default arguments to `arm_move_cartesian_guarded.py` so that they are valid!  And this should be done for Release 12.  It could be folded into this PR...

Hao12450:
> 

@kmdalal The `arm_faults_status` changed immediately because we reset the error message before the arm started moving, so it doesn't matter if it is invalid or not, as we discussed previously. Then two possible scenarios: the new action is invalid, and the `arm_faults_status` will change back to `2`. The new action is valid, `arm_faults_status` will keep the value equal to `0`.
`system_faults_status` changes once the action is finished, so it will wait until the new action is finished or aborted, and based on the result the value might change. 

> The given test worked for me, except for the following anomalies:
> 
>     * when the valid arm motion was commanded, `arm_faults_status` became 0 almost immediately, but `system_faults_status` took a while to become 0.  Why did they not become 0 together?
> 
>     * the "valid" arm motion actually caused the arm to start bouncing indefinitely on completion
> 
> 
> Also, it seems we need to fix the default arguments to `arm_move_cartesian_guarded.py` so that they are valid! And this should be done for Release 12. It could be folded into this PR...

AstroStucky:
> * when the valid arm motion was commanded, `arm_faults_status` became 0 almost immediately, but `system_faults_status` took a while to become 0.  Why did they not become 0 together?

They are updated via different systems, and I think the development and testing needed to get `system_fault_status` to reset at the beginning of the next called action is more time than we have if we want this PR to be merged today. 

> * the "valid" arm motion actually caused the arm to start bouncing indefinitely on completion

Did it collide with the ground? 

> Also, it seems we need to fix the default arguments to arm_move_cartesian_guarded.py so that they are valid!

The arm move cartesian defaults can be changed in this PR easily enough, but I could also do it in the ClearFaults PR.

Hao12450:
> 




>     * the "valid" arm motion actually caused the arm to start bouncing indefinitely on completion


I have no idea about this.

kmdalal:
> > * when the valid arm motion was commanded, `arm_faults_status` became 0 almost immediately, but `system_faults_status` took a while to become 0.  Why did they not become 0 together?
> 
> They are updated via different systems, and I think the development and testing needed to get `system_fault_status` to reset at the beginning of the next called action is more time than we have if we want this PR to be merged today.

Okay, this can be future work.

> 
> > * the "valid" arm motion actually caused the arm to start bouncing indefinitely on completion
> 
> Did it collide with the ground?

I think so.  I tried again with Atacama, and there was no bouncing, but the grinder "massaged" the ground for a while before the action returned.  The test passes, but I think these argument choices could be improved.

> > Also, it seems we need to fix the default arguments to arm_move_cartesian_guarded.py so that they are valid!
> 
> The arm move cartesian defaults can be changed in this PR easily enough, but I could also do it in the ClearFaults PR.

@AstroStucky : yes please do it in the next PR since this one is closed.

kmdalal (commented):
This PR was already merged but I've now completed my review.  The given test passes, and I'll defer any additional tests to @AstroStucky .  Nice work!

AstroStucky:
> This PR was already merged but I've now completed my review. The given test passes, and I'll defer any additional tests to @AstroStucky . Nice work!

Apologies for the unexpected and early merge. I was under the impression we were to have these PRs wrapped up by the end of Friday, barring any major issues of course.

PR #358, opened 2023-08-25
Option to disable power system for performance

Lumgineer:
## Linked Issues:
| EPIC ⚡ | n/a |
| Jira Ticket 🎟️ | [OCEANWATER-1203](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1203) |
| Github :octocat:  | # |

While it's possible to completely disable all functions within the power system, the vast majority of performance issues arise from using GSAP's prognosers for battery predictions. This PR allows for the removal of all prediction work, while still maintaining publications to external components so as to prevent future issues if said external components rely on these publications. Of course, these publications will not be accurate of a real battery, but the published status should be ideal for preventing any errors if the user is not actively utilizing the power system's new modeling capabilities.

## Summary of Changes
* A new flag has been added to ``ow_power_system/config/system.cfg``: ``enable_power_system``, set to ``true`` by default. If ``false``, the power system skips initialization of the vast majority of its objects and instead runs a minimal loop that does not manage or trigger any predictions. Under this case, the typical battery predictions of RUL/SoC/temperature remain at static values, ignoring any attempted modifications from faults or lander actions. Mechanical power is calculated as normal but not applied to the system.
  * If the user tries to inject a fault while the power system is running in this disabled state, a warning will be printed once to the terminal.
  * Debug printouts will *not* work, since the power system is not running the relevant functions when it is in the minimal disabled state.

## Test
* In ``system.cfg``, set ``enable_power_system`` to ``false``.
* Open up 2 terminals.
  * In terminal 1, build & run the simulator in the OceanWATERS directory. The world shouldn't matter, though I used the ``atacama`` world (``roslaunch ow atacama_y1a.launch``).
  * In terminal 2, check the various rostopics relating to the power system:
    * ``rostopic echo /battery_remaining_useful_life``
    * ``rostopic echo /battery_state_of_charge``
    * ``rostopic echo /battery_temperature``
  * **Confirm the expected behavior for these 5 published topics**:
    * RUL, SoC, and temperature should all remain constant. RUL should be the maximum horizon value, SoC should stay at 100% (``1.0``), and temperature should sit at 20C. These should publish at the same rate set in ``loop_rate`` in ``system.cfg``.
    * Mechanical power should be calculated as per usual and published at the same rate as the ``/joint_states`` topic (typically 50Hz).
  * Navigate to the ``Dynamic Reconfigure`` tab on the bottom of the simulation's RQT window, and from there, to the ``faults`` section.
  * **Attempt to inject various power faults. The first time you do so, the terminal should print a warning that the faults will not work.**

### DIFF
--- a/ow_power_system/src/PowerSystemNode.cpp
+++ b/ow_power_system/src/PowerSystemNode.cpp
@@ -42,6 +42,7 @@ static void printPrognoserOutputs(double rul,
                                   int index);
 static void printMechanicalPower(double raw, double mean, int window);
 static void printTopics(double rul, double soc, double tmp);
+static void printFaultDisabledWarning();
 
 /*
  * The primary function with the loop that controls each cycle.
@@ -74,6 +75,7 @@ void PowerSystemNode::initAndRun()
     }
     // If print_debug was false, then all flags remain false as initialized.
 
+    m_enable_power_system = (system_config.getString("enable_power_system") == TRUE);
     m_active_models = system_config.getInt32("active_models");
     m_max_horizon_secs = system_config.getInt32("max_horizon");
     m_num_samples = system_config.getInt32("num_samples");
@@ -120,6 +122,15 @@ void PowerSystemNode::initAndRun()
                             );
 
   initTopics();
+
+  // If the power system is fully disabled, swap to a basic loop instead.
+  if (!m_enable_power_system)
+  {
+    ROS_INFO_STREAM("Power system disabled via system.cfg! Node will publish "
+                    << "static values and ignore faults.");
+    runDisabledLoop();
+    return;
+  }
 
   // Initialize EoD_events and previous_times.
   for (int i = 0; i < NUM_MODELS; i++)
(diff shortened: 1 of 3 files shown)


ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -121,6 +123,15 @@ void PowerSystemNode::initAndRun()
 
   initTopics();
 
+  // If the power system is fully disabled, swap to a basic loop instead.
+  if (!m_enable_power_system)
+  {
+    ROS_INFO_STREAM("Power system disabled via system.cfg! Node will publish "
```
kmdalal, line 129: Suggest this be a WARN.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -382,6 +393,80 @@ void PowerSystemNode::initAndRun()
...
+  while (ros::ok())
+  {
+    // Create the published messages and set them to their ideal states.
+    owl_msgs::BatteryRemainingUsefulLife rul_msg;
+    owl_msgs::BatteryStateOfCharge soc_msg;
+    owl_msgs::BatteryTemperature tmp_msg;
+
+    rul_msg.value = m_max_horizon_secs;   // Will always be the max possible RUL
+    soc_msg.value = 1.0;                  // Hardcoded 100% battery charge
+    tmp_msg.value = 20;                   // Hardcoded 20C battery temperature
```
kmdalal, line 436: Can `initial_soc` and `initial_temperature` from `system.cfg` be easily used to avoid the hardcoding?

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -945,4 +1051,15 @@ static void printTopics(double rul, double soc, double tmp)
...
                   ", max_tmp: " << std::to_string(tmp));
+}
+
+/*
+ * Prints a warning to the user should they attempt to inject a fault while
+ * the power system is disabled. Only prints once.
+ */
+static void printFaultDisabledWarning()
+{
+  ROS_WARN_STREAM_ONCE("Faults cannot be injected while the power system "
```
kmdalal, line 1062: I suggest that this not be ONCE, unless that would flood the terminal with warnings.

kmdalal:
The given test passed, and I experimented a fair amount.  I ran `ReferenceMission1` during it.

I did a regression test, running `ReferenceMission1` with prognostics on.  I injected the maximum power draw (which was capped at 90W), and disconnected 3 nodes early in the scenario, but found that the battery remained strong (SOC, temperature, RUL) until the end, with RUL at 17818, SOC at 0.93, and temperature 21.6.  I should have seen a much greater degradation, while these values seem more like a nominal power system without faults...

ow_power_system/src/PowerSystemNode.cpp
Lumgineer: Because of the way fault injection works (the relevant function is called every cycle), this warning would print multiple times every cycle (once per activated fault) if it wasn't using the ONCE stipulation. I can switch it to THROTTLE with an arbitrary delay instead, if that's preferred.

kmdalal: Yes, a throttle (of say one minute) sounds good.  This way the user is aware that there's a useless fault injection on, in case they miss that first warning.

Lumgineer:
> The given test passed, and I experimented a fair amount. I ran `ReferenceMission1` during it.
> 
> I did a regression test, running `ReferenceMission1` with prognostics on. I injected the maximum power draw (which was capped at 90W), and disconnected 3 nodes early in the scenario, but found that the battery remained strong until the end, with RUL ending at 17818, SOC at 0.93, and temperature at 21.6. I expected a much greater degradation, while these values seem more like a nominal power system without faults...

I think I brought this up in the original ticket relating to the battery disconnection fault, but disconnecting battery nodes doesn't cause any immediate power loss, and simply increases the rate at which the battery degrades. Whether or not that rate is realistic, I wouldn't know; we'd need to talk to Chetan about it. As for your example though, I'm pretty sure the power system starts around 90-100k RUL, so dropping all the way to under 20k as a result of the highly increased power draw on the one remaining node doesn't seem out of the question to me.

Again though, I don't know exactly what numbers the GSAP predictions should be around. Chetan might know better on if the rate of degradation is realistic or not.

kmdalal:
The main point of my last comment is that I always noticed a much steeper degradation of the battery when a maximal power draw was injected (single and multi cell models).  What I'm seeing now is a very gradual one that seems similar to a fault-free scenario.  I don't have graphs for comparison, but the difference is very visually obvious.  I'll try the same test in noetic-devel....

Lumgineer:
> The main point of my last comment is that I always noticed a much steeper degradation of the battery when a maximal power draw was injected (single and multi cell models).

When you saw this in the past, was the draw within the set limits of the single & multi-cell models? I know previously we had a 30W cap on the single-cell model but that was too high; it didn't seem like it could handle more than 15W without the model breaking (the steep degradation there wasn't valid behavior to my understanding).

The multi-cell model, in my tests, can handle up to around 130W before it breaks in an obvious fashion, but the cap is set to 90W at the moment to be safe (it's also 6x the single-cell cap since each model is 6 cells).

kmdalal:
I retested high power draw in noetic-devel and found no difference from this branch.  It seems the sim can run indefinitely at max wattage (360 injected, capped at 90) without triggering a power fault.  In other words, we can't effectively inject power faults.  I could have sworn this tested differently when the original PRs were reviewed, and that power faults would be achieved in a reasonable amount of time.  Perhaps those tests were done before the 90W limit was added.   In any case, this is a serious issue but outside this PR.  I'll create a new Jira.

kmdalal (approved):
Although this review exposed a major but unrelated issue, this feature works well and is a nice addition!

Lumgineer:
I think I'm still confused as to the nature of this issue you're describing. When we inject high power draw, we're just adding wattage to the current draw that is sent to GSAP every cycle. The cap is 360W in the RQT window because any injected power draw is evenly distributed amongst the four 6S1P battery models that represent the battery (so while the cap is 360W, the most any one node will receive is 90W). In the past when high power draw was maximized, if the battery quickly degraded into a failure state it was because the GSAP prognoser couldn't handle that high of a draw and the behavior it exhibited was invalid. At least, that's how I understood it.

In my experience, adding power draw greatly affects the RUL of the next prediction, and while SoC isn't immediately impacted, it does drain much more quickly at higher power draws compared to when additional power draw isn't applied. If the concern is that power draw isn't doing anything, I'm pretty sure it is.

Maybe I'm misunderstanding how the system should work here? Either way, it sounds like we should discuss this tomorrow either in a meeting or over email.

kmdalal:
My report about the high power draw is subjective.  I recall it being fairly easy to trigger _power faults_ (in particular low-SOC and thermal) in a reasonable amount of time (say a few minutes) the last time I tested the battery pack model, just by maxing the wattage.  Now I'm not getting one even after running the sim a long time.  

I haven't paid close attention to RUL, but RUL does not trigger faults -- only SOC and temperature (an instantaneous capacity loss, which is rare) do.

This issue is outside this PR, which I think can be merged.  We should discuss it soon, but maybe not until Tuesday as we are trying to freeze code on Friday.

If there is a problem or bug here, it can be hotfixed later.

PR #360, opened 2023-08-31
Oceanwater 1099 plan optimizaion for consistency

Hao12450:
## Linked Issues:
| Jira Ticket 🎟️ | [OCEANWATER-1099](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1099) |

## Summary of Changes
* This Pull Request introduces a novel optimization algorithm for arm trajectory planning. Previously, the MoveIt! framework generates a single trajectory for arm or antenna movement based on the provided parameters. However, this trajectory may not always be optimal and therefore might result in inconsistent paths. The new method addresses this limitation by generating multiple trajectories, specifically, five different plans. These plans are then compared, and the trajectory with the minimum Euclidean distance is selected for implementation. This ensures a more optimal, efficient, and consistent arm movement in the applications.

## ROS Test
* Before doing the automatic ROS test, you might need to install the following Python package: `Numpy` `pandas` `fastdtw`. This can be done in the ubuntu terminal `pip install numpy`, `pip install pandas`, and `pip installs fastdtw`. Information about fastdtw can be found here: https://github.com/slaypni/fastdtw. The algorithm is used to compare the similarity of the trajectories.
* The ros test will generate 3 statistical documents, under dir `/home/username/Documents`, if need you can change the dir or file name :
   * `test_results_RRTConnect_testing.txt` contains information about the final result of the test, including the number of outlier trajectories. 
   * `all_trajectories_RRTConnect_testing.csv` the detail positions for all testing trajectories 
   * `3D_trajectories_RRTConnect_testing.png` 3D plot for all trajectories, if we only see one trajectory, that means all of them overlap together.
* On Ubuntu terminal, input `rostest ow_sim_tests test_arm_consistency.test`
   * I set the number of trials as 10, if you want to make the statistical data more reliable, you can manually change this number by adjusting the `test_arm_consistency.py` line 87, changing the value of `num_runs` to 100.

- You should expect to see the arm end in this pose, as the result of the most optimal path: Screenshot from 2023-08-31 09-28-09
- The suboptimal pose look like this: 
Screenshot from 2023-08-31 09-30-33

## Manual Test

- You can ignore the ros test to do the manual test, however, the information will be limited in the animation of the gazebo.
   -  We need two terminal, The first one run the simulator: `roslaunch ow europa_terminator_workspace.launch`
   -  `rosrun ow_lander arm_unstow.py`
   -  After we see the arm_unstow action done, run `rosrun ow_lander arm_move_cartesian.py -x 1.166 -y 0.356 -z 0.588 -q -0.254 0.727 -0.456 0.446` the same parameter was used for ros test, the reason using this sets of parameters, it has a high chance to result in an inconsistent path, so can consider as "worst case".
   - Repeat the unstow and arm_move_cartesian again.

## Note and further testing

- For my testing experience, even though we generated 5 plans using RRTConnect, it is still significantly faster compared to RRTStar. So I suggest for all the action if previously switched to RRTStar for constant behavior (| Jira Ticket 🎟️ | [OCEANWATER-932](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-932) |, it might be more computation efficient switch to RRTConnect now and using the optimization method. 
- I only try one location, if you have experienced non-consistent path generation before, let me know the action and parameter you use. I will do further study on.

### DIFF
--- a/ow_lander/src/ow_lander/trajectory_sequence.py
+++ b/ow_lander/src/ow_lander/trajectory_sequence.py
@@ -83,11 +83,29 @@ class TrajectorySequence:
       = list(self._get_final_joint_positions_of(trajectory))
     self._planning_time_total += planning_time
 
+  def _calculate_trajectory_distance(self, trajectory):
+    distance = 0.0
+    for i in range(len(trajectory.joint_trajectory.points) - 1):
+      point1 = trajectory.joint_trajectory.points[i].positions
+      point2 = trajectory.joint_trajectory.points[i + 1].positions
+      segment_distance = sum((x - y)**2 for x, y in zip(point1, point2)) ** 0.5
+      distance += segment_distance
+    return distance
+  
   def _plan(self):
     """Calls on MoveIt to plan the next trajectory of the sequence. If no
     trajectory is provided, the plan is constructed from
     """
-    success, trajectory, planning_time, error_code = self._group.plan()
+    best_distance = float('inf')
+    for i in range(5):
+      success_, trajectory_, planning_time_, error_code_ = self._group.plan()
+      distance = self._calculate_trajectory_distance(trajectory_)
+      if distance < best_distance:
+        best_distance = distance
+        success = success_
+        trajectory = trajectory_
+        planning_time = planning_time_
+        error_code = error_code_
     if not success:
       raise ArmPlanningError(
         f"MoveIt planning failed with error code: {error_code}")
--- a/ow_sim_tests/scripts/test_arm_consistency.py
+++ b/ow_sim_tests/scripts/test_arm_consistency.py
@@ -57,6 +57,7 @@ def _plot_helper(trajectories):
   ax.set_zlabel('Z')
   plt.title('3D Trajectories')
   plt.legend()
+  # TODO change the saving location in your local directory
   fig.savefig('/home/honghao/Documents/3D_trajectories_RRTConnect_testing.png')
   
 class TestArmConsistency(unittest.TestCase):
@@ -129,6 +130,7 @@ class TestArmConsistency(unittest.TestCase):
     outlier_indices, threshold = _dtw_helper(trajectories, num_runs)
     total_outliers = len(outlier_indices)
     try:
+      # TODO change the saving location in your local directory
       with open('/home/honghao/Documents/test_results_RRTConnect_testing.txt', 'w') as f:
         f.write("Total number of runs: {}\n".format(num_runs))
         f.write("Outliers found at runs: {}\n".format(outlier_indices))
@@ -144,6 +146,7 @@ class TestArmConsistency(unittest.TestCase):
     result = pd.concat(dfs, axis = 1)
 
     try:
+      # TODO change the saving location in your local directory
       result.to_csv('/home/honghao/Documents/all_trajectories_RRTConnect_testing.csv')
       print("File saved successfully")
     except Exception as e:
--- a/ow_sim_tests/scripts/test_arm_consistency.py
+++ b/ow_sim_tests/scripts/test_arm_consistency.py
@@ -26,7 +26,7 @@ TEST_NAME = 'arm_consistency'
 common_dir = os.path.expanduser('~/Documents')
 fig_3d_path = os.path.join(common_dir, '3D_trajectories_RRTConnect_testing.png')
 result_path = os.path.join(common_dir, 'test_results_RRTConnect_testing.txt')
-data_path = os.path.join(common_dir, 'all_trajectories_RRTConnect_testing.csv.txt')
+data_path = os.path.join(common_dir, 'all_trajectories_RRTConnect_testing.csv')
 roslib.load_manifest(PKG)
 
 def _dtw_helper(trajectories, runs):
--- a/ow_sim_tests/scripts/test_arm_consistency.py
+++ b/ow_sim_tests/scripts/test_arm_consistency.py
@@ -18,8 +18,6 @@ import matplotlib.pyplot as plt
 from mpl_toolkits.mplot3d import Axes3D
 from geometry_msgs.msg import Pose, Quaternion
 from action_testing import *
-from fastdtw import fastdtw
-from scipy.spatial.distance import euclidean
 
 PKG = 'ow_sim_tests'
 TEST_NAME = 'arm_consistency'
@@ -32,13 +30,14 @@ roslib.load_manifest(PKG)
 def _dtw_helper(trajectories, runs):
   baseline_index = 0
 
-  threshold = 10
+  threshold = 2
 
   while not rospy.is_shutdown():
     baseline_trajectory = trajectories[baseline_index]
     outlier_indices = []
     for i, traj in enumerate(trajectories[1:]):
-      distance, _ = fastdtw(baseline_trajectory, traj, dist = euclidean)
+      distance = _dtw_numpy(baseline_trajectory, traj)
+      rospy.loginfo(distance)
       if distance > threshold:
         outlier_indices.append(i+1)
     if len(outlier_indices) >= (runs / 2):
@@ -47,6 +46,20 @@ def _dtw_helper(trajectories, runs):
       break
       
   return outlier_indices, threshold
+
+def _dtw_numpy(traj_1, traj_2):
+  traj_1 = np.array(traj_1).flatten()
+  traj_2 = np.array(traj_2).flatten()
+  m, n = len(traj_1), len(traj_2)
+  DTW = np.full((m + 1, n + 1), float('inf'))
+  DTW[0, 0] = 0
+
+  for i in range(1, m + 1):
+    for j in range(1, n + 1):
+      dist = (traj_1[i - 1] - traj_2[j - 1]) ** 2
+      DTW[i, j] = dist + min(DTW[i - 1, j], DTW[i, j - 1], DTW[i - 1, j - 1])
+  
+  return np.sqrt(DTW[m, n])
 
 def _plot_helper(trajectories):
   fig = plt.figure()
@@ -113,7 +126,6 @@ class TestArmConsistency(unittest.TestCase):
       while not self.client_arm_move_cartesian.wait_for_result(rospy.Duration(0.1)):
         try:
           trans = self.tfBuffer.lookup_transform('base_link', 'l_scoop_tip',rospy.Time(0))
-          print('x location', trans.transform.translation.x)
           trajectory.append([ trans.transform.translation.x, 
                               trans.transform.translation.y, 
                               trans.transform.translation.z])
--- a/ow_sim_tests/scripts/test_arm_consistency.py
+++ b/ow_sim_tests/scripts/test_arm_consistency.py
@@ -62,7 +62,6 @@ def _plot_helper(trajectories):
   ax.set_zlabel('Z')
   plt.title('3D Trajectories')
   plt.legend()
-  # TODO change the saving location in your local directory
   fig.savefig(fig_3d_path)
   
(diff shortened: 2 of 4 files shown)


ow_lander/src/ow_lander/trajectory_sequence.py
```diff
@@ -83,11 +83,29 @@ def _append_trajectory(self, trajectory, planning_time):
...
+  
   def _plan(self):
     """Calls on MoveIt to plan the next trajectory of the sequence. If no
     trajectory is provided, the plan is constructed from
     """
-    success, trajectory, planning_time, error_code = self._group.plan()
+    best_distance = float('inf')
+    for i in range(5):
+      success_, trajectory_, planning_time_, error_code_ = self._group.plan()
+      distance = self._calculate_trajectory_distance(trajectory_)
```
kmdalal, line 102: My Python isn't strong, but I'm not seeing what would make this call return possibly different values.  Everything looks deterministic and the same arguments are used all 5 times.  The variable `i` is just a counter.  I'm guessing that I'm missing something...

kmdalal:
Minor comment on the description where it says "Previously, the MoveIt! framework generates a single trajectory for arm or antenna movement ..." : I didn't think there was any planning for antenna movement; I think one just commands pan and tilt angles and the joint moves to those.

ow_lander/src/ow_lander/trajectory_sequence.py
Hao12450: Thank you for the reviews. Indeed it runs the same arguments all 5 times, however, each time it runs it will give a different answer. The .plan() function will generate a trajectory by using a planner from OMPL and Moveit! (For the arm default planner is RRTConnect). As I present on Dev day, the planner is probabilistic, which means every time we run the .plan() function, it might give us a different return trajectory.

ow_sim_tests/scripts/test_arm_consistency.py
```diff
@@ -0,0 +1,160 @@
...
+    xs, ys, zs = traj[:, 0], traj[:, 1], traj[:, 2]
+    ax.plot(xs, ys, zs)
+
+  ax.set_xlabel('X')
+  ax.set_ylabel('Y')
+  ax.set_zlabel('Z')
+  plt.title('3D Trajectories')
+  plt.legend()
+  # TODO change the saving location in your local directory
+  fig.savefig('/home/honghao/Documents/3D_trajectories_RRTConnect_testing.png')
```
kmdalal, line 61: We can't have a user's directory hardcoded here.  I'll suggest a portable pathname such as `~/.ros/...` or `~/Desktop/...`.   There are several more of these in this file.

kmdalal:
Regarding the test instruction "Change the save file local directory ...", it would be have been good to explain how to do this.  I figured out later that the script needs editing in 3-4 different places.  My suggestion in the code would eliminate this step.

kmdalal:
The test fails for me, and I'm not sure why.  Here's the log:
[test1.log](https://github.com/nasa/ow_simulator/files/12491631/test1.log)

I'll try again.

Hao12450:
> Minor comment on the description where it says "Previously, the MoveIt! framework generates a single trajectory for arm or antenna movement ..." : I didn't think there was any planning for antenna movement; I think one just commands pan and tilt angles and the joint moves to those.

It should work for antenna movement as well since it also uses Movelt! I will double-check. The pan tile movement is a kind of simple movement system, might not need these steps to ensure consistency.

ow_lander/src/ow_lander/trajectory_sequence.py
kmdalal: Ah I missed that inner call to plan() and so that's obviously the non-deterministic element here.  Thanks.

kmdalal:
The reason for the test failure I reported above is that the script is attempting to open a file that does not exist: `test_results_RRTConnect_testing.txt` .

Hao12450:
> The reason for the test failure I reported above may be that the script is attempting to open a file that does not exist: `test_results_RRTConnect_testing.txt` .

The code should create a file "test_results_RRTConnect_testing.txt".

Hao12450:
> Regarding the test instruction "Change the save file local directory ...", it would be have been good to explain how to do this. I figured out later that the script needs editing in 3-4 different places. My suggestion in the code would eliminate this step.

@kmdalal The most recent commit fixed this issue, now the result file should save under /home/username/Documents, you should expect to see 3 files generated after running the ROS test.

kmdalal:
I'm afraid that the test stills fails, now rather quickly.  Here's a log:
[test2.log](https://github.com/nasa/ow_simulator/files/12501624/test2.log)

The problem may be a missing `fastdtw` package, which I thought I successfully installed yesterday.  However, I'm unable to install it today - see my attempt in the log.   It's apparently a Python 3 package and the default Python version on NASA machines is still 2 for security reasons.  I tried doing all this in a virtual Python 3 environment, and that failed in different ways.  @AstroStucky , I'm curious to know if you have these issues on your NASA machine.

ow_lander/src/ow_lander/trajectory_sequence.py
```diff
@@ -83,11 +83,29 @@ def _append_trajectory(self, trajectory, planning_time):
...
+      distance += segment_distance
+    return distance
+  
   def _plan(self):
     """Calls on MoveIt to plan the next trajectory of the sequence. If no
     trajectory is provided, the plan is constructed from
     """
-    success, trajectory, planning_time, error_code = self._group.plan()
+    best_distance = float('inf')
+    for i in range(5):
```
kmdalal, line 100: A named constant, e.g. `num_samples` instead of the literal 5 would be helpful for understanding this loop.

AstroStucky:
This PR looks like it has the potential to solve our arm consistency issues. We may have to find an alternative to the fastdtw package if it's Python 3 only, or maybe since it's a rostest we can make an exception? 

Given that we're coming down to the wire for Release 12, I do not anticipate this will go into Release 12, but none-the-less I hope we can get it merged into noetic-devel sometime soon.

I will provide my test results and feedback by sometime next week.

Hao12450:
> 

@kmdalal @AstroStucky  Can you use Numpy? If we can use Numpy I can make my own algorithm with Numpy only, it will just little bit more computation-heavy. I can work on the implementation this weekend!

AstroStucky:
> @kmdalal @AstroStucky Can we use Numpy? If we can use Numpy I can make my own algorithm with Numpy only, it will just little bit more computation-heavy compare to fastdtw. I can work on the implementation this weekend!

I see no problem with that. Numpy should be fine. I believe ROS depends on it, so any user of OW should already have it. 

Please do not feel rushed on this. Take your time if you need to.

Hao12450:
> > @kmdalal @AstroStucky Can we use Numpy? If we can use Numpy I can make my own algorithm with Numpy only, it will just little bit more computation-heavy compare to fastdtw. I can work on the implementation this weekend!
> 
> I see no problem with that. Numpy should be fine. I believe ROS depends on it, so any user of OW should already have it.
> 
> Please do not feel rushed on this. Take your time if you need to.

Thank you for the waiting, the new version should run the DTW algorithm with numpy only. Also if needed, the plan selection method can be disabled by changing `trajectory_sequence.py` line 100, change `for i in range(5):` to `for i in range(1)`.

ow_sim_tests/scripts/test_arm_consistency.py
```diff
@@ -0,0 +1,174 @@
...
+    self.client_unstow.send_goal(goal_unstow)
+    self.client_unstow.wait_for_result()
+    for i in range(num_runs):
+      rospy.sleep(0.1)
+
+      self.client_arm_move_cartesian.send_goal(goal_arm_move)
+
+      trajectory = []
+      rate = rospy.Rate(10.0)
+      start_time = rospy.Time.now()
```
AstroStucky, line 124: never used
```suggestion
```

Hao12450:
@AstroStucky It looks like since my internship ended, I was removed from the NASA git list and cannot push anything now, should I update in a forked repository?

ow_sim_tests/scripts/test_arm_consistency.py
```diff
@@ -0,0 +1,174 @@
...
+common_dir = os.path.expanduser('~/Documents')
+fig_3d_path = os.path.join(common_dir, '3D_trajectories_RRTConnect_testing.png')
+result_path = os.path.join(common_dir, 'test_results_RRTConnect_testing.txt')
+data_path = os.path.join(common_dir, 'all_trajectories_RRTConnect_testing.csv')
+roslib.load_manifest(PKG)
+
+def _dtw_helper(trajectories, runs):
+  baseline_index = 0
+
+  threshold = 2
```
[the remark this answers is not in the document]
Hao12450: The DTW does the following things: Pick a baseline trajectory, and assume the trajectory we picked is optimal. Compare with all other testing samples in Euclidean distance. Therefore the `threshold` is used to identify whether the trajectory is sub-optimal or optimal: if the difference between the baseline trajectory and the sample trajectory within the threshold, we conclude the sample trajectory is optimal, if the difference is larger than the threshold, we say the sample trajectory is sub-optimal, and consider as outlier. 
However what if the trajectory we pick is sub-optimal, in that case, we should expect that the total number of outliers is larger than half of the total number of the testing samples. if this happens, we need to choose another trajectory as a baseline and run the algorithm again, therefore we have `if len(outlier_indices) >= (runs / 2):
      baseline_index += 1`.

ow_sim_tests/scripts/test_arm_consistency.py
```diff
@@ -0,0 +1,174 @@
...
+
+PKG = 'ow_sim_tests'
+TEST_NAME = 'arm_consistency'
+common_dir = os.path.expanduser('~/Documents')
+fig_3d_path = os.path.join(common_dir, '3D_trajectories_RRTConnect_testing.png')
+result_path = os.path.join(common_dir, 'test_results_RRTConnect_testing.txt')
+data_path = os.path.join(common_dir, 'all_trajectories_RRTConnect_testing.csv')
+roslib.load_manifest(PKG)
+
+def _dtw_helper(trajectories, runs):
```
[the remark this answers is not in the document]
Hao12450: `_dtw_numpy(traj_1, traj_2) `compare the two input trajectory, the` _dtw_helper `do the comparison for all the sample trajectory that we collected and output indices for outliers. 
Do you have any suggestions for the name here? Or should I integrate the code in `test_01_multiple_runs`.

AstroStucky:
> @AstroStucky It looks like since my internship ended, I was removed from the NASA git list and cannot push anything now, should I update in a forked repository?
> 
> Thank you so much for your review!

Ah yea, that makes sense. I was hoping your removal might be delayed for longer . 

Yes, as you suggest you can still make contributions from your own fork of ow_simulator. [This section of our README describes how to in some detail](https://github.com/nasa/ow_simulator#contributing), except do not make the pull request against `master`, make it against `noetic-devel`. Unfortunately this may require a new pull request to be created.

ow_sim_tests/scripts/test_arm_consistency.py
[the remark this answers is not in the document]
AstroStucky: So is the units of threshold meters?

AstroStucky:
If the new fork requires new pull requests, I suggest you keep this one open for now even if it has to exist in parallel with the other one, just so we can finish our discussions in this pull request before moving onto the next.

kmdalal:
I'm staying on the sidelines for now, since I already did a preliminary review a while back.  Please re-request my review when all outstanding code changes are complete.

Hao12450:
Please see the latest PR I made on the fork repository. Thank you so much for the reviews.

PR #361, opened 2023-08-31
ClearFaults Action

AstroStucky:
## Linked Issues:
Jira Ticket 🎟️ | [OCEANWATER-1198](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1198) |

## Summary of Changes
* Refactoring of how action servers interface with faults. 
  * Shortened name of faults_interface.py to faults.py
* **Fault Behavior Changes**:
  * GOAL_ERROR faults in `/system_faults_status` will no longer automatically clear when the next action in the group is called.
  * **FaultClear** action has been added to facilitate clearing of only GOAL_ERROR faults.
  * Subsystem faults (e.g. `/arm_faults_status`) are cleared at the start of the next action called in that subsystem.
    * Presently, only TRAJECTORY_GENERATION in `/arm_faults_status` has been implemented, and the other subsystem messages (`/camera_faults_status` and `/pan_tilt_fault_status`) never change. 
  * An action called within a group that is faulted will be rejected. While the log message use the language "rejected", to actionlib the action is merely ABORTED since we use SimpleActionServer at present. 

## Test (quick, non-comprehenisve)
This test will be thorough about all possible fault behaviors with only a single action. The later test will be comprehensive of all actions. All reviewers, please do this test. 
1. Launch any world.
2. In 2nd terminal `rostopic echo /system_faults_status`
3. In 3rd terminal `rostopic echo /arm_faults_status`
4. Call an invalid ArmMoveCartesian `rosrun ow_lander arm_move_cartesian.py`. 
5. Wait a little. Simulator output should be (nothing's changed since previous versions)
```
[INFO:/lander_action_servers] ArmMoveCartesian action started
[ERROR:/move_group] arm/arm: Unable to sample any valid states for goal tree
[ WARN:/move_group] Timed out
[ WARN:/move_group_commander_wrappers_1693511951616356524] Fail: ABORTED: TIMED_OUT
[ERROR:/lander_action_servers] ArmMoveCartesian: Aborted - MoveIt planning failed with error code: val: -6                                                                                    
[INFO:/lander_action_servers] ArmMoveCartesian action complete
```
6. Both your `/system_faults_status` and `/arm_faults_status` output should show a value of 2. The equality of these values is coincidental: one indicates an ARM_GOAL_ERROR and the other indicates a TRAJECTORY_GENERATION fault.
7. Call another ARM group action `rosrun ow_lander arm_unstow.py`. Immediately simulator output should be 
```
[...]

PR #363, opened 2023-09-01
goal errors are cleared at the beginning of the next command

AstroStucky:
This PR spawned from discussion in #361

## Summary of Changes
* Reset the relevant action group GOAL_ERROR flag at the beginning of an action. 

## Test
Launch any simulator world
1. Monitor the system faults topic `rostopic echo /system_faults_status` and the arm faults topic `rostopic echo /arm_faults_status` in two separate terminals. 
2. Call an invalid ArmMoveCartesian `rosrun ow_lander arm_move_cartesian.py`
3. The command will fail. Both topics should show a value of 2. 
4. Call a valid TaskGrind `rosurn ow_lander task_grind.py`
5. The system fault will remain 2, because we are calling an action in a different group (TASK), but arm faults will be reset to 0 at the beginning of the TaskGrind because it uses the arm. Let the TaskGrind complete.
6. Call a valid ArmMoveCertaisn `rosrun ow_lander arm_move_cartesian.py -x 1.8 -z 0.5`
7. System faults will become 0 immediately. The action will complete and both topics should both now be 0.

### DIFF
--- a/ow_lander/src/ow_lander/server.py
+++ b/ow_lander/src/ow_lander/server.py
@@ -138,7 +138,7 @@ class ActionServerBase(ABC):
                    "This should never happen!")
       return
     rospy.loginfo(f"{self.name} action started")
-    # reset execution fault flag by sending a PENDING state
+    # reset execution fault flag by sending a SUCCEEDED state
     # HACK: The SUCCEEDED state does not reflect the current state of the action
     #   that has been called. It cannot have succeeded yet because it has yet to
     #   be attempted here. This is a hack. ActionGoalStatus.msg is effectively
--- a/ow_lander/src/ow_lander/server.py
+++ b/ow_lander/src/ow_lander/server.py
@@ -138,6 +138,14 @@ class ActionServerBase(ABC):
                    "This should never happen!")
       return
     rospy.loginfo(f"{self.name} action started")
+    # reset execution fault flag by sending a PENDING state
+    # HACK: The SUCCEEDED state does not reflect the current state of the action
+    #   that has been called. It cannot have succeeded yet because it has yet to
+    #   be attempted here. This is a hack. ActionGoalStatus.msg is effectively
+    #   interpreted by ow_fault_detector as a boolean array because an element
+    #   can either be SUCCEEDED or ABORTED. To reset system_faults_status goal
+    #   error flags to 0 at the beginning of an action, we broadcast SUCCEEDED.
+    self.__publish_state(actionlib_msgs.msg.GoalStatus.SUCCEEDED)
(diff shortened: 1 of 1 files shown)


kmdalal:
Unrelated to this PR, but the default invocations of `arm_move_cartesian.py` and all the action scripts should be valid, and we should make this so for Release 12...

ow_lander/src/ow_lander/server.py
```diff
@@ -138,6 +138,14 @@ def __on_action_called(self, goal):
                    "This should never happen!")
       return
     rospy.loginfo(f"{self.name} action started")
+    # reset execution fault flag by sending a PENDING state
```
kmdalal, line 141: Is the rest of this paragraph consistent with this first sentence, which says PENDING state?

kmdalal (approved):
The given test passes, but I didn't do any more.  I don't fully understand the logic of the code change, but it sounds reasonable so will leave it to your judgement.

AstroStucky:
> Unrelated to this PR, but the default invocations of `arm_move_cartesian.py` and all the action scripts should be valid, and we should make this so for Release 12...

The reason ArmMoveCartesian, ArmMoveCartesianGuarded, and other actions do not have zero XYZ defaults is because:
    **A)** From the user perspective, there is not a whole lot of use in a default move cartesian position. If you use the action, there is a presumption that you have a very particular pose in mind you would like the arm to go to. We, as developers and testers of the actions, are perhaps the only ones that use it without much care for where the arm winds up. 
    **B)** It makes the use of the relative flag harder. Currently if you merely want to move downward by 20cm it is as simple as `rosrun ow_lander arm_move_cartesian.py -r -z -0.2`. If there are defaults for x, y, and z, you would have to explicitly state which coordinates should have zero translation `rosrun ow_lander arm_move_cartesian.py -r -x 0 -y 0 -z -0.2`

It's my opinion that the actions are more useful to the user when X, Y, Z default to zero. Keep in mind, this is only particular to how the action client script is defined and has nothing to do with the action definition itself. We could discuss this further and come up with a different argparse interface that still satisfies the usability of **B**, but I still think my point in **A** is valid and correct.

ow_lander/src/ow_lander/server.py
AstroStucky: Whoops. I originally planned to reset it with a PENDING state. It does not ultimately matter given how ActionGoalStates works. I

Good catch. This is fixed.

AstroStucky, line 141:
```suggestion
    # reset execution fault flag by sending a SUCCEEDED state
```

PR #364, opened 2023-09-02
New dig model

AstroStucky:
## Linked Issues:
EPIC ⚡ | [OCEANWATER-1146](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1146)

## Sibling PR
[ow_autonomy #126](https://github.com/nasa/ow_autonomy/pull/126)

## Reviewing
No need to review code. This is an epic branch merge, and all of the code here has already been thoroughly reviewed. 

## Summary of Changes
* All dig trajectories (TaskGrind, TaskScoopLinear, and TaskScoopCircular) for greater precision. 
* For all dig trajectories the dig point is always the center of the trench. 
* Various improvements to the balovnev model so that a consistent dig force is applied. 

## Caveats
The behavior of the scoop when dig force is too high for the scoop to advance is not ideal. It tends to get pushed deeper into the terrain, but ideally it should just stall. This could be solved as a Release 12 hotfix. Ideally you avoid scooping too deep, but more realistic behavior when this happens would be preferred. 

## Test 1
The sibling PR is unfinished, but this test can be performed without it.
1. Open any world.
2. Call a grind in any location `rosrun ow_lander task_grind.py`
3. The resulting trajectory won't look massively different, but it is more accurate than previously and the dig point is always in the center of the trench. 
4. Call a TaskScoopLinear `rosrun ow_lander task_scoop_linear.py`
5. The trajectory will start with a circular portion, then a linear portion, then a circular portion. The linear portion now correctly matches the length provided by the user. The depth is much more shallow.
6. Dig slightly deeper. The default is 0.2, if you command 0.06 you will notice that the additional depth is roughly 0.04 as one should expect `rosrun ow_lander task_scoop_linear.py -d 0.06`. 
7. Try a perpendicular trench in a different location. `rosrun ow_lander task_grind.py --perpendicular -x 1.7 -y 0.5`
8. Call a TaskScoopCircular in the exact same spot, note that the x, y coordinates line up perfectly between TaskGrind and scoop action calls. `rosrun ow_lander task_scoop_circular --perpendicular -x 1.7 -y 0.5`.
9. The resulting dig trajectory will look perfectly circular, and like TaskScoopLinear will dig much more shallow than in previous versions.

## Test 2 - Autonomy
See the sibling PR for tests required for the autonomy portion of this PR...


kmdalal (approved):
Test passed and looked great!  Skipping code review per suggestion.

PR #365, opened 2023-09-06
More regolith for shallow digs

AstroStucky:
The new dig model means we dig more shallow trenches at a time. This change ensures we still see a some regolith spawn out of those more shallow digs

## Summary of Changes
* 5x as many regolith particles will spawn for the same amount of dug volume.
* Updates to sample_collection.py, so it works well with the new dig model. 

## Test
Run `rostest ow_sim_tests sample_collection.test ignore_action_checks:=True` and ensure it passes.

### DIFF
--- a/ow/launch/common.launch
+++ b/ow/launch/common.launch
@@ -90,7 +90,7 @@
     <!-- model that gets spawned into scoop -->
     <param name="regolith_model_uri" type="string" value="model://sphere_2cm"/>
     <!-- cubed meters that must be removed from terrain before a regolith particle is spawned -->
-    <param name="spawn_volume_threshold" type="double" value="1e-3"/>
+    <param name="spawn_volume_threshold" type="double" value="2e-4"/>
     <!-- horizontal spacing between each consecutive regolith spawned -->
     <param name="spawn_spacing" type="double" value="0.02"/>
   </node>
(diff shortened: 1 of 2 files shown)


AstroStucky:
@kmdalal This is now ready for testing. The sample collection test should be sufficient to test this branch.

kmdalal:
The result looks great, but one test failed:
```
[ow_sim_tests.rosunit-sample_collection/test_01_arm_unstow][passed]
[ow_sim_tests.rosunit-sample_collection/test_02_task_grind][passed]
[ow_sim_tests.rosunit-sample_collection/test_03_task_scoop_linear][passed]
[ow_sim_tests.rosunit-sample_collection/test_04_task_discard_sample][FAILURE]---
0.20350989850799858 not less than or equal to 0.16 : Regolith fell out of scoop!
```

AstroStucky:
Considering it now produces more regolith than previously, I am considering making the test less strict. Like maybe I could allows some regolith to fall out of the scoop, so long as it retains 90% of what was dug? I don't know. What are your thoughts?

In any case, there is some amount of randomness inherit to this test, and I am not terribly concerned about a single regolith that fell out of the scoop even though that did lead to a test failure.

AstroStucky:
Side question, how many regolith particles did you see at once, and how was your performance while they were active?

kmdalal:
> Side question, how many regolith particles did you see at once, and how was your performance while they were active?

I didn't count, but it was _many_, the most I've ever seen by far.  Performance was good.

I agree with your earlier comment that perfect retention isn't important at this stage.

AstroStucky:
Just FYI, I think the changes required to let sample collection monitor for a percentage of regolith retained are too large to do in this branch, and very tangential to its purpose. I would prefer to have another PR for that work. 

Please let me know if there is anything else needed, besides that, for an approval.

kmdalal (approved):

kmdalal:
Nice simple improvement!

PR #366, opened 2023-09-07
OW-1209: Artifical Power Faults

Lumgineer:
This PR is critical for Release 12 testing of Fault Detection & Injection, as it allows the user to easily (though artificially) trigger conditions where power faults should occur.

## Linked Issues:
| EPIC ⚡ | N/A |
| Jira Ticket 🎟️ | [OCEANWATER-1209](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1209) |
| Github :octocat:  | # |

## Summary of Changes
* The user can now inject low state-of-charge, instantaneous capacity loss, and thermal failure. These faults each modify the outputs of GSAP post-prediction, and thus do not interact with GSAP at all. 
  * This means their behavior is unrealistic, particularly in regards to RUL (which is not affected at all) and when removing the fault (the predictions immediately return to pre-fault values as if nothing had occurred). Still, it allows users to test for power fault behavior, which is important for Release 12 as the new power system can no longer feasibly reach a state where a power fault would naturally be triggered otherwise. The dramatically increased battery life means dramatically increased time before power starts to run low.

## Test
* Open up four terminals:
  * The first terminal will run the simulation. Build & run the simulator in the OceanWATERS directory. The world shouldn't matter, though I used the ``atacama`` world (``roslaunch ow atacama_y1a.launch``).
  * Run the following command in the second terminal to monitor published state of charge:
    * ``rostopic echo /battery_state_of_charge/value``
  * Run the following command in the third terminal to monitor published battery temperature:
    * ``rostopic echo /battery_temperature/value``
  * Run the following command in the fourth terminal to monitor power fault flags:
    * ``rostopic echo /power_faults_status/value``
* Once the simulation has started, navigate to the ``Dynamic Reconfigure`` tab on the bottom of the RQT window, and from there, to the ``faults`` section. Scroll down until you see the ``power_faults`` subsection that contains three additional faults: ``low_state_of_charge``, ``instantaneous_capacity_loss``, & ``thermal_failure``.
* Individually inject each fault and ensure its behavior lines up as expected:
  * ``low_state_of_charge`` should reduce the published values on the ``/battery_state_of_charge`` topic to under 10%. Additionally, ``/power_faults_status/value`` should increase by 1 while the fault is active.
    * Note that for one or two publications upon activation, ``/power_faults_status/value`` may increase by an additional 2 due to the likely sudden jump in state of charge triggering the threshold for an instantaneous capacity loss fault. It should stabilize to 1 quickly.
  * ``instantaneous_capacity_loss`` should reduce the current charge published by ``/battery_state_of_charge`` by at least 5% of its current state of charge, every cycle. Additionally, ``/power_faults_status/value`` should increase by 2 while the fault is active.
  * ``thermal_failure`` should set set the published value of the ``/battery_temperature`` topic to at least 70. Additionally, ``/power_faults_status/value`` should increase by 4 while the fault is active.
* Once the behavior of all 3 new faults has been individually verified, experiment with activating multiple artificial power faults at once. These 3 faults should all work at the same time if desired, though keep in mind that while they are overriding the outputs from GSAP, other power faults that modify inputs to GSAP (for example, ``high_power_draw`` and ``disconnect_battery_nodes``) will not have an effect on the state of charge & temperature topics. These other faults will still affect GSAP and its predictions as expected, even while these artificial faults are in use. Their effects will just not be published until the artificial faults are removed.

### DIFF
--- a/ow_faults_detection/src/FaultDetector.cpp
+++ b/ow_faults_detection/src/FaultDetector.cpp
@@ -261,7 +261,7 @@ void FaultDetector::cameraRawCb(const sensor_msgs::Image& msg)
 void FaultDetector::powerTemperatureListener(const BatteryTemperature& msg)
 {
   // check for excessive battery temperature
-  if (msg.value > POWER_THERMAL_MAX) {
+  if (msg.value >= POWER_THERMAL_MAX) {
     m_power_faults_flags |= PowerFaultsStatus::THERMAL_FAULT;
   } else {
     m_power_faults_flags &= ~PowerFaultsStatus::THERMAL_FAULT;
--- a/ow_power_system/src/PowerSystemNode.cpp
+++ b/ow_power_system/src/PowerSystemNode.cpp
@@ -20,10 +20,14 @@
 
 using namespace PCOE;
 
-const std::string FAULT_NAME_HPD           = "high_power_draw";
-const std::string FAULT_NAME_HPD_ACTIVATE  = "activate_high_power_draw";
-const std::string FAULT_NAME_DBN           = "battery_nodes_to_disconnect";
-const std::string FAULT_NAME_DBN_ACTIVATE  = "disconnect_battery_nodes";
+// Fault names
+const std::string FAULT_NAME_HPD              = "high_power_draw";
+const std::string FAULT_NAME_HPD_ACTIVATE     = "activate_high_power_draw";
+const std::string FAULT_NAME_DBN              = "battery_nodes_to_disconnect";
+const std::string FAULT_NAME_DBN_ACTIVATE     = "disconnect_battery_nodes";
+const std::string FAULT_NAME_LOW_SOC          = "low_state_of_charge";
+const std::string FAULT_NAME_ICL              = "instantaneous_capacity_loss";
+const std::string FAULT_NAME_THERMAL_FAILURE  = "thermal_failure";
 const int CUSTOM_FILE_EXPECTED_COLS           = 2;
 
 // Error flags.
@@ -86,6 +90,7 @@ void PowerSystemNode::initAndRun()
     m_max_gsap_input_watts = system_config.getDouble("max_gsap_power_input");
     m_loop_rate_hz = system_config.getDouble("loop_rate");
     m_spinner_threads = system_config.getInt32("spinner_threads");
+    m_prev_soc = m_initial_soc;
   }
   catch(const std::exception& err)
   {
@@ -646,7 +651,7 @@ void PowerSystemNode::injectCustomFault()
 /*
  * Handles defined faults within the RQT window (currently only high power draw).
  */
-void PowerSystemNode::injectFault (const std::string& fault_name)
+void PowerSystemNode::injectFault(const std::string& fault_name)
 {
   bool fault_enabled;
   double hpd_wattage = 0.0;
@@ -715,7 +720,7 @@ void PowerSystemNode::injectFault (const std::string& fault_name)
       m_disconnect_battery_nodes_fault_activated = false;
     }
 
-    // Continual beahvior with DBN fault.
+    // Continual behavior with DBN fault.
     if (m_disconnect_battery_nodes_fault_activated && fault_enabled)
     {
       // Get the number of nodes set by the user.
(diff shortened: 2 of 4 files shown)


ow_faults_detection/src/FaultDetector.cpp
```diff
@@ -261,7 +261,7 @@ void FaultDetector::cameraRawCb(const sensor_msgs::Image& msg)
 void FaultDetector::powerTemperatureListener(const BatteryTemperature& msg)
 {
   // check for excessive battery temperature
-  if (msg.value > POWER_THERMAL_MAX) {
+  if (msg.value >= POWER_THERMAL_MAX) {
```
kmdalal, line 264: Minor point, but wasn't the original expression correct?

Also assuming this change is unrelated to the feature at hand.

kmdalal: Never mind, I looked at more code and get it now :)

Lumgineer: Ah, I forgot to mention that in the summary. Ultimately I think it's down to a basic interpretation of the thermal failure threshold: do we want it to trigger when the temperature hits 70C, or do we want it to trigger when the temperature exceeds 70C?

I can change it back if the latter is preferred, but the other power faults use the former's logic (low state of charge triggers when the SoC hits 10%, not when it goes under 10%, & ICL triggers when the total loss in one cycle hits 5% or higher, not when it exceeds 5%), so it's somewhat of a consistency change.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -751,15 +756,119 @@ void PowerSystemNode::injectFault (const std::string& fault_name)
...
+
+
+      ROS_INFO_STREAM(fault_name << " activated!");
+      m_low_soc_activated = true;
+      // Warn the user about the invalid behavior associated with power faults.
+      ROS_WARN_STREAM("Note that artificial power fault injection simply "
+                   << "temporarily overrides GSAP's predictions. RUL will "
+                   << "not react to the fault and the system "
+                   << "will continue without any loss of battery health after "
+                   << "the fault is removed, which is not realistic.");
```
kmdalal, line 781: This repeated message should be factored into one place, and I suggest this rewording: 

"Note that artificial power fault injection simply overrides GSAP's predictions. Remaining Useful Life will not be affected and the system will continue without any loss of battery health after the fault is cleared."

Also, you can break this up into short lines simply by juxtaposing strings, e.g. 
```
"foo" 
" bar" 
```
instead of separating them with `<<`, but this is a minor point.

kmdalal:
Injecting an ICL fault causes successive drops in SOC as expected, and it's good that it bottoms out at a positive value.  But when SOC dropped below 0.1 I expected a _real_ low-SOC fault to get detected, and this didn't happen...

kmdalal:
I noticed that an artificial ICL fault trumps the artificial low-SOC fault when both are injected.  This is okay.

Lumgineer:
When SoC is below 0.1 and ICL conditions are met, the _/power_faults_status_ flag should be 3 (low SoC flips the 1-bit, ICL flips the 2-bit). That's what I observed in my own testing; did you see something else?

The ICL fault indeed cannot go below 0 because it's taking a little less than 95% of the previous SoC value each time rather than subtracting a flat %, but this doesn't mean SoC can't go below 0. The prognoser will still eventually return a negative SoC, but the battery fail check should come in before then and flatline the battery at 0% instead.

kmdalal:
To refine my earlier comment, when I inject a low-SOC fault, and then an ICL fault, system faults status is 3 and SOC is less than 0.09 and decreasing.  This is all expected.

But if I inject an ICL fault first, and then the low-SOC fault, the latter seems ignored: SOC is continues decreasing steadily from 0.94 and system fault status remains 2.

Lumgineer:
Ah, I get it now. I didn't think to test that.

I have a quick idea to address it. I'll boot up the laptop and see if it works, and push the commit in a moment if so.

Lumgineer:
Give it another try, it should work better now.

I'll get to the other PR changes tomorrow morning; I just wanted to get this one out of the way while the idea was fresh in my mind.

kmdalal:
> Give it another try, it should work better now.

Works exactly as expected now!  I tried different combinations.

Lumgineer:
Made the other change now. Unsure if you finished your review from last night or not (so unsure if I should re-request a review).

> Also, you can break this up into short lines simply by juxtaposing strings, e.g.

I could go back through all of the work and modify the other outputs to remove any unnecessary ``<<``'s, but I didn't think that was appropriate for this PR since it's generally unrelated and we're just trying to get this done sooner than later.

kmdalal:
> I could go back through all of the work and modify the other outputs to remove any unnecessary `<<`'s, but I didn't think that was appropriate for this PR since it's generally unrelated and we're just trying to get this done sooner than later.

That was an optional, stylistic suggestion only -- no need to convert any other code.  String juxtaposition is a carryover from ANSI C, and I'm guessing has its origin in the days when 80-character line limits were a universal standard.  (I still try to stick to that for readability).

kmdalal (approved):
Well-done workaround on short notice, and I also tested it with the health monitor demo plan, which can now easily have all its execution paths exercised.  Nice work!

PR #368, opened 2023-09-26
Updated for PLEXIL 6, updated staff list into separate file

kmdalal:
## Summary of Changes
* Updated PLEXIL installation instructions for version 6.
* Updated team lists, put into new file `CONTRIBUTORS.md`.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -38,5 +38,3 @@ Devin Wright
 Yinan Xu
 Honghau Zhen
 
-In addition, a number of external users have made valuable code
-contributions over the years.
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -1,0 +1,42 @@
+OceanWATERS Team
+----------------
+
+K. Michael Dalal (project lead)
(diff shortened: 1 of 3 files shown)


CONTRIBUTORS.md
```diff
@@ -0,0 +1,42 @@
...
+Antoine Tardy
+Pranay Thangeda
+Orkan Umurhan
+Hans Utz
+Uland Wong
+Devin Wright
+Yinan Xu
+Honghau Zhen
+
+In addition, a number of external users have made valuable code
```
mogumbo, line 41: It's common to include these names as well. You could have a separate category if you like.

mogumbo (approved):

CONTRIBUTORS.md
kmdalal: The truth is, only a handful of external users have made contributions, where were all small fixes.  I've just removed this sentence, because it would be work to create and maintain a list here.  If we ever get a substantial external contribution I think we should mention it here.

AstroStucky (approved):
A name was misspelled. Otherwise looks good.

PR #370, opened 2023-10-07
hotfix/oceanwater 1159 fix for DockIngestSample

AstroStucky:
Jira Ticket 🎟️ | [OCEANWATER-1159](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1159)

## Reviewers
@mikedalal Feel free to defer the full review to Terry, seeing as I'm getting this in late before your vacation.
@mogumbo Please test and provide a technical review. 

## Summary of Changes
* DockIngestSample rebuilt to not rely on ContactSensorPlugin
  * _Note_ ContactSensorPlugin is still in use for removing regolith that collides with the terrain, and so will not yet be removed at this time.

This rebuild was necessary. The previous approach to DockIngestSample was over complicated and reliant on ContactSensorPlugin, which appears to be bugged. I dug deep into the Gazebo API, but the contact sensor API will randomly fall into a state where there are clearly contacts present on the collision body, but it reports zero contacts. Unable to see a work around to this bug, and finding the original DockIngestSample code difficult to debug (even as the person who wrote it), I decided to rewrite it from scratch using a different method. 

The new method is to just treat the sample dock as the axis-aligned bounding box that it is and do a containment check with each regolith present in the world. It would have been simpler if I could use tf to transform a regolith's position from the world frame to the sample dock's frame, but [OW-1194](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1194) made that impossible without hard-coding offsets for each and every world into the action. The transformation is done manually using only data from Gazebo (which does not suffer from OW-1194), and I have confirmed it works in different worlds. 

## Test
1. Launch any world in which you can dig.
2. In another terminal `roslaunch ow_plexil ow_exec.launch plan:=TestActions.plx`

Each time DockIngestSample sample is called in the above procedures, sample should be removed from the sample dock, even if it had to cascade into it following after other removed sample, and when all the sample is removed the action will end 3 seconds later. Sample still may remain if it did not roll into the sample dock.

### DIFF
--- a/ow_lander/src/ow_lander/actions.py
+++ b/ow_lander/src/ow_lander/actions.py
@@ -1128,19 +1128,16 @@ class DockIngestSampleServer(ActionServerBase):
     return transformed
 
   def _is_position_in_sample_dock(self, position):
-    # relative_to_dock = FrameTransformer().transform_geometry(
-    #   position, "lander_sample_dock_link", "world")
     relative_to_dock = self._transform_relative_to_dock(position)
     if relative_to_dock is None:
       raise ActionError("Transform of regolith position failed")
-    # perform a simple axis-aligned bounding box check, since in the sample
-    # dock's frame it is a an axis_aligned box centered
+    # perform an axis-aligned box containment check since, in the sample dock's
+    # frame, it happens to be an axis-aligned box
     X_DIM = 0.3 # meters
     Y_DIM = 0.05
     Z_DIM = 0.095
     # NOTE: This value can be found by adding the y-value on line 25 of
     #   lander_sample_dock.xacro to the point assigned to the origin on line 34
-    #   and multiplying the result by negative one
     OFFSET_RELATIVE_TO_FRAME = Point(0.0, -0.33, 0.025)
     p = math3d.subtract(relative_to_dock, OFFSET_RELATIVE_TO_FRAME)
     if (    -X_DIM/2 < p.x < X_DIM/2
--- a/ow_lander/src/ow_lander/actions.py
+++ b/ow_lander/src/ow_lander/actions.py
@@ -1112,25 +1112,22 @@ class DockIngestSampleServer(ActionServerBase):
   def _on_link_states_msg(self, msg):
     self._message_buffer = msg
 
-  def _transform_relative_to_dock(self, position):
-    # FIXME: This must be done manual due to the tf inaccuracy caused by OW-1194
+  def _get_dock_pose(self):
     try:
       i = self._message_buffer.name.index('lander::lander_sample_dock_link')
     except ValueError:
-      rospy.logwarn(
+      raise ActionError(
         "lander::lander_sample_dock_link not found in /gazebo/link_states")
-    dock_pose = self._message_buffer.pose[i]
+    return self._message_buffer.pose[i]
+
+  def _is_position_in_sample_dock(self, position, dock_pose):
+    # transform world frame position to a sample dock frame position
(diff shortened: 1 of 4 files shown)


ow_regolith/src/ContactSensorPlugin.cpp
```diff
@@ -18,12 +18,14 @@ using std::string, std::set, std::vector, std::bind, std::dynamic_pointer_cast,
       std::endl, std::regex_match, std::begin, std::end, std::regex,
       std::make_unique;
 
-const static string NODE_PREFIX = "contact_sensor_";
+const string NODE_PREFIX = "contact_sensor_";
 
-const static string PLUGIN_NAME = "ContactSensorPlugin";
+const string PLUGIN_NAME = "ContactSensorPlugin";
```
kmdalal, line 23: Thanks for removing `static` - I've always preferred this!

kmdalal (approved):
The `TestActions` plan ran to successful completion, though there were 3 regolith left in dock.   My review was light, deferring to @mogumbo for a more thorough review.

ow_lander/src/ow_lander/actions.py
```diff
@@ -1139,7 +1136,7 @@ def _is_position_in_sample_dock(self, position):
     # NOTE: This value can be found by adding the y-value on line 25 of
```
mogumbo, line 1136: Line numbers can change. It would be better to describe where the values are coming from with link or joint names.

ow_lander/src/ow_lander/actions.py
```diff
@@ -1139,7 +1136,7 @@ def _is_position_in_sample_dock(self, position):
...
     #   lander_sample_dock.xacro to the point assigned to the origin on line 34
     OFFSET_RELATIVE_TO_FRAME = Point(0.0, -0.33, 0.025)
```
mogumbo, line 1138: Please do a separate ticket to eliminate this hardcoded transform by moving the origin of the sample dock to the center of the sample dock.

mogumbo (changes requested):

ow_lander/src/ow_lander/actions.py
```diff
@@ -1102,77 +1103,103 @@ class DockIngestSampleServer(ActionServerBase):
...
-      Contacts,
-      self._on_dock_contacts
-    )
-    self._dock_contacts = list()
-    self._ingesting = False # true when action is active
-    self._last_contact_time = 0.0 # seconds
-    self._remove_lock = False # true when dock contacts are being processed
+    self._link_states_sub = rospy.Subscriber("/gazebo/link_states", LinkStates,
+                                             self._on_link_states_msg)
+    self._message_buffer = LinkStates()
```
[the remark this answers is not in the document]
AstroStucky: Done. I also added a wait for it to receive messages. Makes the action more rigorous.

ow_lander/src/ow_lander/actions.py
AstroStucky: Done

ow_lander/src/ow_lander/actions.py
AstroStucky: Agreed. Here's the ticket https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1215

AstroStucky:
Suggestions applied, and additionally the code now handle the case where the action is called before the link_states message has been received.

mogumbo (approved):

PR #371, opened 2023-10-28
Populate material grid from heightmap albedo

AstroStucky:
## Linked Issues:
| Jira Ticket 🎟️ | [OCEANWATER-1170](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1170) |

## Sibling PR
https://github.com/nasa/ow_europa/pull/29

## Summary of Changes
* The material grid is now populated at the start of the simulation based on the pixel directly above it and the value of `reference_color` for a material.
* Materials are now defined with a `visualize_color` which is used only for the grid and dug points visualizations in Rviz. The `color` is was no longer needed.
* A `reference_color` is defined in the plugin's SDF, which is in a world's `model.sdf` file. The `reference_color` is a way  to reinterpret albedo colors as materials. 
* A voxel that lies above the terrain will contain an empty MaterialBlend which results in it not being drawn in the Rviz grid visualization.

## Test
You will want to have 2 terminals open for this test. 
1. Monitor your gazebo log to make sure the plugin initializes correctly. `tail -F ~/.gazebo/server-11345/default.log`
2. Launch atacama_y1a in terminal 2 with Rviz running
3. MaterialDistributionPlugin produces a lot of output that gets mixed in with other gazebo processes. Excluding message from those other processes, it should eventually look like this.
```
(1698450680 426765698) MaterialDistributionPlugin: loading...
(1698450680 598083073) MaterialDistributionPlugin: Material grid centered at (0.65, 0, 0.15) with dimensions 1.3 x 2 x 0.3 meters.
(1698450680 606897368) MaterialDistributionPlugin: Materials database populated with 3 materials.
(1698450680 610247500) MaterialDistributionPlugin: awaiting scene heightmap...
(1698450681 789233168) MaterialDistributionPlugin: Gazebo heightmap acquired.
(1698450681 795985547) MaterialDistributionPlugin: Heightmap albedo texture acquired.
(1698450681 812097119) MaterialDistributionPlugin: Populating grid from heightmap albedo...
(1698450687 736894557) MaterialDistributionPlugin: Grid generation took 5.92472 seconds.
(1698450687 828745851) MaterialDistributionPlugin: Grid visualization now ready for viewing in Rviz.
```
4. Go to Rviz and under **Displays** tick the `material_grid` point cloud to on. You should see a chunky, recolored version of the atacama_y1a terrain. Your FPS may tank in either Rviz, Gazebo, or both while this is being rendered. 
Screenshot_20231027_175555

## Computing Material Concentrations from Heightmap Color
For each pixel of the heightmap's albedo texture that lies within the material grid, the inverse of the RGB-distance is computed between the pixel's RGB value and each material's reference color. The result divided by the summed total of all the inverse color distances computed for all materials in the database is used as the concentration for that material.

To put it in math terms, lets suppose the colors are just position vectors where the red is the x-value, green is the y-value, and blue if the z-value. The concentration for a material of reference color, **C**, underneath an  albedo texture pixel of color, **C**<sub>a</sub> is given by the function _f_(**C**, **C**<sub>a</sub>)

equation

where **C**<sub>1</sub>, **C**<sub>2</sub>, ... ,  **C**<sub>_N_</sub> are the reference colors of all materials within a material database that contains _N_ different materials. 

The piecewise nature is important to avoid a division by zero. What this representation does not capture is the fact that when there is a perfect color match, **C** = **C**<sub>a</sub>, the matching material is given a concentration of `1.0` in the resulting MaterialBlend, but also the other materials are excluded from the blend. This formula ensures concentrations in a material blend always add up to `1.0`.

It's worth noting that whenever, **C** =/= **C**<sub>a</sub>, every material in the database is represented in the voxel even if its concentration is minuscule. This could reduce performance if a large number of materials are added to the database. A lower-bound threshold (_i.e._ `0.05`) could be used here whereby concentrations below this threshold will not be added to the blend and their computed concentration would instead be distributed among the materials whose concentrations were computed to be above that threshold. I chose to stick with the simpler approach for now since I do not think the user is likely to use large enough amounts of materials for this to become a problem.


AstroStucky:
Just removed a file I did not mean to commit. Reviews can continue and shouldn't be affected

ow_materials/include/Materials.h
```diff
@@ -11,16 +11,24 @@
 namespace ow_materials
 {
 
-using MaterialID = std::uint8_t;
-
 struct Color
 {
-  double r, g, b;
+  Color(float red, float green, float blue) : r(red), g(green), b(blue) { };
```
mogumbo, line 16: There's already an ignition::math::Color and Ogre::ColourValue that you could probably use here. Don't know whether it is convenient, though. I prefer ignition::math::Color just because it uses the American spelling which is consistent with the rest of our code.

ow_materials/src/MaterialDistributionPlugin.cpp
```diff
@@ -160,19 +361,17 @@ void MaterialDistributionPlugin::handleCollisionBulk(MaterialBlend const &blend)
...
-      // TODO: define + and * operators for color
-      return Color(
-        {
-          value.r + p.second * m.color.r,
-          value.g + p.second * m.color.g,
-          value.b + p.second * m.color.b
-        }
-      );
+      return Color {
+        value.r + p.second * (m.visualize_color.r - value.r),
```
mogumbo, line 371: ignition::math::Color has arithmetic operators that would simplify this code.

mogumbo (approved):
Looks good overall.
This might be better for a separate PR: the material grid appears to be positioned in front of the center of the lander but should maybe be in front of the center of the shoulder joint. This would allow the arm equal reach to all points on the material grid. It's possible I'm wrong about this if the lander legs interfere with arm reach too much.

ow_materials/include/Materials.h
AstroStucky: I encountered that just recently and felt tempted to switch to it. There's even some code where I have to take an ignition color and convert it to my custom color. It has an alpha channel that won't be used, but I don't that's a real problem. 

I could attempt to switch to ignition color in the next PR.

ow_materials/src/MaterialDistributionPlugin.cpp
AstroStucky: Great to know! I nearly went the route of implementing my own. I think I am now sold on switching to ignition::math::Color

PR #374, opened 2023-11-07
Antenna exceptions  are now correctly handled by pan/tilt actions

AstroStucky:
I noticed while testing https://github.com/nasa/ow_autonomy/pull/133 that pan/tilt actions would not handle their timeout exception resulting in a long error print-out and the action not being set to aborted. 

In master the output is
```
[ERROR:/lander_action_servers] Exception in your execute callback: Timed out waiting for pan value to reach goal.
  File "/opt/ros/noetic/lib/python3/dist-packages/actionlib/simple_action_server.py", line 289, in executeLoop                                 
    self.execute_callback(goal)                                                                                                                
  File "/usr/local/home/tstucky/ow/alpha_ws/src/ow_simulator/ow_lander/src/ow_lander/server.py", line 149, in __on_action_called               
    self.execute_action(goal)                                                                                                                  
  File "/usr/local/home/tstucky/ow/alpha_ws/src/ow_simulator/ow_lander/src/ow_lander/actions.py", line 1259, in execute_action                 
    not_preempted = self.move_pan(goal.pan)                                                                                                    
  File "/usr/local/home/tstucky/ow/alpha_ws/src/ow_simulator/ow_lander/src/ow_lander/mixins.py", line 420, in move_pan                         
    raise AntennaExecutionError(                                                                                                               
ow_lander.exception.AntennaExecutionError: Timed out waiting for pan value to reach goal.   
```

In this branch the output is shorter and does not crash a callback function.

## Summary of Changes
* Antenna exception objects are now handled for pan/tilt actions
* **EDIT** Antenna pan/tilt control loop now breaks when both joint velocities are near enough to zero. This fixes [OW-1105](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1105)

## Test
1. Run the simulation in any world with rqt open so you can set faults. 
2. Call any of the following pan/tilt actions
    * Pan
    * Tilt
    * PanTiltMoveJoints
3. While the antenna is still moving to its destination, trigger a fault for pan/tilt depending which action you have called. 
4. The antenna will stop and you will immediately see one of the following messages.
```
[ERROR:/lander_action_servers] PanTiltMoveJoints: Aborted - Both pan and tilt joints failed to reach their goal position.
```
```
[ERROR:/lander_action_servers] PanTiltMoveCartesian: Aborted - Pan joint failed to reach its goal position.
```
```
[ERROR:/lander_action_servers] PanTiltMoveJoints: Aborted - Tilt joint failed to reach its goal position.
```

_NOTE_ If you are calling a PanTiltMoveJoints action, and only lock one of the joints, then the unlocked joint will continue moving. When the unlocked joint arrives at its goal position, or when it too is locked, only then will the above error message display. 
5. **EDIT** Now unlock the joint. It may jerk very slightly, but otherwise will not continue moving towards its original goal from before the fault injection. 

## Regression Test
@kmdalal In addition to the above test, please run ReferenceMission1.plx on atacama_y1a and make sure there are no issues. I'm asking for this test now because this PR now addresses [OW-1105](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1105), which affected ReferenceMission1 outcome.

### DIFF
--- a/ow_lander/src/ow_lander/actions.py
+++ b/ow_lander/src/ow_lander/actions.py
@@ -1227,7 +1227,7 @@ class PanTiltMoveJointsServer(mixins.PanTiltMoveMixin, ActionServerBase):
   def execute_action(self, goal):
     try:
       not_preempted = self.move_pan_and_tilt(goal.pan, goal.tilt)
-    except ArmExecutionError as err:
+    except (AntennaPlanningError, AntennaExecutionError) as err:
       pan, tilt = self._ant_joints_monitor.get_joint_positions()
       self._set_aborted(str(err), pan_position=pan, tilt_position=tilt)
     else:
@@ -1257,7 +1257,7 @@ class PanServer(mixins.PanTiltMoveMixin, ActionServerBase):
   def execute_action(self, goal):
     try:
       not_preempted = self.move_pan(goal.pan)
-    except ArmExecutionError as err:
+    except (AntennaPlanningError, AntennaExecutionError) as err:
       pan, _ = self._ant_joints_monitor.get_joint_positions()
       self._set_aborted(str(err), pan_position=pan)
     else:
@@ -1284,7 +1284,7 @@ class TiltServer(mixins.PanTiltMoveMixin, ActionServerBase):
   def execute_action(self, goal):
     try:
       not_preempted = self.move_tilt(goal.tilt)
-    except ArmExecutionError as err:
+    except (AntennaPlanningError, AntennaExecutionError) as err:
       _, tilt = self._ant_joints_monitor.get_joint_positions()
       self._set_aborted(str(err), tilt_position=tilt)
     else:
--- a/ow_lander/src/ow_lander/actions.py
+++ b/ow_lander/src/ow_lander/actions.py
@@ -1226,7 +1226,7 @@ class PanTiltMoveJointsServer(mixins.PanTiltMoveMixin, ActionServerBase):
 
   def execute_action(self, goal):
     try:
-      not_preempted = self.move_pan_and_tilt(goal.pan, goal.tilt)
+      not_preempted = self.move(pan = goal.pan, tilt = goal.tilt)
     except (AntennaPlanningError, AntennaExecutionError) as err:
       pan, tilt = self._ant_joints_monitor.get_joint_positions()
       self._set_aborted(str(err), pan_position=pan, tilt_position=tilt)
@@ -1256,7 +1256,7 @@ class PanServer(mixins.PanTiltMoveMixin, ActionServerBase):
 
   def execute_action(self, goal):
     try:
-      not_preempted = self.move_pan(goal.pan)
+      not_preempted = self.move(pan = goal.pan)
     except (AntennaPlanningError, AntennaExecutionError) as err:
       pan, _ = self._ant_joints_monitor.get_joint_positions()
       self._set_aborted(str(err), pan_position=pan)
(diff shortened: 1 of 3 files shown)


mogumbo:
Behavior is strange. I don't see a "Timed out waiting" message until I Ctrl-C the movement script or until I unfreeze the joint. And in each of these cases the motion still completes.

kmdalal:
I've tried twice and experience similar issues as reported by @mogumbo  :
 - I never get the aborted message
 - When I clear the fault after the timeout, the antenna continues moving to its commanded position.

ow_lander/src/ow_lander/actions.py
```diff
@@ -1227,7 +1227,7 @@ class PanTiltMoveJointsServer(mixins.PanTiltMoveMixin, ActionServerBase):
   def execute_action(self, goal):
```
kmdalal, line 1227: Line 1319 of this file needs the same change.

AstroStucky:
> Behavior is strange. I don't see a "Timed out waiting" message until I Ctrl-C the movement script or until I unfreeze the joint. And in each of these cases the motion still completes.

Are you sure you waited long enough? The time out is pretty long. I recommend waiting for at least a minute to be sure.

AstroStucky:
> Behavior is strange. I don't see a "Timed out waiting" message until I Ctrl-C the movement script or until I unfreeze the joint. And in each of these cases the motion still completes.

Movement will complete when you unfreeze the joint, even if the action has already been aborted. I know the cause behind this, but I thought the fix might be better suited as a noetic-devel improvement.

mogumbo:
> > Behavior is strange. I don't see a "Timed out waiting" message until I Ctrl-C the movement script or until I unfreeze the joint. And in each of these cases the motion still completes.
> 
> Are you sure you waited long enough? The time out is pretty long. I recommend waiting for at least a minute to be sure.

Yes. First I was waiting 20 seconds like the test instructions said. Then I tried waiting over a minute.

AstroStucky:
Turns out that timeout is actually 60 seconds, which is huge. This may be why reviewers weren't seeing the timeout message. 

I have been wanting rework the control loop for pan/tilt control so that there is no timeout and a locked joint will abort the action immediately. This will make testing easier. @kmdalal decided to just to that work in this hotfix. So the scope of this PR will be increased. I will let reviewers know when it's good to review again.

ow_lander/src/ow_lander/actions.py
AstroStucky: That's more just a misleading line. It's actually is an `ArmExecutionAction` that it may catch, and that's only because the FrameMixin is typically only used by the arm, so it throws Arm errors. 

So that it doesn't confused readers of the code I went ahead and changed it to the base clase `ActionError`. I should probably clean-up the semantics of FrameMixin and make it more general, but that would affect so many arm actions that I would much rather not attempt such a change in a hotfix.

AstroStucky:
@kmdalal @mogumbo I have added to the list of improvements in this PR, and it is now ready to review again. Please re-read the updated PR post.

kmdalal (approved):
Tested fix using FaultDependenciesDemo plan.  Regression-tested ReferenceMission1.  Works well!  My code review was light.

mogumbo (approved):
Approved, but behavior is a little inconsistent. pan.py and tilt.py will fail immediately if the appropriate joint is locked. But pan_tilt_move_joints.py will not fail if only one joint is locked. It will complete the other joint's motion before failing, giving you time to unlock the locked joing to achieve success. Anyway, not sure if this is a bad behavior; it just seems inconsistent to me.

PR #375, opened 2023-11-16
Oceanwater 1220 correct detected voxels

AstroStucky:
## Linked Issues:
| Jira Ticket 🎟️ | [OCEANWATER-1220](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1220) |
| Jira Ticket 🎟️ | [OCEANWATER-887](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-887) |

## Sibling PR
https://github.com/nasa/ow_europa/pull/31

## Summary of Changes
* Increased queue rates and improved performance so that messages between ow_dynamic_terrain and ow_materials do not drop. If there is a dropped message, a ROS warning will be logged.
* Improved performance and simplified voxel intersection detection. 
* Fixed errors in voxel intersection detection:
  * Differential image coordinates are now correctly transformed to world space coordinates.
  * Modifications on the far edge of the voxel grid no long cause a crash.
  * (ow_dynamic_terrain) The result image was flipped along the y-direction from the diff image.
* Parallelized voxel intersection and bulk blend computations. 
* Increased voxel size from 5mm (which is about the pixel size of the heightmap albedo) to 7mm to further improve performance. 
* Removed plugin implementation in europa_terminator_workspace. The grid was not placed in the correct location, and had no affect on the world anyways since the abledo is only one color. 
* (ow_dynamic_terrain) The gzclient version of the DynamicTerrainVisual plugin no longer publishes a modification differential image. This prevents 2 versions of this message being created with conflicting sequence codes.  

## Test
1. Launch atacama_y1a with rviz. 
2. In another terminal, monitor the Gazebo server log `tail -F ~/.gazebo/server-11345/default.log`
4. In Rviz under Displays, tick `material_grid`. The result should look like this
Screenshot_20231116_140055
5. Untick `material_grid` and tick `dug_points`.
6. Command a TaskGrind. Feel free to provide whatever parameters you want, or none at all. The end of a scoop/grind will result in Rviz populating with dug points and the Gazebo log printing the statistics of a batch of modifications. 

The following is the resulting Rviz view and Gazebo log from running the default TaskGrind.
Screenshot_20231116_140510
```
(1700168697 229649280) Batch of modifications from topic /ow_dynamic_terrain/modification_differential/collision completed.
Total number of modifications in batch was 293
Average processor time per modification was 0.00489569
Average merges per modification was 1644
```
The dug_points visualization extend from the very top of the voxel grid to the very bottom. The reason it extends so deep is because the grinder always digs out a volume 2-times it's own size. The visualization should be mostly uniform, with only a few island voxels on the boundaries, and the top surface should approximate the shape of the terrain. 

Here's another example result from `rosrun ow_lander task_grind.py --perpendicular`, following the previous grind.
Screenshot_20231116_140856
```
(1700168697 229649280) Batch of modifications from topic /ow_dynamic_terrain/modification_differential/collision completed.
Total number of modifications in batch was 293
Average processor time per modification was 0.00489569
Average merges per modification was 1644
(1700168909 774710113) Batch of modifications from topic /ow_dynamic_terrain/modification_differential/collision completed.
Total number of modifications in batch was 83
Average processor time per modification was 0.00475439
Average merges per modification was 1500
(1700168918 210253425) Batch of modifications from topic /ow_dynamic_terrain/modification_differential/collision completed.
Total number of modifications in batch was 164
Average processor time per modification was 0.00335787
Average merges per modification was 1053
```
The valley in the middle exists because the previous TaskGrind command already dug out that part. We see 3 separate instances of the batch statistics message because this grind happened to also modify the visual terrain by going deep enough, and that occurred in 2 separate areas along the trajectory that were far enough apart to cause 2 batch messages to be generated. 

Now for a scooping task. `rosrun ow_lander task_scoop_circular.py --perpendicular --depth 0.06`
Screenshot_20231116_141852
```
(1700169530 207163823) Batch of modifications from topic /ow_dynamic_terrain/modification_differential/visual completed.
Total number of modifications in batch was 90
Average processor time per modification was 0.000883876
Average merges per modification was 281
```

Feel free to try whatever digging commands you want to try and break things. Let me know your results.


ow_materials/src/MaterialDistributionPlugin.cpp
```diff
@@ -345,7 +342,7 @@ void MaterialDistributionPlugin::handleVisualBulk(MaterialBlend const &blend)
        "the bulk contains\n";
   for (auto const &x : blend.getBlendMap())
     s << "\t" << x.second << "%% of " << static_cast<int>(x.first) << "\n";
-  gzlog << s.str() << endl;
+  // gzlog << s.str() << endl;
```
mogumbo, line 345: Is this supposed to be commented out? Will `s` be used for anything at all now?

ow_materials/include/MaterialIntegrator.h
```diff
@@ -49,15 +53,36 @@ class MaterialIntegrator
...
+  boost::asio::thread_pool m_threads;
+
+  void onStatLogTimeout(const ros::TimerEvent&);
+
+  void resetStatLogTimeout();
+
   void onModification(
     const ow_dynamic_terrain::modified_terrain_diff::ConstPtr &msg);
 
+  void integrate(
```
mogumbo, line 83: It would help to have a comment with a colloquial explanation of what integrate does. The same is true for other methods, but integrate seems especially meaty.

mogumbo (approved):

ow_materials/src/MaterialDistributionPlugin.cpp
AstroStucky: They are just debugging messages. they print 5 lines per modification that look like this
```
(1700596155 692425365) handleCollisionBulk: STUBBED
the bulk contains
        69.5007% of 1
        22.7597% of 2
        7.73968% of 0
```
There are a huge number of modifications per second, so they clutter up the gazebo log quite a bit and drown out the batch statistics log. I will add a comment, but would rather leave both disabled for now.

ow_materials/include/MaterialIntegrator.h
AstroStucky: Added some more comments and clarified variable names

PR #376, opened 2023-11-16
Oceanwater 1218 improve lighting realism

mogumbo:
| Jira Ticket 🎟️ | [OCEANWATER-1218](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1218) |

This also requires sibling PR: https://github.com/nasa/ow_europa/pull/30

## Summary of Changes
* Convert light units from lux to W/m^2. Lux is a measure of intensity as perceived by the human eye, which is not so appropriate while we're using simulated cameras for all our images. Lights are now about 3X brighter, which I believe to be a reasonable value based on work on other projects (there's a lot of wiggle-room here depending on specific mission design).
* Use Oren-Nayar reflectance instead of Lambertian. Surfaces now have stronger backscatter and their radiance is slightly view-dependent. All along we had neglected to divide diffuse reflectance by pi. That's fixed now and causes the specular reflection to be considerably brighter relative to diffuse reflection.
* Use more realistic energy_conversion and exposure values. Default exposure went from 0.05 to 0.001.

In lieu of testing, here are some before and after screenshots:
* 
camera_before 
camera_after
* 
terrain_before
terrain_after

### DIFF
--- a/ow_ephemeris/README.md
+++ b/ow_ephemeris/README.md
@@ -14,5 +14,5 @@ OceanWATERS:
  - A config file for `irg_planetary_ephemeris` that sets up a mission site on
    Europa.
  - A script that converts sunlight distance from `irg_planetary_ephemeris` into 
-   lux values for our shaders.
+   irradiance values for our shaders.
 
--- a/ow_ephemeris/scripts/sunlight_values_for_shaders.py
+++ b/ow_ephemeris/scripts/sunlight_values_for_shaders.py
@@ -25,7 +25,7 @@ class SunValues:
     self.listener = tf.TransformListener()
 
     # Must use separate messages or they sometimes interfere with one another
-    self.lux_msg = ShaderParamUpdate()
+    self.flux_msg = ShaderParamUpdate()
     self.visibility_msg = ShaderParamUpdate()
     self.shader_param_pub = rospy.Publisher("/gazebo/global_shader_param", ShaderParamUpdate, queue_size=1)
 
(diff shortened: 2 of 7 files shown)


AstroStucky:
It looks like a major visual improvement! I can see the colors in the albedo more easily. 

When I launch the simulator I get a ton of these messages
```
ServiceCaller.on_refresh_services_button_clicked(): could not get class of service /move_group_commander_wrappers_1700110974441773620/get_loggers:
Unable to communicate with service [/move_group_commander_wrappers_1700110974441773620/get_loggers], address [rosrpc://flynn.ndc.nasa.gov:47773]
```

I've seen them in noetic-devel before, but in this branch far more are printed at the start of the simulation. It does not look related to the changes, so I wonder the messages are related to some process in gzclient that is suspended during startup and the changes in this branch have simply made gzclient take a little bit longer to start up.

AstroStucky:
Does the new lighting make the lens flare more colorful? I remember it was previously mostly red.
flare
Just curious.

AstroStucky:
Unfortunately this branch breaks the ReferenceMission1, 2, and likely EuropaMission PLEXIL plans. All of these plans identify their sample location from an image of the surface, and they all result in this message. 
```
[ERROR:/ow_exec_node] Could not get sample point from IdentifySampleLocation
[ERROR:/ow_exec_node] PLEXIL: IdentifySampleTarget: failed to find one.
[ERROR:/ow_exec_node] PLEXIL: Mission aborted because a sampling location was not found.
```
They all take an identical image of the surface
ReferenceMission1

For comparison, the same image taken by these plans in noetic-devel looks like 
flare-noetic

Now that I have both these images side-to-side, I notice that there is far more ambient light brightening shadows in the image from this branch. Is that correct for a moon with no atmosphere?

mogumbo:
> Does the new lighting make the lens flare more colorful? I remember it was previously mostly red.
It looks the same in noetic-devel.

mogumbo:
> Now that I have both these images side-to-side, I notice that there is far more ambient light brightening shadows in the image from this branch. Is that correct for a moon with no atmosphere?

I have set the lander lights to what I believe are more reasonable values. They're about 3X brighter now, which accounts for most of the light you're now seeing in the shadows. Sorry, I should have put this in the PR description. I'll add it.

kmdalal (approved):
I tried 4 sim worlds and the lighting looks great!  My code review was light.  When this is merged, so must be `https://github.com/nasa/ow_autonomy/pull/137` for `ReferenceMission1` to work.

AstroStucky (approved):
My code review was also light as I'm not very familiar with algorithms being used.

PR #381, opened 2024-02-09
Multiterrain fixes and cleanup

AstroStucky:
## Sibling PRs
[ow_europa #33](https://github.com/nasa/ow_europa/pull/33)

## Summary of Changes
* Worked around a gzserver crash that would occasionally happen when a particle was removed from the world.
  * **Caveat** There appears to be a significantly less common version of this crash, but since it seems to originate from Gazebo code, it's not clear how to work around it. Whereas the crash this PR fixes would happen after about 60 regolith removals, this less common crash took over 2000 regolith removals before it occurred in my testing. It's not ideal, but for now I say we just list it in caveats on our wiki. 
* BulkExcavation messages now show material names for easier reading. 
* Gave appropriate densities for salt and magnesium hydrates. 
* Removed unused material parameters relating to terramechanics.
* Removed unused files and cleaned up some commented-out code. 
* Commented-out code relating to collision modifications, which exclusively support the stubbed feature of grinder terramechanics. Handling grinder modifications were also the most resource intensive, so this ensures better performance on lower end machines. 

## Test
Have ow_simulator and ow_europa on **multiterrain_fixes_and_cleanup** and all other repos on noetic-devel. 
1. Launch atacama_y1a.
2. Have a third terminal open to monitor the ground truth topic. 
```
rostopic echo /ground_truth/material_ingested
``` 
3. Sample the 2nd site sampled in the previous PR test.
```
rosrun ow_lander task_grind.py -x 1.85 -y 0.2 -z -0.14 --perpendicular
rosrun ow_lander task_scoop_circular.py -x 1.85 -y 0.2 -z -0.14 --perpendicular
```
4. This should result in at least 2 regolith particles. Deliver them, and ingest the sample
```
rosrun ow_lander task_deliver_sample.py
rosrun ow_lander dock_ingest_sample.py
```
7. These are the results in my ground_truth topic. 
```
header: 
  seq: 0
  stamp: 
    secs: 1916525066
    nsecs: 642499924
  frame_id: ''
volume: 0.0012000000000000001
composition: 
  - 
    id: 1
    name: "MG_HYDRATES"
    proportion: 0.21510052680969238
  - 
    id: 2
    name: "SALT"
    proportion: 0.4329657554626465
  - 
    id: 0
    name: "ICE"
    proportion: 0.35193371772766113
```
**Note** there are actually far fewer hydrates than I thought were being collected in the test for the previous PR. I trust this result because previously I was simply assuming that hydrates would be the most prevalent as a way to determine which materials goes to an ID. There was in fact no easier way to read the ID value in this message. Because those IDs are based on the ordering of an unordered_map, they do not correspond to the order materials are defined in `materials.yaml`. This illustrates the need for having the material name in the message as well.


kmdalal:
In my first trial, no regolith was collected.  Nothing published on the ground truth topic.  Also, I selected display of Material Grid and Dug Points in RVIZ but did not see the multi-colored volume I'd seen before.

kmdalal:
Same result as my previous comment on my second trial.  In both trials many warnings like the following occur during the grind and scoop.
```
[ WARN]: Visual modification message was dropped! At least 12 message(s) may have been missed.
```

AstroStucky:
I have updated the PR with information about a necessary sibling PR in ow_europa that I had forgotten was necessary. I suspect the omission of the ow_europa changes resulted in the MaterialDistributionPlugin crashing at start-up, which means regolith would not spawn and the material grid visualizations would not be visible in Rviz. 

Please switch your ow_europa branch to multiterrain_fixes_and_cleanup and try the test again.

kmdalal (approved):
Test passes (multiple trials).  Light code review but looked reasonable.  Awesome work!

PR #382, opened 2024-02-13
Oceanwater 812 multicomposite terrain

AstroStucky:
## Linked Issues:
| EPIC ⚡ | [OCEANWATER-812(https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-812) |

## Sibling PR
[ow_europa #34](https://github.com/nasa/ow_europa/pull/34)

## Summary of Changes
* Multi-material terrain implemented for sample collection purposes.
* Full documentation for this feature is in review on branch OCEANWATER-812_multicomposite_terrain of the ow_simulator wiki repository. 

## Test
Up to reviewer discretion. Regression tests, particularly with PLEXIL plans like ReferenceMission1 and EuropaMission are recommended. 

## Caveats
The following will be added to ow_simulator Wiki's caveat section until they have been addressed. 
1. During scooping operations the following message may occur if your machine is unable to keep up with the demands of the multi-material terrain system. 
```
[ WARN]: Visual modification message was dropped! At least 12 message(s) may have been missed.
```
One way to improve performance is to change the size of grid voxels by modifying the value of **cell_side_length** per world that supports this feature (at present only atacama_y1a). This can be done in the world's `model.sdf` where **MaterialDistributionPlugin** is defined. See the **Grid Generation** section of the **Environment Simulation** page in the wiki branch for more details.

2. Unfortunately due to an unforeseen bug, possible originating from ow_dynamic_terrian, this update breaks regolith collection in europa_terminator_workspace.word. After this is merged, atacama_y1a will both be the only world that supports multi-material terrain and the only world that supports regolith collection. This breaks the `sample_collection.test` rostest script.


kmdalal:
In my first run of ReferenceMission1, everything was proceeding normally until the scoop got stuck, and then the simulation froze in a strange state -- the scoop is buried deep, and regolith particles are frozen in mid-air.  I'd never seen anything like this before.  Screenshot attached.
RM1-1

kmdalal:
The previous behavior repeated itself on my second run of ReferenceMission1, and I noticed the path of the scoop is quite off.  It begins a circular movement and then presses down and back, deep into the ground.   And this time Gazebo crashed:
```
[Wrn] [ParticlePool.cpp:230] Regolith model regolith_24 is has no link.
[Msg] GlobalShaderParamPlugin::cacheParams(sunIntensity, 73)
[Msg] GlobalShaderParamPlugin::cacheParams(sunVisibility, 69)
[Wrn] [ParticlePool.cpp:230] Regolith model �� ��^ is has no link.
[ WARN]: Modification message on topic /ow_dynamic_terrain/modification_differential/visual was dropped! At least 1 message(s) may have been missed.
[Msg] GlobalShaderParamPlugin::cacheParams(sunIntensity, 77)
[Msg] GlobalShaderParamPlugin::cacheParams(sunVisibility, 73)
[Msg] GlobalShaderParamPlugin::cacheParams(sunIntensity, 89)
[Msg] GlobalShaderParamPlugin::cacheParams(sunVisibility, 85)
[Wrn] [ParticlePool.cpp:230] Regolith model  is has no link.
gzserver: /usr/include/boost/smart_ptr/shared_ptr.hpp:734: typename boost::detail::sp_member_access<T>::type boost::shared_ptr<T>::operator->() const [with T = gazebo::physics::Link; typename boost::detail::sp_member_access<T>::type = gazebo::physics::Link*]: Assertion `px != 0' failed.
Aborted (core dumped)
[gazebo-2] process has died [pid 60528, exit code 134, cmd 
```
This bug is a show-stopper for merging this branch...

kmdalal:
Tested ReferenceMission1 again on a different, slower laptop.  The same behavior and crash reported above occurred.  In addition, during the Grind there were 3 of the following warnings:
```
[ WARN]: Visual modification message was dropped! At least 1 message(s) may have been missed.
```
Three more of these warnings appeared during the circular scoop, some with this variation:
```
[ WARN]: Modification message on topic /ow_dynamic_terrain/modification_differential/visual was dropped! At least 1 message(s) may have been missed.
```

AstroStucky:
That bug was a good find, and did indicate a huge problem that took me on a journey to fix. 

Firstly, ReferenceMission1, ReferenceMission2, and EuropaMission all attempt to dig too deep. I was surprised to find this out because I thought I already addressed it back when the BalovnevModelPlugin was merged. Fixing this should be simple enough, but I'll leave it for another PR since it's not related. The ideal bite depth per consecutive pass is about 2cm; however, as of the updates I am about to talk about, you can probably get away with 3cm just fine. 

Digging too deep results in scoop stalling behavior which I feel could be improved. The scoop stops forward progression, which makes sense; but also it gets pushed deeper and deeper, which this does not make sense for a stall scenario. The solution is potentially complex, and should be left for another PR. 

What you have uncovered in your tests revealed to me that when stalling occurs it almost always results a simulator crash, and that is what I considered to be the show-stopping bug. User autonomy may command unreasonably deep digs, and it's one thing for that to result in less than perfect physics, but it's unacceptable for it to result in a crash. 

Fixing this crash was not easy, and a lot of different methods were tried to address it. I should note, a backtrace will show that this crash originates from within Gazebo code, but is somehow triggered by the particular way in which I was calling Gazebo API functions. Therefore, my best bet was to diagnose what about my API use was causing the crash. Since digging deep causes particless to spawn rapidly, I first investigated the way in which new particles were spawned. I sought to prevent rapid spawning resulting in particle volume overlap, which resulted in the first improvement to this branch that I intend to keep.

1. Added a throttle to the particle spawn rate to prevent particle overlap. 

That did not stop the crashing. 

I then tried to address the excessive force pushing on the scoop when it dug deep, not due to the Balovnev model, but rather due to the pushback force each particle receives during a dig in order to keep it in the scoop. I never liked the pushback force because it made the BalovnevModelPlugin essentially worthless. I got rid of it. It turns out particles remain in the scoop for the most part if they are given a simple initial velocity. This means for deep digs some particles will fall out at the start, but this does not seem like a big problem to me. 

2. Particles are now kept in the scoop during a dig with an initial velocity. 

Getting back to addressing the crash, when particles spawn rapidly they will also be removed rapidly as the scoop overflows and they collide with the terrain. It turns out this was the real source of the crash, but before discovering exactly the solution to the crash I wound up with one more unrelated improvement.

 3. Particles now deleted by a delete plane far below the terrain.

This required adjusting the particle collide bitmask (and the scoop's) to allow them to collide with the scoop but pass through the terrain. This means particles will no longer bounce on the surface, which is not very dirt-like behavior, but will instead fall through it and be deleted by a large plane 2 meters below. I suspect this is a performance improvement too since collision checks between complex terrain geometry and many many particles are no longer performed. 

Finally I found the solution to the crash. It turns out Gazebo just likes to crash when you delete things. I am not sure of exactly the cause, but I have seen numerous unanswered questions and bug reports about the problem. The one that looked most similar, but not exactly because it results in a slightly different backtrace, can be viewed here. 
https://github.com/gazebosim/gazebo-classic/issues/1448

Although it's a very old Issue, I was able to replicate it in Gazebo 11.14.0.

So the finally improvement that actually fixed the problem

4. Particles are now queued for deletion and RemoveModel is called only ever in the BeforeWorldUpdate event. 

I have extensively tested particle deletion, ReferenceMission1, ReferenceMission2, and EuropaMission with no crashes. Please retest on your own machine. There were also changes made to the sibling PR branch in ow_europa, so please do not forget to update that. 

These changes require a code review too. To view only these changes, select only the most recent 6 commits, or all commits since your last review.

kmdalal:
I've run ReferenceMission1 3 times.  It completes normally, no crashes, though there are various issues.

1. Many (15-40) warnings of the following kind during the Grinds, and somewhat fewer during the Scoops:
```
[ WARN]: Visual modification message was dropped! At least 16 message(s) may have been missed.
```
The average number of dropped messages per warning is around 15.  Minor but weird point: these messages are printed in two different shades of orange.

2. A few warnings of the following kind are often interspersed with the above:
```
[ WARN]: Collision modification message was dropped! At least 28 message(s) may have been missed.
[ WARN]: Visual modification message was dropped! At least 1 message(s) may have been missed.
[ WARN]: Modification message on topic /ow_dynamic_terrain/modification_differential/visual was dropped! At least 1 message(s) may have been missed.
```

3. On one run, the final ScoopCircular failed, though regolith was collected and delivered:
```
[ERROR]: TaskScoopCircular: Aborted - The follow_joint_trajectory action server for arm_controller resulted in error code -5 with the message: j_dist_pitch goal error 0.022555
```

4. On two of the runs, little or no regolith was delivered.  I didn't look closely at the entire scoop though.  On the other run, a large quantity was delivered.

I'll experiment some more -- any tips are helpful.

AstroStucky:
> 2. A few warnings of the following kind are often interspersed with the above:
> 
> ```
> [ WARN]: Collision modification message was dropped! At least 28 message(s) may have been missed.
> [ WARN]: Visual modification message was dropped! At least 1 message(s) may have been missed.
> [ WARN]: Modification message on topic /ow_dynamic_terrain/modification_differential/visual was dropped! At least 1 message(s) may have been missed.
> ```

These warning indicate the machine is not keeping up with terrain modification. The first warning in my quotation may only ever be generated by a grind action, and it originates from ow_dynamic_terrain. It essentially says that the plugin is unable to keep up with the rate at which the script `modify_terrain_grinder_pub.py` is sending messages to it. That rate is essentially 1 modification request per 5mm of movement, which is the same rate as enforced when scooping. Using scooping actions, I carefully selected that rate to be a large as possible without losing trench visual fidelity (the larger the mod rate, the more chunky the resulting trench looks). The reason for the optimization is that if I select too small of a rate, then the machine will be incapable of keeping up with the modification messages, as you are experiencing on yours. 

I did not perform this optimization for the grinder, I simply carried over the same rate I used for the scoop and used it in the grinder's script. I notice that grinder operation moves the grinder end-effector along the terrain at a faster rate than either of the scoop actions move the scoop end-effector at. Because of this, I think that number does not work well for the grinder script, and I may need to select a larger value. I will try and do the same optimization as I did for the scoop, and give that script a different value. Another option is I could set the speed of a grinder action slower, but unfortunately I am not sure how to enforce speed limits in MoveIt planning,

What machine did you perform these tests on? I ask because these message warnings were implemented all the way back in https://github.com/nasa/ow_simulator/pull/375, and I'm wondering why they've only cropped up in testing of this PR. 

> 
> 3. On one run, the final ScoopCircular failed, though regolith was collected and delivered:
> ```
> [ERROR]: TaskScoopCircular: Aborted - The follow_joint_trajectory action server for arm_controller resulted in error code -5 with the message: j_dist_pitch goal error 0.022555
> ```

I believe this is caused by [OW-1099](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1099). I see it happen occasionally too.  Unrelated to this PR as best as I can tell. 

> 
> 4. On two of the runs, little or no regolith was delivered.  I didn't look closely at the entire scoop though.  On the other run, a large quantity was delivered.

Was the gzclient window in a different workspace by chance? This could be caused by an unusual bug I reported a long time ago now where if gzclient is in some kind of sleep state according to the OS, visual terrain modifications by the scoop simply do not occur (see [OW-1082](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1082)) It has to do with any gazebo visual plugin going to sleep whenever gzclient goes to sleep. 

It is also possible some sporadic physics instability simply exploded all the regolith out of the scoop/dock.

AstroStucky:
One more point about the missed message warnings. It is possible these messages are missed on noetic-devel as well, but since noetic-devel does not check the `seq` number and report the discontinuity at present, we wouldn't know.

kmdalal:
From the PR description:

> One way to improve performance is to change the size of grid voxels by modifying the value of cell_side_length per world that supports this feature (at present only atacama_y1a). This can be done in the world's model.sdf where MaterialDistributionPlugin is defined.

I'd like to try this but am having difficulty finding the right `model.sdf` file...which one is it?  And I presume I want to increase the value of `cell_side_length`.

kmdalal:
Perhaps unrelated to this PR, but the most verbose message during grinding and scooping are ones of the form:
```
[Msg] GlobalShaderParamPlugin::cacheParams(sunVisibility, 103)
```
I'm wondering why these occur only during grinding and scooping, and it would be nice to suppress them if they are harmless.  Or add them to our Caveats section.

kmdalal:
Perhaps unrelated to this PR, but it's the first place I'm seeing it: in several runs of ReferenceMission1/2, the arm path for DeliverSample is poor -- the scoop tilts early on and loses a lot of regolith.

kmdalal:
The plans `ReferenceMission2` and `EuropaMission` really should ingest the sample so that we get the ground truth.  I've added these on a complimentary branch in `ow_autonomy` with the same name.

kmdalal:
Would it be easy to disable multi-material at sim startup, so that users not interested in it won't suffer the performance hit?

AstroStucky:
> From the PR description:
> 
> > One way to improve performance is to change the size of grid voxels by modifying the value of cell_side_length per world that supports this feature (at present only atacama_y1a). This can be done in the world's model.sdf where MaterialDistributionPlugin is defined.
> 
> I'd like to try this but am having difficulty finding the right `model.sdf` file...which one is it? And I presume I want to increase the value of `cell_side_length`.

You'll find it here in ow_europa
https://github.com/nasa/ow_europa/blob/62d7f9693b7da23fe1081af2fb87538e73a21836/models/atacama_y1a/model.sdf#L90

You can use Rviz to visualize how chunky your new grid winds up being by ticking the `material_grid` on under the **Displays** pane. 

> Perhaps unrelated to this PR, but the most verbose message during grinding and scooping are ones of the form:
> 
> ```
> [Msg] GlobalShaderParamPlugin::cacheParams(sunVisibility, 103)
> ```
> 
> I'm wondering why these occur only during grinding and scooping, and it would be nice to suppress them if they are harmless. Or add them to our Caveats section.

I feel like we brought this up a long time ago and wound up suppressing one message during particle spawn, but not this one. It comes from here in irg_open
https://github.com/nasa/irg_open/blob/35b449bead5feb050c48c48b4c90ce4c12f054a9/irg_gazebo_plugins/src/IRGGlobalShaderParamPlugin/GlobalShaderParamPlugin.cpp#L229
The information does not seem all that important, and I don't see why it couldn't simply be a gzdebug or gzlog. It would be a single-line PR.

> Perhaps unrelated to this PR, but it's the first place I'm seeing it: in several runs of ReferenceMission1/2, the arm path for DeliverSample is poor -- the scoop tilts early on and loses a lot of regolith.

I've definitely seen this a lot in noetic-devel unfortunately. I don't believe it's this branch. I think adding path planning constraints to MoveIt would fix it, but it's never been high enough priority. I made [OW-1240](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1240) for this.
 
> Would it be easy to disable multi-material at sim startup, so that users not interested in it won't suffer the performance hit?

Unfortunately now that regolith triggers off of material grid interactions if the user turns off the material grid, they also turn off regolith spawning. I considered leaving the previous regolith logic intact for exactly the situation where a user wants no material grid, but still wants regolith, but in interest of getting this feature in I shelved that idea. It's still possible to do, but I for now a user can always just increase the grid's cell side length until performance improves.

kmdalal:
On my older Ubuntu machine, there were fewer missed message warnings - go figure!

Increasing `cell_side_length` to 0.010 was disastrous:
- About the same number of dropped message warnings, until physics seemed to blow up and have the lander buried in the terrain.
- Many instances of two new warnings during grind and scoop:
    - [Wrn] [RegolithPlugin.cpp:271] RegolithPlugin: Regolith particle spawn skipped to avoid particle overlap.
    - [Wrn] [RegolithPlugin.cpp:306] RegolithPlugin: More volume is being displaced per frame than can be spawned as regolith. Consider increasing the value of spawn_volume_threshold in common.launch to avoid this warning.

kmdalal:
Changing `cell_side_length` to 0.08 was also disastrous, in a very different way.  Here's the entire trace after unstow until the sim hung.
```
[INFO]: ArmUnstow action complete
[INFO]: TaskGrind action started
[ERROR]: grinder/grinder: Unable to sample any valid states for goal tree
[ WARN]: Timed out
[ WARN]: Fail: ABORTED: TIMED_OUT
[ERROR]: TaskGrind: Aborted - MoveIt planning failed with error code: val: -6
[INFO]: TaskGrind action complete
[INFO]: TaskGrind action started
[ERROR]: grinder/grinder: Unable to sample any valid states for goal tree
[ WARN]: Timed out
[ WARN]: Fail: ABORTED: TIMED_OUT
[ERROR]: TaskGrind: Aborted - MoveIt planning failed with error code: val: -6
[INFO]: TaskScoopCircular action started
[INFO]: TaskGrind action complete
[Msg] GlobalShaderParamPlugin::cacheParams(sunIntensity, 24)
[Msg] GlobalShaderParamPlugin::cacheParams(sunVisibility, 20)
[ERROR]: TaskScoopCircular: Aborted - The follow_joint_trajectory action server for arm_controller resulted in error code -5 with the message: j_shou_pitch goal error -0.431228
[INFO]: TaskScoopCircular action complete
[INFO]: TaskDiscardSample action started
[INFO]: TaskDiscardSample: Succeeded - TaskDiscardSample trajectory succeeded
[INFO]: TaskDiscardSample action complete
[INFO]: TaskScoopCircular action started
[ERROR]: TaskScoopCircular: Aborted - The follow_joint_trajectory action server for arm_controller resulted in error code -5 with the message: j_shou_pitch goal error -0.448331
[INFO]: TaskScoopCircular action complete
[INFO]: TaskDiscardSample action started
[INFO]: TaskDiscardSample: Succeeded - TaskDiscardSample trajectory succeeded
[INFO]: TaskDiscardSample action complete
[INFO]: TaskGrind action started
[ERROR]: grinder/grinder: Unable to sample any valid states for goal tree
[ WARN]: Timed out
[ WARN]: Fail: ABORTED: TIMED_OUT
[ERROR]: TaskGrind: Aborted - MoveIt planning failed with error code: val: -6
[INFO]: TaskGrind action complete
```

AstroStucky:
> On my older Ubuntu machine, there were fewer missed message warnings - go figure!
> 
> Increasing `cell_side_length` to 0.010 was disastrous:
> 
> * About the same number of dropped message warnings, until physics seemed to blow up and have the lander buried in the terrain.
> * Many instances of two new warnings during grind and scoop:
>   
>   * [Wrn] [RegolithPlugin.cpp:271] RegolithPlugin: Regolith particle spawn skipped to avoid particle overlap.
>   * [Wrn] [RegolithPlugin.cpp:306] RegolithPlugin: More volume is being displaced per frame than can be spawned as regolith. Consider increasing the value of spawn_volume_threshold in common.launch to avoid this warning.

0.01 is massive, and definitley more than you need to improve performance. 0.0075 or 0.008 is a better place to start. I have an okay idea as to why this happened, but I don't think it's fixable in this PR. 

Basically when the the edge of the scoop's geometry overlaps with a voxel I did not want the overlap algorithm to futz with calculating the exact overlap of the voxel region, so it instead claims the entire volume of that voxel. Its a better approximation the smaller the voxels are. This means that while larger voxels get you a performance boost, it also results in more volume being detected during digs by ow_regolith, and way more regolith being spawned. 

This is great data! It shows we need a better way to optimize the simulation for slower machines, whether that's the ability to turn off multi-material entirely but maintain regolith spawning or a fix to the above issue I explained. I think there is a solution that addresses both at the same time, but I need more time to think about it. 

If you want to retry your tests with the values I suggest 0.008 or 0.0075, I think things will go better. You could also counteract the effect by increasing the `spawn_volume_treshold` on line 973 of ow_lander/urdf/lander.xacro, but it will require some tweaking.

AstroStucky:
> Changing `cell_side_length` to 0.08 was also disastrous, in a very different way. Here's the entire trace after unstow until the sim hung.
> 
> ```
> [INFO]: ArmUnstow action complete
> [INFO]: TaskGrind action started
> [ERROR]: grinder/grinder: Unable to sample any valid states for goal tree
> [ WARN]: Timed out
> [ WARN]: Fail: ABORTED: TIMED_OUT
> [ERROR]: TaskGrind: Aborted - MoveIt planning failed with error code: val: -6
> [INFO]: TaskGrind action complete
> [INFO]: TaskGrind action started
> [ERROR]: grinder/grinder: Unable to sample any valid states for goal tree
> [ WARN]: Timed out
> [ WARN]: Fail: ABORTED: TIMED_OUT
> [ERROR]: TaskGrind: Aborted - MoveIt planning failed with error code: val: -6
> [INFO]: TaskScoopCircular action started
> [INFO]: TaskGrind action complete
> [Msg] GlobalShaderParamPlugin::cacheParams(sunIntensity, 24)
> [Msg] GlobalShaderParamPlugin::cacheParams(sunVisibility, 20)
> [ERROR]: TaskScoopCircular: Aborted - The follow_joint_trajectory action server for arm_controller resulted in error code -5 with the message: j_shou_pitch goal error -0.431228
> [INFO]: TaskScoopCircular action complete
> [INFO]: TaskDiscardSample action started
> [INFO]: TaskDiscardSample: Succeeded - TaskDiscardSample trajectory succeeded
> [INFO]: TaskDiscardSample action complete
> [INFO]: TaskScoopCircular action started
> [ERROR]: TaskScoopCircular: Aborted - The follow_joint_trajectory action server for arm_controller resulted in error code -5 with the message: j_shou_pitch goal error -0.448331
> [INFO]: TaskScoopCircular action complete
> [INFO]: TaskDiscardSample action started
> [INFO]: TaskDiscardSample: Succeeded - TaskDiscardSample trajectory succeeded
> [INFO]: TaskDiscardSample action complete
> [INFO]: TaskGrind action started
> [ERROR]: grinder/grinder: Unable to sample any valid states for goal tree
> [ WARN]: Timed out
> [ WARN]: Fail: ABORTED: TIMED_OUT
> [ERROR]: TaskGrind: Aborted - MoveIt planning failed with error code: val: -6
> [INFO]: TaskGrind action complete
> ```

These look like trajectory generation errors from the lander arm action servers. I don't see why that value would cause these. I wonder if something else was going on here?

AstroStucky:
I just tried a TaskGrind and TaskScoopLinear (default values) with `cell_side_length` of 0.01. The grid winds up chunkier, but doesn't look terrible. 
Screenshot_20240411_172218
Scooping resulted in a quite a few more particles, as I predicted, but the physics did not blow up. 

I'm worried I cannot effectively test these issues on my machine.

AstroStucky:
Okay! Things are agreeing better now. I got behavior much more similar to what you described by running ReferenceMission1 with `cell_side_length` of 0.01 m. I am truly surprised by the difference between the first dig being at 2cm depth versus 3cm depth.

The problem really comes from the significantly larger volume being detected by the MaterialDistributionPlugin. I'm going to now play with decreasing the`spawn_volume_threshold` to counteract the effect, so that if nothing else we have suggested values for underpowered machines that we can document in the user guide.

kmdalal:
Yes, I've only been testing this with ReferenceMission1 (and occasionally 2 and EuropaMission, but they have approx or exactly the same dig sequence).

I retried (on my older laptop) with `cell_side_length` at 0.075 and got much better results: normal execution with only two dropped message warnings (one in unstow, one in grind).  Should perhaps this be the default value?

kmdalal (approved):
I'll leave it to you if you want to slightly change the default `cell_side_length` in this PR.  I don't need to do any further testing.  My code review was light because of unfamiliarity with these packages, but if there's a part of it you want feedback on, you can go over it with me, and/or see if Terry is willing to have a look.

I don't have the impression that this is a really reliable change because of all the random errors that occurred during testing.  I'm also concerned about the performance hit for users with weaker machines.  I would suggest a future "hotfix" (though it's not really that) that can disable multicomposite terrain on startup.   It'll be your call here.

In any case this is an important and huge new feature that will enable more useful simulation and research.   Thanks for your persistence in tackling the tough issues.  Awesome work!

AstroStucky:
I have now made a change where the volume of a particle is no longer dependent on how many voxels the MaterialIntegrator counts within a given modification, but instead based on the changes of height of the differential image it receives. This is important because the differential image will always be either finer or equal resolution than the material grid. This allows a user to increase the `cell_side_length` and not see more regolith spawn as a result, which means the only impact of increasing this value is that the material blends may be less accurate as compared to the colors in the terrain. 

I did a tests with ReferenceMission1 and 2, and everything looks great! I did decide to increase the default `cell_side_length` to 0.0075 as you suggested. 

Making this fix has given me some insight into how I could allow the user to disable multi-material terrain entirely, but still maintain regolith spawning, which will also re-enable regolith spawning in europa_terminator_workspace.launch. I think this would be a straight forward enough hotfix.

PR #384, opened 2024-04-05
Oceanwater 700, 701 - activate comms action and camera power

AstroStucky:
## Linked Issues:
| EPIC ⚡ | [OCEANWATER-697](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-697) |
| Jira Ticket 🎟️ | [OCEANWATER-700](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-700) |
| Jira Ticket 🎟️ | [OCEANWATER-701](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-701) |

## Summary of Changes
* Fixed light power consumption not being distributed across power nodes. 
* ActivateComms actions created. It will simply draw 28 W (settable in system.cfg) for 10 seconds.
* Power system node now listens to the /electrical_power topic and applies the power load it transmits evenly across all power cells. 
* Renamed `efficiency` to `motor_efficiency`.
* Camera controller, an always on device, now consumes power similar to lander lights and mechanical power. It is settable in `system.cfg` as `power_camera_controller`.
* Each camera being exposed draws a power of 2.0 W (see Caveats).

## Understanding the Wattage Values. 
Any power consumption applied to the battery, such as a single light at 100% intensity that consumes 0.3 W is divided by the number of still running nodes (4 if the `disconnect_battery_nodes` has not bee ticked). This means that a single light shows up in a cell's input power as 0.3 / 4 = 0.075 W if there are 4 power nodes operating. If one power node has failed the input power is 0.3 / 3  = 0.1 W.

## Test 1 - Activate Comms and Lights Power Fix
1. In `system.cfg`: increase`active_models` to at least 2, and change both `print_debug` and `inputs_print_debug` to true. 
2. Start any simulation world. 
3. The following will print
```
[ INFO:/power_system_node] M00 INPUT power: 7.500000.  volts: 17.675267.  tmp: 18.132230
[ INFO:/power_system_node] M01 INPUT power: 7.500000.  volts: 17.675267.  tmp: 18.132230
```
4. Tilt lander mast down so light project onto lander body `rosrun ow_lander tilt.py 1.2`
5. Shutdown one of the lights `rosrun ow_lander light_set_intensity.py left 0.0`
6. Verify output is now 
```
[ INFO:/power_system_node] M00 INPUT power: 6.300000.  volts: 17.662830.  tmp: 21.319773
[ INFO:/power_system_node] M01 INPUT power: 6.300000.  volts: 17.662830.  tmp: 21.319773
```
7. Shutdown the other light `rosrun ow_lander light_set_intensity.py right 0.0`
8. Verify the output
```
[ INFO:/power_system_node] M00 INPUT power: 5.100000.  volts: 17.644684.  tmp: 20.618757
[ INFO:/power_system_node] M01 INPUT power: 5.100000.  volts: 17.644684.  tmp: 20.618757
```
9. Turn on a single light `rosrun ow_lander light_set_intensity.py right 1.0` 
10. Increase the number of disconnected power cells by 1 by setting the value of `battery_nodes_to_disconnect` to 1, and ticking `disconnect_battery_nodes`. 
11. The output should now be
```
[ INFO:/power_system_node] M00 INPUT power: 6.833333.  volts: 17.657024.  tmp: 21.135746
[ INFO:/power_system_node] M01 INPUT power: 6.833333.  volts: 17.657024.  tmp: 21.135746
```
12. Power down the light `rosrun ow_lander light_set_intensity.py right 0.0` so that output is (the value is different because other power base lines are also being distributed across only 3 nodes now).
```
[ INFO:/power_system_node] M00 INPUT power: 5.233333.  volts: 17.627244.  tmp: 18.482299
[ INFO:/power_system_node] M01 INPUT power: 5.233333.  volts: 17.627244.  tmp: 18.482299
```
13. Run the ActivateComms action with default arguments to see that it draws 9.6 for 10 seconds (in simulator time). `rosrun ow_lander activate_comms.py`
15. The output while the action runs should be 
```
[ INFO:/power_system_node] M00 INPUT power: 8.433333.  volts: 17.609452.  tmp: 19.775166
[ INFO:/power_system_node] M01 INPUT power: 8.433333.  volts: 17.609452.  tmp: 19.775166
```
16. Run ActivateComms again with a different duration `rosrun ow_lander activate_comms.py --duration 5`. The output will be the same as above, but for only 5 seconds this time.

## Test 2 - Camera Power
1. In system.cfg: change both print_debug and inputs_print_debug to true. You can leave `active_models` at 2 or change it back to 1.
2. Start any simulation world.
3. You should see the output 
```
[ INFO:/power_system_node] M00 INPUT power: 7.500000.  volts: 17.650162.  tmp: 21.013723
```
* You will see more lines if `active_models` is still more than 1.  
4. Close the simulation
5. Change the value of `power_baseline_camera_controller` in system.cfg to `8.0`
6. Run the simulation, and you should see the following output
```
[ INFO:/power_system_node] M00 INPUT power: 9.100000.  volts: 17.667873.  tmp: 18.347697
```
Indicating that the extra camera controller power load of 6.4 has been distributed across all 4 power cells. 

## Caveat
The power draw from cameras being exposed has been implemented and is settable as `power_active_camera` in system.cfg; however, it appears our camera model does not actually take any simulation time to expose, no matter how high you set the exposure time. Meaning the value does not work as expected. The implementation is left for in case this camera exposure bug is addressed in the future. See [OW-1239](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1239)


kmdalal:
Please update the "Understanding the Wattage values" of the Description now that 'efficiency' is understood.

ow_lander/src/ow_lander/actions.py
```diff
@@ -1042,6 +1043,14 @@ def execute_action(self, goal):
 
     self._pub_trigger.publish()
 
+    # apply power usage for 2 cameras
```
kmdalal, line 1046: I think we should discuss, with @chetankul, whether our stereo camera should be treated as one entity or two.  I'm not sure which his power estimate is based on.  The two cameras are always triggered together, regardless of whether a left or right image capture is commanded.

AstroStucky: Sure. Bare in mind that at present this value makes a very minor difference in any case due to [OW-1239](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1239). Also I actually based how camera power works off of comments made by Larry Edwards that were quoted in the [Jira ticket](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-701).

ow_power_system/config/system.cfg
```diff
@@ -28,16 +28,21 @@ voltage_range: 0.1			        # volts
...
 baseline_power_computing: 0.1
 baseline_power_heating: 1.2
 baseline_power_science_instr: 0.6
 baseline_power_sample_handling: 0.6
 baseline_power_other: 2.2
 
+
 # Power coefficients for dynamically computed power draw.
-wattage_coefficient_lights: 0.3    # power draw (W) per light at max intensity
+power_lights_coefficient: 0.3       # power draw (W) per light at max intensity
```
kmdalal, line 39: Although I introduced the `coefficient` suffix, I'm not sure I like it any more.  Would `power_lights` be better?

ow_power_system/config/system.cfg
```diff
@@ -28,16 +28,21 @@ voltage_range: 0.1			        # volts
...
+power_camera_controller: 1.0        # power draw of the camera controller
```
kmdalal, line 40: Isn't this a baseline draw, hence better named `baseline_power_camera_controller` ?

ow_power_system/config/system.cfg
```diff
@@ -28,16 +28,21 @@ voltage_range: 0.1			        # volts
...
+power_camera_exposed_coefficient: 2.0  # power draw while a camera is exposing
```
kmdalal, line 41: I suggest `camera_active` .  "exposed" sounds a bit odd to me.

ow_power_system/config/system.cfg
```diff
@@ -28,16 +28,21 @@ voltage_range: 0.1			        # volts
...
+power_comms_active: 28.0           # power draw when comms is activated
+
+# length of time the antenna draws power during an ActivateComms operation
+communication_duration: 10.0       # seconds
```
kmdalal, line 45: If the value here takes that much real time, I suggest making the default smaller, like 2-5 seconds, to lessen overall simulation time.  It would be good to find out what the average comms time is, though not hold up the PR over this.

ow_power_system/config/system.cfg
AstroStucky: I think I'd like to prefix any power draw related value with `power_`, so how about we do `power_baseline_camera_controller`?

Addressing your other comment, what if we made `power_lights_coefficient` into `power_active_lights`?

ow_power_system/config/system.cfg
AstroStucky: How about `power_active_camera`?

ow_power_system/config/system.cfg
```diff
@@ -28,16 +28,21 @@ voltage_range: 0.1			        # volts
...
 baseline_power_sample_handling: 0.6
 baseline_power_other: 2.2
 
+
 # Power coefficients for dynamically computed power draw.
-wattage_coefficient_lights: 0.3    # power draw (W) per light at max intensity
+power_lights_coefficient: 0.3       # power draw (W) per light at max intensity
+power_camera_controller: 1.0        # power draw of the camera controller
+power_camera_exposed_coefficient: 2.0  # power draw while a camera is exposing
+power_comms_active: 28.0           # power draw when comms is activated
```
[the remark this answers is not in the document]
kmdalal: Let's get @chetankul 's opinion, since the values he's been proposing seem to be of a different scale.  Also, note the other values are per-cell based on a specific battery pack.  Personally, I think the total power draw of each component would be more intuitive.  Let's get @chetankul 's opinion on this, too.

ow_power_system/config/system.cfg
kmdalal: Since we don't have a good intuition or any data on this, but I think comms times could vary a lot, I'd like to request that the duration (in seconds)  be made a parameter (goal) of the action.  I'm assuming this is easy, and would make the action a lot more flexible.

ow_lander/src/ow_lander/actions.py
```diff
@@ -1042,6 +1043,14 @@ def execute_action(self, goal):
...
+    ### FIXME: Exposure does not appear to be simulated in time (e.g. a 10
+    ###   second exposure will not take 10 seconds to complete), so this model
+    ###   per camera power consumption does not actually work the way it should.
```
kmdalal, line 1049: Were you intending to fix this for this PR, or as future work?

ow_lander/src/ow_lander/actions.py
```diff
@@ -1367,3 +1377,33 @@ def execute_action(self, goal):
...
+
+  def execute_action(self, goal):
+
+    # This action only draws a large amount of power from the battery for some
+    # period of time. There is no simulation of mission to Earth comms and
+    # uplinking and downlinking, for the purpose of this action, are treated the
+    # same and draw the same amount of power.
+
+    power_load_while_active = rospy.get_param('/ow_power_system/comms_active')
+    duration_of_activation = rospy.get_param('/ow_power_system/comms_duration')
```
kmdalal, line 1402: This may be a dumb question, but are these values both ROS parameters and read from a config file?  Perhaps I just haven't yet seen the code that connects them.

kmdalal, line 1402: In case you saw it, I'm replacing my previous (dumb) question with this one, which might also be dumb.  How do you know that these calls will happen _after_ the param is set?

ow_power_system/include/PowerSystemNode.h
```diff
@@ -182,10 +191,12 @@ class PowerSystemNode
...
   bool m_thermal_failure_activated = false;
 
+  double m_mean_mechanical_power = 0.0;
+  double m_electrical_power = 0.0;
+
   // Used for mechanical power debug printouts.
   double m_power_watts;
-  double m_mean_mechanical_power;
-  double m_sum_mechanical_power = 0;
+  double m_sum_mechanical_power = 0.0;
```
kmdalal, line 199: Although this existed before, it should not be a member variable, but rather a local variable in jointStatesCb().

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -264,15 +283,19 @@ void PowerSystemNode::initAndRun()
...
     // Save the current average mechanical power, in case jointStatesCb runs
     // while processing which would update m_mean_mechanical_power mid-loop.
-    double mech_power = m_mean_mechanical_power;
+    const double actual_mech_power
+      = m_mean_mechanical_power / m_mechanical_efficiency;
+    const double shared_power_load
+      = (
+          actual_mech_power + m_electrical_power + m_power_camera_controller
+          + electricalPowerLights()
+        ) / (NUM_MODELS - m_deactivated_models);
```
kmdalal, line 292: I haven't seen where the baseline power values are included - am I missing something?

ow_lander/src/ow_lander/actions.py
AstroStucky: Future work. I would not even know where to start having never worked on the stereo cam.

ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -264,15 +283,19 @@ void PowerSystemNode::initAndRun()
     // corresponding prognoser via its message bus.
     // Save the current average mechanical power, in case jointStatesCb runs
     // while processing which would update m_mean_mechanical_power mid-loop.
-    double mech_power = m_mean_mechanical_power;
+    const double actual_mech_power
+      = m_mean_mechanical_power / m_mechanical_efficiency;
```
kmdalal, line 287: Outside the scope of this PR, but I'd like to know if `m_mean_mechanical_power` is properly computed from joint states.

ow_lander/src/ow_lander/actions.py
AstroStucky: `execute_action` is not called until the action is called. The user could even adjust the parameter after simulation start-up, and that adjustment would be picked up the next time they call this action.

ow_power_system/include/PowerSystemNode.h
AstroStucky: This needs to be a member variable. In the `jointStatesCb()` function it's being operated on by a `+=`.

ow_power_system/src/PowerSystemNode.cpp
AstroStucky: Those baseline powers are only present in the PrognoserInputHandler, which is exactly the problem I was pointing out. They aren't being distributed across all batteries. 

My understanding was that I would address that in a PR following this PR.

kmdalal: Got it.  That sounds fine -- it can be a hotfix.

ow_power_system/include/PowerSystemNode.h
kmdalal: That was my point.  I see it used only in that function, so it can be local to that function.

AstroStucky: It's irrelevant whether its only used in a single function, it carries state between function calls. It accumulates a value each time. Making it local would break things.

ow_power_system/config/system.cfg
kmdalal: Yes, as long as the naming is consistent.

ow_power_system/src/PowerSystemNode.cpp
AstroStucky: This was an optimization I suggested a ways back. I believe Liam's implementation is correct. 
https://github.com/nasa/ow_simulator/pull/352#discussion_r1289233696

ow_power_system/config/system.cfg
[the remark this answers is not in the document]
AstroStucky: Will treat the stereocam as a single system.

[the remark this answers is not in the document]
AstroStucky: Will change CCU consumed value to 1.6W

ow_power_system/config/system.cfg
AstroStucky: Will make communication duration a parameter in the ActivateComms action

ow_lander/src/ow_lander/actions.py
AstroStucky: Will make this a single light system

ow_power_system/config/system.cfg
AstroStucky: Done

ow_power_system/config/system.cfg
AstroStucky: Done

ow_lander/src/ow_lander/actions.py
AstroStucky: Done

ow_power_system/config/system.cfg
AstroStucky: Now changed to `power_active_lights`

ow_power_system/config/system.cfg
AstroStucky: done

ow_power_system/config/system.cfg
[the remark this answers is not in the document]
AstroStucky: done

ow_power_system/config/system.cfg
```diff
@@ -27,17 +27,20 @@ voltage_range: 0.1			        # volts
...
 baseline_power_computing: 0.1
 baseline_power_heating: 1.2
 baseline_power_science_instr: 0.6
 baseline_power_sample_handling: 0.6
 baseline_power_other: 2.2
 
-# Power coefficients for dynamically computed power draw.
-wattage_coefficient_lights: 0.3    # power draw (W) per light at max intensity
+# Baseline power draws (Watts)
+power_baseline_camera_controller: 1.6  # power draw of the camera controller
```
kmdalal, line 38: I like the new naming convention.  Small request: change the other `baseline_power*` entries to match this one.

AstroStucky: > I like the new naming convention. Small request: change the other `baseline_power*` entries to match this one.

Can we do that in the hotfix that will make those baselines distribute across all batteries? I like that they're differentiated right now, since they currently behave differently.

kmdalal:
The tests perform exactly as given in the description, but there are two things that confused me:
1. I always started with `active_models` at 2, yet the description says power is distributed across "all 4 power cells" (I think you meant 'model' instead of cell).
2. I am able to disconnect up to 3 nodes, which gives: `[ INFO]: M00 INPUT power: 22.300000.  volts: 17.619061.  tmp: 19.923713`.  How can I disconnect 3 nodes when I started with 2?  If this is a bug, it seems outside this PR.

kmdalal (approved):
Test passed.  I deleted a comment I just wrote when I realized my own confusion.  Nice new features!   

Note that the corresponding ow_autonomy PR is not yet ready for review -- I need to add the comms duration parameter.

PR #388, opened 2024-04-17
Hotfix for camera goal error

kmdalal:
### Summary

Recently, fault detection was modified to flag a goal error if there is no point cloud received after the camera is triggered.  The problem here is that the check was done *just* after the trigger.  Some point clouds may need more time to arrive.

This "fix" also allowed goal errors to get flagged when the camera was triggered without the `CameraCapture` action, i.e. via a message.  However this is invalid, because goal errors should only be associated with actions.  I.e. actions have goals, pure messages do not.

This hotfix is simply to remove this invalid "fix".

## Test
* Try ReferenceMission1 without this hotfix and see most or every CameraCapture fail due to spurious goal errors.
* Apply the hotfix, and ReferenceMission1 should be free of the above problem.
* Inject the camera fault and run `camera_capture.py`.  You should experience a goal error.

### DIFF
--- a/ow_faults_detection/src/FaultDetector.cpp
+++ b/ow_faults_detection/src/FaultDetector.cpp
@@ -237,7 +237,6 @@ void FaultDetector::cameraTriggerCb(const std_msgs::Empty& msg)
   if (m_camera_data_pending) {
     m_camera_faults_flags |= CameraFaultsStatus::NO_IMAGE;
     m_system_faults_flags |= SystemFaultsStatus::CAMERA_EXECUTION_ERROR;
-    m_system_faults_flags |= SystemFaultsStatus::CAMERA_GOAL_ERROR;
 
     // publish updated faults messages (exonerated faults published in cameraRawCb)
     publishFaultsMessage(m_camera_faults_msg_pub, CameraFaultsStatus(), m_camera_faults_flags);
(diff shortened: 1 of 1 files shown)


AstroStucky:
> * Try ReferenceMission1 without this hotfix and see most or every CameraCapture fail due to spurious goal errors.

I was unable to replicate this on noetic-devel.

kmdalal:
> > * Try ReferenceMission1 without this hotfix and see most or every CameraCapture fail due to spurious goal errors.
> 
> I was unable to replicate this on noetic-devel.

And I'm pretty sure I didn't always see the problem either.  I suspect this a machine performance/load issue, and if the point cloud arrives fast enough, the error won't happen.  Please try the hotfix and see if all works.  Applying it corrected the problem for me in EuropaMission as well.

AstroStucky:
I agree. This probably has to do with machine load and timing, making it's occurrence sporadic. 

I think your assessment that goal errors should only be associated with actions is correct. It makes sense and appears to be how we originally programmed them in. Even if I cannot replicate the original problem, I agree from the code that there is a problem that goes against our design and this PR has fixed it.  

In the documentation I think we should discourage the use of the image_trigger topic to capture images. Otherwise it becomes difficult to guarantee consistent behavior.

AstroStucky (approved):

kmdalal:
> In the documentation I think we should discourage the use of the image_trigger topic to capture images. Otherwise it becomes difficult to guarantee consistent behavior.

Thanks for the quick review.  With regard to the comment above, the message trigger seems okay for experimental or interactive use.  But I agree it shouldn't be used in general by autonomy and that the CameraCapture action is the right approach.

PR #390, opened 2024-04-18
Planning timeout increase for TaskDiscardSample

AstroStucky:
## Sibling PR
https://github.com/nasa/ow_autonomy/pull/149

## Summary of Changes
* Added a planning time override for TaskDiscardSample only that restores the previous planning time values after completion of the action. 
* Added an `Iteration` termination condition for all RRTStar planned actions. This means the action will either reach the timeout or the iteration limit. 
* Fixed a bug where GuardedMove was not switching the planner back to RRTConnect after concluding. 

## Test
1. Have this ow_simulator on planning-timeout-test and ow_autonomy on remove-tailings-default-discard-arguments. Everything else on noetic-devel. 
2. Run ReferenceMission1.plx. several times. This passes so long as there is not a single plan trajectory failure.

### DIFF
--- a/ow_lander/src/ow_lander/actions.py
+++ b/ow_lander/src/ow_lander/actions.py
@@ -542,6 +542,8 @@ class TaskDiscardSampleServer(mixins.FrameMixin, mixins.ArmTrajectoryMixin,
     discard_surface_pos = self.transform_to_planning_frame(
       self.get_intended_position(goal.frame, goal.relative, goal.point)).point
     self._arm.move_group_scoop.set_planner_id("RRTstar")
+    PLANNING_TIMEOUT = 0.1
+    self._arm.move_group_scoop.set_planning_time(PLANNING_TIMEOUT)
     try:
       sequence = TrajectorySequence(
         self._arm.robot, self._arm.move_group_scoop, 'l_scoop')
--- a/ow_lander/src/ow_lander/actions.py
+++ b/ow_lander/src/ow_lander/actions.py
@@ -225,7 +225,7 @@ class GuardedMoveServer(mixins.ArmActionMixin, ActionServerBase):
         self._pub_result.publish(False, '', Point())
         self._set_succeeded("No ground detected", final=Point(), success=False)
     finally:
-      self._arm.move_group_scoop.set_planner_id('RRTstar')
+      self._arm.move_group_scoop.set_planner_id('RRTConnect')
 
 
 class ArmUnstowServer(mixins.ArmTrajectoryMixin, ActionServerBase):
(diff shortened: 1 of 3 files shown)


kmdalal:
In my first attempt of RM1, the first Discard was fine, but the ScoopCircular that followed failed in a bizarre way.  The arm stopped in the position shown below and just kept bouncing every few minutes.
scoop

AstroStucky:
That's a surprise. Have you been able to reproduce that?

kmdalal:
My second run of RM1 completed successfully.  However, the second Discard took a bizarre path, ending with the scoop touching the ground and not dumping its contents.  I tried to take a screenshot but wasn't fast enough.

AstroStucky:
I actually just encountered a path planning failure of TaskDiscardSample myself, so I've decided to increase the iteration limit to 3000. This means it may very well take the full planning timeout to plan on both our computers.

I have also made the TaskDiscardSample values in RemoveTailings the defaults, which make it discard from higher up, reducing the likelihood is picks a  patch that collides with terrain. This PR now has a sibling https://github.com/nasa/ow_autonomy/pull/149. Description has been editted.

ow_lander/src/ow_lander/actions.py
```diff
@@ -225,7 +225,7 @@ def ground_detect_cb():
         self._pub_result.publish(False, '', Point())
         self._set_succeeded("No ground detected", final=Point(), success=False)
     finally:
-      self._arm.move_group_scoop.set_planner_id('RRTstar')
+      self._arm.move_group_scoop.set_planner_id('RRTConnect')
```
kmdalal, line 228: I don't know the implications of changing the planner.  Is this safe to do without more extensive testing?

kmdalal:
> I actually just encountered a path planning failure of TaskDiscardSample myself, so I've decided to increase the iteration limit to 3000

This update is not in my recent pull - do you intend to push it?   In the meantime I had two consecutive flawless runs of RM1.

ow_lander/src/ow_lander/actions.py
AstroStucky: I'm pretty confident it's safe. It should've always switched back to RRTConnect, the same as the other RRTStar using actions do. Things don't break in an obvious way without this change, you just suddenly have all of your arm trajectories being planned with RRTstar instead of RRTConnect. It's a silent, but fundamental change to arm operation that would have been very easy to trip during a simulation.

AstroStucky:
> > I actually just encountered a path planning failure of TaskDiscardSample myself, so I've decided to increase the iteration limit to 3000
> 
> This update is not in my recent pull - do you intend to push it? In the meantime I had two consecutive flawless runs of RM1.

Whoops! You are correct I forgot to push.

ow_lander/src/ow_lander/actions.py
kmdalal: Anjan had experimented a lot with both these planners and had strong reasons for one over the other.  I forget the details, but for a while we switched to one and saw much faster planning but other problems including poor trajectories, and then he switched some actions back to the original planner.  The history and rationale should be found in closed PRs and Jiras.

kmdalal (approved):
I've had two consecutive flawless runs of RM1 with the latest push and I think we can call this good to go!

PR #391, opened 2024-04-19
HOTFIX All baseline power distributed across all 4 power cells

AstroStucky:
## Linked Issues:
 Jira Ticket 🎟️ | [OCEANWATER-1245](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1245)

## Summary of Changes
* All baseline power is now distributed evenly across as power nodes connected in parallel. 

## Test
1. Go into `ow_power_system/config/system.cfg` and change `print_debug` to **true** and `inputs_print_debug` to **true**.
2. Start the simulation in any world. You will see the following line continually print. 
```
[ INFO:/power_system_node] M00 INPUT power: 3.975000.  volts: 17.611784.  tmp: 18.298205
```
3. In RQT increase `battery_nodes_to_disconnect` to 1 and tick `disconnect_battery_nodes`. The message should now equal the same power, but distributed across only 3 nodes instead of 4. We can compute what this value should be with 3.975 * 4 / 3 = 5.3
```
[ INFO:/power_system_node] disconnect_battery_nodes activated!
[ INFO:/power_system_node] disconnect_battery_nodes updated! 1 out of 4 models are now disconnected.
[ INFO:/power_system_node] M00 INPUT power: 5.300000.  volts: 17.656972.  tmp: 19.390101
```
4. Increase `battery_nodes_to_disconnect` to 2 for a new value of 3.975 * 4 / 2 = 7.95
```
[ INFO:/power_system_node] disconnect_battery_nodes updated! 2 out of 4 models are now disconnected.
[ INFO:/power_system_node] M00 INPUT power: 7.950000.  volts: 17.692063.  tmp: 18.740624
```
5. Increase `battery_nodes_to_disconnect` to 3 for a new value of 3.975 * 4 / 1 = 15.9
```
[ INFO:/power_system_node] disconnect_battery_nodes updated! 3 out of 4 models are now disconnected.
[ INFO:/power_system_node] M00 INPUT power: 15.900000.  volts: 17.624586.  tmp: 20.168456
```

### DIFF
--- a/ow_power_system/src/PowerSystemNode.cpp
+++ b/ow_power_system/src/PowerSystemNode.cpp
@@ -93,8 +93,14 @@ void PowerSystemNode::initAndRun()
     m_spinner_threads = system_config.getInt32("spinner_threads");
     m_mechanical_efficiency = system_config.getDouble("motor_efficiency");
     m_power_active_lights = system_config.getDouble("power_active_lights");
-    m_power_baseline_camera_controller
-      = system_config.getDouble("power_baseline_camera_controller");
+    m_power_baseline = (
+        system_config.getDouble("power_baseline_camera_controller")
+      + system_config.getDouble("power_baseline_computing")
+      + system_config.getDouble("power_baseline_heating")
+      + system_config.getDouble("power_baseline_science_instr")
+      + system_config.getDouble("power_baseline_sample_handling")
+      + system_config.getDouble("power_baseline_other")
+    );
     // use ROS Param to make other nodes, such as the lander action servers,
     // aware of power parameters
     m_nh.setParam(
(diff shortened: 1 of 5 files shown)


ow_power_system/src/PowerSystemNode.cpp
```diff
@@ -284,7 +290,7 @@ void PowerSystemNode::initAndRun()
       = (
           actual_mech_power
           + m_electrical_power
-          + m_power_baseline_camera_controller
+          + m_power_baseline
           + electricalPowerLights()
         ) / (NUM_MODELS - m_deactivated_models);
```
kmdalal, line 295: Outside the scope of this PR, but I find the double use of the term "model" quite confusing.  It refers to a GSAP instance in some places, and a battery node in others (with battery node meaning a bank of cells, 6 in this case).  @chetankul do you have thoughts on this?

kmdalal: There's also the power system ROS node :)

kmdalal (approved):
Test passes, and code change is clear.  Good catch of a long-standing problem!

PR #392, opened 2024-04-23
Updated power consumption values

kmdalal:
### Summary

Updated power consumption values in configuration file to be total power draw of each subsystem, and revised comments accordingly.

### Test

The Release 13 power system test, which was revised based on this hotfix and should be done using this hotfix branch.


AstroStucky (approved):
The power test worked great! Configuration changes look good.

PR #395, opened 2024-05-11
Fixed instantaneous capacity loss overriding low state of charge

AstroStucky:
## Summary of Changes
* Fixed a bug where instantaneous capacity loss and low state of charge being injected on the same battery cycle would cause the low state of charge fault to be ignored. 
* Cleaned up some commented out code.

## Test
1. Start the simulator in any world.
2. In a second terminal, monitor the state of charge topic `rostopic echo /battery_state_of_charge/value`
3. Bring up Dynamic Reconfigure in RQT and navigate to the fault menu.
4. In quick succession, tick `instantaneous_capacity_loss` then `low_state_of_charge`.
5. In the terminal printing SOC values, you will should see values immediately drop to about 0.01, then continue dropping by a significant amount each consecutive value (by about 5% of whatever the previous value was). 
```
0.9508093595504761
---
0.9508093595504761
---
0.09006000018358229
---
0.08537688141941069
---
0.08093728476621655
---
0.07672854707759652
---
0.07273866369058515
---
0.06895625418452515
---
0.06537052992047607
---
0.06197126326857315
---
0.058748758435563174
---
0.05569382380930803
```
6. Untick both faults in RQT and you should see the previous battery SOC be restored (about 0.95).

### DIFF
--- a/ow_power_system/src/PowerSystemNode.cpp
+++ b/ow_power_system/src/PowerSystemNode.cpp
@@ -21,6 +21,7 @@
 using namespace PCOE;
 
 // Fault names
+
 const std::string FAULT_NAME_HPD              = "high_power_draw";
 const std::string FAULT_NAME_HPD_ACTIVATE     = "activate_high_power_draw";
 const std::string FAULT_NAME_DBN              = "battery_nodes_to_disconnect";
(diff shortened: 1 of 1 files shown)


AstroStucky:
Performed test and it passes. Awaiting a GitHub policy change before merging.

kmdalal:
I think an additional regression test or two is needed.  I'll specify this in a day or two, and try them myself.

kmdalal (commented):
The given test passed.  I did several more regression tests:

- tried low SOC and ICL faults independently
- tried both in succession in both orders, slowly
- tried the other order, quickly

Worked perfectly.  Glad the fix was so simple!

kmdalal (approved):

kmdalal (approved):
It's great that a color filter change was all needed, and that the search rejects several locations before finding a good one.  Thanks for this improvement, well worth a patch to the release.

AstroStucky:
> It's great that a color filter change was all needed, and that the search rejects several locations before finding a good one. Thanks for this improvement, well worth a patch to the release.

We were lucky it was this simple!

AstroStucky:
> > It's great that a color filter change was all needed, and that the search rejects several locations before finding a good one. Thanks for this improvement, well worth a patch to the release.
> 
> We were lucky it was this simple!

I think we both got PR's mixed up. But no harm.

PR #399, opened 2024-05-29
Added a dynamic reconfigure option to auto clear goal errors

AstroStucky:
## Linked Issues:
Jira Ticket 🎟️ | [OCEANWATER-1248](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1248)

## Summary of Changes
* All world launch files have an option to automatically clear goal errors when an action in the goal error category completes successfully. 
  * This is now the default behavior, but the more complex behavior that requires the use of FaultClear can still be used by setting the flag to false at simulation launch or modifying during runtime via the ros parameter server. 
* Made a category in the faults section of Dynamic Reconfigure called "fault_behaviors" for both "arm_motion_continues_in_fault" and "auto_clear_goal_errors".

## Test Default (simple) Fault Behavior
1. Launch any simulation world 
2. Monitor faults `rostopic echo /system_fault_status/value`. It should start at 0.
3. Call an impossible action, e.g. `rosrun ow_lander arm_move_cartesian.py -x 100`. It will fail, and system_fault_status will be 2 or whatever the value of the action category you called. 
4.  Call a possible action in the same category, e.g. `rosrun ow_lander arm_move_cartesian.py -x 1.8 -z 0.5`. At the start of the action the system fault topic will go back to 0 and the action should complete.  
5. Go into RQT GUI, open Dynamic Reconfigure tab, go to faults, and under the "fault_behaviors" section untick "auto_clear_goal_errors". 
6. Call an impossible action, e.g. `rosrun ow_lander arm_move_cartesian.py -x 100`. It will fail, and system_fault_status will be 2 or whatever the value of the action category you called. 
7.  Call a possible action in the same category, e.g. `rosrun ow_lander arm_move_cartesian.py -x 1.8 -z 0.5`. The action  will be rejected. 
8. Clear the goal error `rosrun ow_lander fault_clear.py 2`, or use the value correct for the action category you used. The echo'd value should now be zero.
9.  Call a possible action in the same category, e.g. `rosrun ow_lander arm_move_cartesian.py -x 1.8 -z 0.5`. The action should  complete successfully. 

## Notes on FaultClear
I decided not to produce a warning/error when FaultClear is used and auto_clear_goal_errors is true because one could imagine a user's autonomy to be configured such that clearing goal errors is still necessary before issuing a follow-up lander action.

### DIFF
--- a/ow_lander/src/ow_lander/server.py
+++ b/ow_lander/src/ow_lander/server.py
@@ -122,6 +122,9 @@ class ActionServerBase(ABC):
     if self.fault_handler:
       self.fault_handler.notify_aborted()
 
+  def __should_auto_clear_goal_errors(self):
+    return rospy.get_param("/lander_action_servers/auto_clear_goal_errors", True)
+
   def __should_reject_due_to_fault(self):
     """Return true if the current state of faults and simulation settings permit
     an action to run.
@@ -129,7 +132,7 @@ class ActionServerBase(ABC):
     return (
       # Actions may proceed if the simulation is configured such that goal
       # errors are automatically cleared by a subsequent successful run.
-      not rospy.get_param("/lander_action_servers/auto_clear_goal_errors", True)
+      not self.__should_auto_clear_goal_errors()
       # Assuming this action has a fault_handler, do not permit the action to
       # proceed if it's handler is faulted.
       and (self.fault_handler and self.fault_handler.is_faulted())
@@ -155,6 +158,8 @@ class ActionServerBase(ABC):
     if self.__should_reject_due_to_fault():
       self.__set_rejected(self.fault_handler.get_rejected_message())
     else:
+      if self.fault_handler and self.__should_auto_clear_goal_errors():
+        self.fault_handler.reset_system_faults()
       self.execute_action(goal)
     rospy.loginfo(f"{self.name} action complete")
 
--- a/ow_lander/src/ow_lander/server.py
+++ b/ow_lander/src/ow_lander/server.py
@@ -123,7 +123,8 @@ class ActionServerBase(ABC):
       self.fault_handler.notify_aborted()
 
   def __should_auto_clear_goal_errors(self):
-    return rospy.get_param("/lander_action_servers/auto_clear_goal_errors", True)
+    """Return value of auto clear goal errors rosparam"""
+    return rospy.get_param("/faults/auto_clear_goal_errors")
 
   def __should_reject_due_to_fault(self):
     """Return true if the current state of faults and simulation settings permit
--- a/ow_lander/src/ow_lander/server.py
+++ b/ow_lander/src/ow_lander/server.py
@@ -122,6 +122,19 @@ class ActionServerBase(ABC):
     if self.fault_handler:
       self.fault_handler.notify_aborted()
 
+  def __should_reject_due_to_fault(self):
+    """Return true if the current state of faults and simulation settings permit
+    an action to run.
+    """
+    return (
+      # Actions may proceed if the simulation is configured such that goal
+      # errors are automatically cleared by a subsequent successful run.
+      not rospy.get_param("/lander_action_servers/auto_clear_goal_errors", True)
+      # Assuming this action has a fault_handler, do not permit the action to
+      # proceed if it's handler is faulted.
+      and (self.fault_handler and self.fault_handler.is_faulted())
+    )
+
   def __set_rejected(self, msg):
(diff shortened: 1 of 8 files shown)


ow_lander/src/ow_lander/server.py
```diff
@@ -122,6 +122,19 @@ def _set_aborted(self, msg, **kwargs):
...
       self.fault_handler.notify_aborted()
 
+  def __should_reject_due_to_fault(self):
+    """Return true if the current state of faults and simulation settings permit
+    an action to run.
+    """
+    return (
+      # Actions may proceed if the simulation is configured such that goal
+      # errors are automatically cleared by a subsequent successful run.
+      not rospy.get_param("/lander_action_servers/auto_clear_goal_errors", True)
```
kmdalal, line 132: Given this is a parameter, can one dynamically change its value using dynamic_reconfigure?

kmdalal:
Feedback on the new default (auto-clear goal errors): Although the behavior now matches what we had before, there is a pre-existing flaw: when a goal error is present, it is not cleared until the _completion_ of the next error-free action; ideally it should be cleared first.

Also, while I agree FaultClear should not print any warnings, it would be good to see a warning (or INFO message) whenever a goal error is automatically cleared.

ow_lander/src/ow_lander/server.py
AstroStucky: No. It does not show up in the dynamic reconfigure GUI. I had not considered doing it that way, but it could be. If it becomes a dynamic reconfigure value, I certain if it can also be settable as a launch file argument.

What would be more preferable: setting it from command line, or setting it from rqt? If I can do both I will.

AstroStucky:
> Feedback on the new default (auto-clear goal errors): Although the behavior now matches what we had before, there is a pre-existing flaw: when a goal error is present, it is not cleared until the _completion_ of the next error-free action; ideally it should be cleared first.

That is how it used to work, and the old behavior is what was specified in the ticket. I am worried a different behavior will take more time to implement than I have left on this project along with everything else I have to do, but I will at least investigate the possibility. 

> Also, while I agree FaultClear should not print any warnings, it would be good to see a warning (or INFO message) whenever a goal error is automatically cleared.

This already happens in simulator output. Keep in mind action clients scripts do not print anything.

ow_lander/src/ow_lander/server.py
kmdalal: Ideally both.  If that's hard, I'd prefer RQT, but maybe only if the default could still be true (auto-clear).  A command-line option alone is acceptable as is, it's just more cumbersome.

AstroStucky: Actually I got an easy solution that only took a couple lines of changes. It's a little hacky but it should do the trick. Update with it later tonight.

kmdalal: I'm done with the given tests (plus a few of my own) and code review, and would approve this PR as is.  But I'd like to try the dynamic reconfigure option, so I'll wait for that...

AstroStucky:
> > Feedback on the new default (auto-clear goal errors): Although the behavior now matches what we had before, there is a pre-existing flaw: when a goal error is present, it is not cleared until the _completion_ of the next error-free action; ideally it should be cleared first.
> 
> That is how it used to work, and the old behavior is what was specified in the ticket. I am worried a different behavior will take more time to implement than I have left on this project along with everything else I have to do, but I will at least investigate the possibility.
> 
> > Also, while I agree FaultClear should not print any warnings, it would be good to see a warning (or INFO message) whenever a goal error is automatically cleared.
> 
> This already happens in simulator output. Keep in mind action clients scripts do not print anything.

AstroStucky:
I implemented both of your suggestions. Goal errors are now cleared at the start of the next action, and "auto_clear_goal_errors" is now set in RQT through the dynamic reconfigure server. Please retest as needed. The PR test has been updated.

kmdalal:
I don't know if it's related to this change, but it seemed like every CameraCapture in ReferenceMission2 had a goal error:

[ERROR]: CameraCapture: Aborted - Timed out waiting for point cloud

At least this allowed me to toggle the option during the sequence, and the fault clearing was as expected (automatic or via the plan).

kmdalal:
Correction to my last comment: a few of the CameraCaptures succeeded, including the final one which is presumably the one that found a sample target.

kmdalal (approved):
Seems to be working well and the code changes are straightforward.  I especially like the new fault group in RQT.  Nice work!

PR #401, opened 2024-05-31
Oceanwater 1164 means to reposition lander

AstroStucky:
## Linked Issues:
Jira Ticket 🎟️ | [OCEANWATER-1164](https://babelfish.arc.nasa.gov/jira/browse/OCEANWATER-1164)

## Sibling PR
https://github.com/nasa/ow_europa/pull/38

## Summary of Changes
* Arguments can now be provided to all world launch files to specify lander initial pose in 3D position and Euler angles.   
* The voxel grid created by MaterialDistributionPlugin can now be positioned relative to a model in the world. 
  * The model should be relatively level with the world frame. Only yaw (z-rotation) will be applied to the grid. 
  * This allows the voxel grid, and therefore the workspace in which a material distribution will be simulated, to follow the lander to wherever it's moved to. 

## Test
For this test we will run ReferenceMission1 in two different locations of atacama_y1a
### In the default position (regression)
1. Launch atacama_y1a with the lander in the default position `roslaunch ow atacama_y1a.launch`
2. Run ReferenceMission1 in another terminal `roslaunch ow_plexil ow_exec.launch plan:=ReferenceMission1.plx`

The test should work the same as it does in noetic-devel. 

### In a new lander position
1. Launch atacama_y1a with the lander in a new position. `roslaunch ow atacama_y1a.launch init_x:=-2.0 init_y:=-2.0 init_Y:=0.707`. The lander will wind up positioned as in the image.
default_gzclient_camera(1)-2024-05-31T16_26_09 791872
Feel free to use a position of your own choice, the mission plan should work so long as there is terrain in front of the lander for it to dig through. 
2. Going into Rviz and ticking material_grid will reveal that the generated material grid looks entirely different form the one that's generated in the default lander position. This is the because the grid domain now follows the lander wherever it is moved. If you used the same lander pose as I used, it will look like the following image. 
Screenshot_20240531_160527
3. Run ReferenceMission1 in another terminal `roslaunch ow_plexil ow_exec.launch plan:=ReferenceMission1.plx`

#### Test Caveats
Test should work, but it may have to use the fallback position, and the scoop will likely dig excessively deep. Because the scoop digs excessively deep, the following warnings are likely to be spammed. 
```
[Wrn] [RegolithPlugin.cpp:273] RegolithPlugin: Regolith particle spawn skipped to avoid particle overlap.
[Wrn] [RegolithPlugin.cpp:308] RegolithPlugin: More volume is being displaced per frame than can be spawned as regolith. Consider increasing the value of spawn_volume_threshold in common.launch to avoid this warning.
```  
The real issue here is that the reference mission plans need to be made more robust to new terrain positions. I would like to address this in a hotfix that ensures GuardedMove always happens so a proper terrain height is queried. 

## Caveats
In worlds where multimaterial is used, the "relative_to" parameter must be a model that is level to the XY plane. This is because the voxel grid must be level to align itself with the heightmap modifications.

### DIFF
--- a/ow/launch/atacama_y1a.launch
+++ b/ow/launch/atacama_y1a.launch
@@ -9,16 +9,22 @@
   <arg name="rqt_gui" default="true"/>
   <arg name="use_rviz" default="true"/>
   <arg name="sim_regolith" default="true" />
+  <arg name="init_x" default="-1"/>
+  <arg name="init_y" default="0"/>
+  <arg name="init_z" default="0.37"/>
+  <arg name="init_R" default="0"/>
+  <arg name="init_P" default="0"/>
+  <arg name="init_Y" default="0"/>
 
   <include file="$(find ow)/launch/common.launch">
     <arg name="gazebo_launch_file" default="$(arg gazebo_launch_file)" />
     <arg name="world_name" value="$(find ow_europa)/worlds/atacama_y1a.world"/>
-    <arg name="init_x" value="-1"/>
-    <arg name="init_y" value="0"/>
-    <arg name="init_z" value="0.37"/>
-    <arg name="init_R" value="0"/>
-    <arg name="init_P" value="0"/>
-    <arg name="init_Y" value="0"/>
+    <arg name="init_x" value="$(arg init_x)"/>
+    <arg name="init_y" value="$(arg init_y)"/>
+    <arg name="init_z" value="$(arg init_z)"/>
+    <arg name="init_R" value="$(arg init_R)"/>
+    <arg name="init_P" value="$(arg init_P)"/>
+    <arg name="init_Y" value="$(arg init_Y)"/>
     <arg name="freeze_base_link" value="true"/>
     <arg name="arm_sim_enable" value="$(arg arm_sim_enable)"/>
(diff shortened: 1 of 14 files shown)


kmdalal:
In the test description, is the following line accurate? "init_x:=-2.0 init_y:=-2.0 init_Y:=0.707".  I'm referring to the lowercase and uppercase Y.

kmdalal:
The given coordinates don't work in europa_terminator -- the lander is partially buried.  How does one specify the elevation of the lander, and why are both 'y' and 'Y' used?

AstroStucky:
The coordinates I provided are specific to atacama_y1a. 

The arguments are
init_x, init_y, init_z for the world (x, y, z) position of the lander. Changing z will change elevation.

init_R, init_P, init_Y for the lander's orientation provided in Euler angles. init_R is roll, init_P is pitch, and init_Y is yaw.

AstroStucky:
Just FYI changing the lander's roll and pitch is not supported along with multimaterial terrain.

kmdalal (approved):
Tests passed.  Code review was light.  Good new feature!

